PR #1, opened 2022-12-15
Initial client and bridge

cretz:
## What was changed

This is the first _real_ work on this repo. Most notably it includes:

* README and LICENSE
* Scaffolding for:
  * Proto gen
  * Code formatting
  * Testing
  * API doc gen
* Generated proto classes
* Rust bridge which includes:
  * Low-level manual memory management
  * C header file gen
  * C# interop generator
  * Hand-written C# layer for the bridge
* High-level connection with access to raw gRPC for primary services
* High-level client (no high-level calls yet)
* API docs for all user-facing API
* Workflow test environment for local Temporalite use

The near future:

* CI (cross platform and including dotnet.temporal.io doc site push)
* Update sdk-core to get new API's as https://github.com/temporalio/sdk-java/pull/1572 and https://github.com/temporalio/api/pull/251 are merged
* High-level client methods
* Converters
* Exceptions
* Tests

This is mostly based on the proposal at https://github.com/temporalio/proposals/pull/71. Please review.


.editorconfig
```diff
@@ -0,0 +1,129 @@
+# Inspired by other projects
+[*.cs]
+
+### Temporal additions ##
+
+# Warn on unused imports
+dotnet_diagnostic.IDE0005.severity = warning
+dotnet_diagnostic.IDE0005.severity = warning
```
cretz: Not sure why I have a dupe here, will fix

src/Temporalio/Api/Dependencies/Gogo.cs
```diff
@@ -0,0 +1,290 @@
...
+//     source: dependencies/gogoproto/gogo.proto
+// </auto-generated>
+#pragma warning disable 1591, 0612, 3021, 8981
+#region Designer generated code
+
+using pb = global::Google.Protobuf;
+using pbc = global::Google.Protobuf.Collections;
+using pbr = global::Google.Protobuf.Reflection;
+using scg = global::System.Collections.Generic;
+namespace Gogoproto {
```
cretz: Consider a custom code rewrite step to move this (and references) to `Temporalio.Api.Dependencies.Gogoproto` (this gogo proto Go lib has infected all of our SDKs 😞)

src/Temporalio/Bridge/ByteArray.cs
```diff
@@ -0,0 +1,46 @@
+using System;
+using System.IO;
+using System.Runtime.InteropServices;
+using Google.Protobuf;
+
+namespace Temporalio.Bridge
+{
+    internal class ByteArray : SafeHandle
```
cretz, line 11: It took a long time to learn how to safely use some of these pointers in .NET. Close scrutiny on this `Temporalio.Bridge` namespace as a whole is welcome.

src/Temporalio/Bridge/include/temporal-sdk-bridge.h
```diff
@@ -0,0 +1,253 @@
+#pragma once
```
cretz, line 1: Note how there is nothing .NET specific here. This whole Rust project should be extracted out of this repo and repurposed for general C FFI use when we need it (e.g. for CGo, JNI, etc).

src/Temporalio/Bridge/src/lib.rs
```diff
@@ -0,0 +1,173 @@
...
+
+use std::collections::HashMap;
+
+#[repr(C)]
+pub struct ByteArrayRef {
+    data: *const u8,
+    size: libc::size_t,
+}
+
+impl ByteArrayRef {
```
cretz, line 23: Earlier versions of these impls were all sorts of `AsRef`, `From`, `TryFrom`, etc impls. But it became a bit confusing what was and wasn't and what exact conversions were happening, so I moved it back to explicit conversions which clarified everything at a minor loss of ergonomics.

src/Temporalio/Bridge/src/lib.rs
```diff
@@ -0,0 +1,173 @@
...
+        unsafe { std::slice::from_raw_parts(self.data, self.size) }
+    }
+
+    fn to_vec(&self) -> Vec<u8> {
+        unsafe { std::slice::from_raw_parts(self.data, self.size).to_vec() }
+    }
+
+    fn to_str(&self) -> &str {
+        // Trust caller to send UTF8
+        unsafe { std::str::from_utf8_unchecked(std::slice::from_raw_parts(self.data, self.size)) }
```
cretz, line 36: Note on these calls I trust the lang side to send the right thing all the time. Even if I changed this to `from_utf8` and propagated the error, lang could still mess us up in a lot of ways including just bad pointers.

src/Temporalio/Bridge/src/client.rs
```diff
@@ -0,0 +1,453 @@
...
+                service_call!(call_operator_service, client, options, cancel_token)
+            }
+            RpcService::Test => service_call!(call_test_service, client, options, cancel_token),
+            RpcService::Health => service_call!(call_health_service, client, options, cancel_token),
+        };
+        let (success, fail) = match res {
+            Ok(b) => (ByteArray::from_vec(b).into_raw(), std::ptr::null_mut()),
+            Err(err) => (
+                std::ptr::null_mut(),
+                ByteArray::from_utf8(format!("{}", err)).into_raw(),
```
cretz, line 213:
```suggestion
                client.runtime.alloc_utf8(format!("{}", err)).into_raw(),
```

Even though we're not using a shared buffer pool currently, we _could_ one day, so we need to centralize allocations (I forgot this one and the `ByteArray::from_vec` above)

src/Temporalio/Bridge/src/client.rs
```diff
@@ -0,0 +1,453 @@
...
+    //     Err(err) => {
+    //         Err(Python::with_gil(move |py| {
+    //             // Create tuple of "status", "message", and optional "details"
+    //             let code = err.code() as u32;
+    //             let message = err.message().to_owned();
+    //             let details = err.details().into_py(py);
+    //             RPCError::new_err((code, message, details))
+    //         }))
+    //     }
+    // }
```
cretz, line 389: Commented out from Python, but the goal is to get proper advanced errors over to .NET (gRPC status details is something no other .NET gRPC user has :-))

src/Temporalio/Client/TemporalClientOptions.cs
```diff
@@ -0,0 +1,32 @@
+namespace Temporalio.Client
+{
+    /// <summary>
+    /// Options for a <see cref="TemporalClient" />.
+    /// </summary>
+    public class TemporalClientOptions
```
cretz:
```suggestion
    public class TemporalClientOptions : ICloneable
```

I implemented the method but not the interface

src/Temporalio/Client/TemporalClientOptions.cs
```diff
@@ -0,0 +1,32 @@
...
+        public TemporalClientOptions()
+        {
+            // TODO(cretz): DataConverter = Converters.DataConverter.Default;
+        }
+
+        /// <summary>
+        /// Create a shallow copy of these options.
+        /// </summary>
+        /// <returns>A shallow copy of these options.</returns>
+        public object Clone()
```
cretz: We need shallow copying of all of our options classes. Options were:

1. Use records as options, but that doesn't fit https://learn.microsoft.com/en-us/dotnet/core/extensions/options
2. Create a copy constructor, but that is prone to hand-typed mistakes
3. Impl shallow `ICloneable` and expect caller to cast

3 seems best (there are other nuances, but this is the gist of it). I wish I could just have immutable classes/records and use `with`, but it doesn't fit with the .NET config/options preferences.

src/Temporalio/Client/TemporalConnection.cs
```diff
@@ -0,0 +1,79 @@
...
+        public TestService TestService { get; private init; }
+
+        /// <summary>
+        /// Check health for the given service type.
+        /// </summary>
+        /// <param name="service">Service type to check health for. Defaults to
+        /// <see cref="TemporalConnection.WorkflowService" /></param>
+        /// <param name="options">RPC options for the check call.</param>
+        /// <returns>True if healthy, false otherwise.</returns>
+        public Task<bool> CheckHealthAsync(RpcService? service = null, RpcOptions? options = null)
```
cretz, line 45: Can't decide whether to put this on the `ITemporalConnection` interface.

And in general I am not a fan of interfaces just for the sake of having them. But it's the .NET way for these things that may be DI-based, so I accommodate.

src/Temporalio/Client/ITemporalClient.cs
```diff
@@ -0,0 +1,14 @@
+namespace Temporalio.Client
+{
+    /// <summary>
+    /// Interface to a client to Temporal.
+    /// </summary>
+    /// <seealso cref="TemporalClient" />
+    public interface ITemporalClient : Worker.IWorkerClient
```
cretz, line 7: In general I am not a fan of interfaces just for the sake of having them. But it's the .NET way for these things that may be DI-based, so I accommodate. That means that I will have to put all those `StartWorkflowAsync` overloads on here. One might think I could put one on here and make everything else an extension method, but .NET prefers putting things on the types themselves if you can, and extension methods if you can't (for compat or external-code reasons). See https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/extension-methods#general-guidelines

src/Temporalio/Client/TlsOptions.cs
```diff
@@ -0,0 +1,35 @@
+namespace Temporalio.Client
+{
+    /// <summary>
+    /// TLS options for Temporal connections.
+    /// </summary>
+    public class TlsOptions
```
cretz:
```suggestion
    public class TlsOptions : ICloneable
```

And properly impl

src/Temporalio/Client/TemporalConnectionOptions.cs
```diff
@@ -0,0 +1,112 @@
...
+                            throw new ArgumentException("TargetHost does not have valid port");
+                        }
+                        port = portInt;
+                    }
+                }
+            }
+        }
+
+        /// <summary>
+        /// Create a shallow copy of these options.
```
cretz, line 104:
```suggestion
        /// Create a shallow copy of these options and their child options.
```

This means I need to change `Clone` below to clone the `TlsOptions` and the `RpcRetryOptions`

src/Temporalio/Client/TemporalClientOptions.cs
cretz: Looks like I missed it on `RpcOptions`, `Testing.TemporaliteOptions`, etc...

src/Temporalio/Testing/WorkflowEnvironmentStartLocalOptions.cs
```diff
@@ -0,0 +1,29 @@
+namespace Temporalio.Testing
+{
+    /// <summary>
+    /// Options for <see cref="WorkflowEnvironment.StartLocalAsync" />.
+    /// </summary>
+    public class WorkflowEnvironmentStartLocalOptions : Client.TemporalClientConnectOptions
```
cretz, line 6: Override `Clone` and make sure to clone the `TemporaliteOptions` options too (same for `WorkflowEnvironmentStartTimeSkippingOptions`)

src/Temporalio/Worker/IWorkerClient.cs
```diff
@@ -0,0 +1,20 @@
+namespace Temporalio.Worker
+{
+    /// <summary>
+    /// Client that can be used to create a worker.
+    /// </summary>
+    public interface IWorkerClient
```
cretz, line 6: This interface is implemented by `TemporalClient` and is technically all that is needed for a worker to function. This is a nice abstraction separation (and the reason `IBridgeClientProvider` exists). This means users can have their own impl of `ITemporalClient` (e.g. a wrapper around ours) and since this interface is still required to be impl'd, it'll work for a worker.

.editorconfig
cretz: Fixed

src/Temporalio/Client/TemporalClientOptions.cs
cretz: Fixed

src/Temporalio/Client/TlsOptions.cs
cretz: Fixed

src/Temporalio/Client/TemporalConnectionOptions.cs
cretz: Fixed

src/Temporalio/Testing/WorkflowEnvironmentStartLocalOptions.cs
cretz: Fixed

.gitignore
```diff
@@ -0,0 +1,7 @@
+
+# Build results
+bin/
+obj/
+/src/Temporalio/Bridge/target
+/src/Temporalio/Bridge/Cargo.lock
```
mjameswh: I wouldn't gitignore `Cargo.lock` in this project.

README.md
```diff
@@ -1,4 +1,94 @@
...
+Can add options like:
+
+* `--logger "console;verbosity=detailed"` to show logs
+  * TODO(cretz): This doesn't show Rust stdout. How do I do that?
+* `--filter "FullyQualifiedName=Temporalio.Tests.Client.TemporalClientTests.ConnectAsync_Connection_Succeeds"` to run a
+  specific test
+
+### Rebuilding Rust extension and interop layer
+
+To rebuild DLL, make sure `protoc` is on the `PATH`, then:
```
mjameswh: Clarification: Is this (`cargo build` and ClandSharpPInvokeGenerator) something that devs are expected to do manually from time to time, or will that be included in msbuild build process? Seems to me like `cargo build` is already covered by Bridge/Temporalio.csproj.

mjameswh: Depending on your answer, might want to consider adding `include/temporal-sdk-bridge.h` and `Interop/*.cs` to gitignore.

README.md
```diff
@@ -1,4 +1,94 @@
...
+
+Then, run:
+
+    ClangSharpPInvokeGenerator @src/Temporalio/Bridge/GenerateInterop.rsp
+
+### Regenerating protos
+
+Must have `protoc` on the `PATH`, then:
+
+    dotnet run --project src/Temporalio.Api.Generator
```
mjameswh, line 69: Again... When/Who shold be doing this? Commited to git or gitignore?

.gitignore
cretz: Agreed, this is a mistake by me. Will fix.

README.md
cretz: > Is this (cargo build and ClandSharpPInvokeGenerator) something that devs are expected to do manually from time to time, or will that be included in msbuild build process? Seems to me like cargo build is already covered by Bridge/Temporalio.csproj.

The `cargo build` is now part of the msbuild process. Just added, forgot to remove this from the README (this README is almost just like local dev notes for now).

The `ClangSharpPInvokeGenerator` will be expected to be invoked manually whenever any public-facing C FFI piece changes. There will be a CI step ensuring that what's in the header file matches what would be generated when re-run. We do want the generated code committed so anyone can just clone + build and so that it's easy to browse the code.

README.md
cretz: Developer should do this on API update. There will be a CI step that ensures that what's in repo matches what would be generated when re-run. We do want the generated code committed so anyone can just clone + build and so that it's easy to browse the code.

src/Temporalio/Bridge/ByteArrayRef.cs
```diff
@@ -0,0 +1,112 @@
...
+                }
+            }
+            if (stream.Length == 0)
+            {
+                return Empty;
+            }
+            return new ByteArrayRef(stream.GetBuffer(), (int)stream.Length);
+        }
+
+        public static ByteArrayRef FromNewlineDelimited(IEnumerable<string> values)
```
bergundy, line 75: Would be nice to have some doc strings on these methods.

src/Temporalio/Bridge/Interop/Interop.cs
```diff
@@ -0,0 +1,408 @@
...
+        public byte* data;
+
+        [NativeTypeName("size_t")]
+        public UIntPtr size;
+
+        [NativeTypeName("size_t")]
+        public UIntPtr cap;
+
+        [NativeTypeName("bool")]
+        public byte disable_free;
```
bergundy, line 101: Do bool and byte necessarily have the same memory layout?

src/Temporalio/Bridge/OptionsExtensions.cs
```diff
@@ -0,0 +1,149 @@
...
+            Temporalio.Runtime.Version
+        );
+        private static readonly ByteArrayRef SdkName = ByteArrayRef.FromUTF8("sdk-dotnet");
+
+        public static unsafe Interop.ClientOptions ToInteropOptions(
+            this Temporalio.Client.TemporalConnectionOptions options,
+            Scope scope
+        )
+        {
+            if (options.TargetHost == null)
```
bergundy: Check for empty string too?

src/Temporalio/Bridge/OptionsExtensions.cs
```diff
@@ -0,0 +1,149 @@
...
+            {
+                throw new ArgumentException("TargetHost is required");
+            }
+            else if (options.TargetHost.Contains("://"))
+            {
+                throw new ArgumentException("TargetHost cannot have ://");
+            }
+            return new Interop.ClientOptions()
+            {
+                target_url = scope.ByteArray($"http://{options.TargetHost}"),
```
bergundy: AFAIR, when using TLS, the URL should start with `https://`.

src/Temporalio/Bridge/Runtime.cs
```diff
@@ -0,0 +1,48 @@
...
+        internal readonly unsafe Interop.Runtime* ptr;
+
+        public Runtime() : base(IntPtr.Zero, true)
+        {
+            using (var scope = new Scope()) { }
+            unsafe
+            {
+                var options = new Interop.RuntimeOptions();
+                var res = Interop.Methods.runtime_new(&options);
+                GC.KeepAlive(options);
```
bergundy, line 20: It took me a second to figure out what this does.
It's a bit weird that it keeps `options` alive until `KeepAlive` is called.

src/Temporalio/Bridge/src/client.rs
```diff
@@ -0,0 +1,453 @@
...
+    fail: *const ByteArray,
+);
+
+macro_rules! service_call {
+    ($service_fn:ident, $client:ident, $options:ident, $cancel_token:ident) => {{
+        let call_future = $service_fn(&$client.core, &$options);
+        if let Some(cancel_token) = $cancel_token {
+            tokio::select! {
+                _ = cancel_token.cancelled() => Err(anyhow::anyhow!("Cancelled")),
+                v = call_future => v,
```
bergundy, line 178: Too bad we can't cancel the request here, I couldn't find that option in tonic.
I'd at least document that it doesn't cancel the request (AFAICT).

src/Temporalio/Bridge/src/client.rs
```diff
@@ -0,0 +1,453 @@
...
+    success: *const ByteArray,
+    fail: *const ByteArray,
+);
+
+macro_rules! service_call {
+    ($service_fn:ident, $client:ident, $options:ident, $cancel_token:ident) => {{
+        let call_future = $service_fn(&$client.core, &$options);
+        if let Some(cancel_token) = $cancel_token {
+            tokio::select! {
+                _ = cancel_token.cancelled() => Err(anyhow::anyhow!("Cancelled")),
```
bergundy, line 177: Do you want to use anyhow or better typed errors?

bergundy (approved):
I got all the way to the Rust bridge client and had to go.
Figured I'd submit what I have so far.
Looks good!

src/Temporalio/Bridge/ByteArrayRef.cs
cretz: They are internal methods to the project (methods public, but class is `internal`). I can add _some_ docs, but it's not reasonable to document like public API IMO.

src/Temporalio/Bridge/Interop/Interop.cs
cretz: This is auto-generated from ClangSharp. https://github.com/dotnet/ClangSharp/issues/350#issuecomment-1155385609 mentions that `bool` is `byte` by intention here. After doing some digging (e.g. [this SO answer](https://stackoverflow.com/a/53925256), [this SO answer](https://stackoverflow.com/a/32115697), and [this SO answer](https://stackoverflow.com/a/20037067)) show that `bool` is "non-blittable" and therefore unsafe for certain interop uses. So that's why the ClangSharp project did this.

Note btw that this specific field is never used on the C# side, it's for Rust to know not to free the space (it'd be used for static byte arrays).

src/Temporalio/Bridge/OptionsExtensions.cs
cretz: :+1: will update (C# has a nice `IsNullOrEmpty` extension for just this case)

src/Temporalio/Bridge/OptionsExtensions.cs
cretz: :+1: Will fix

src/Temporalio/Bridge/Runtime.cs
cretz: Yeah, the method actually does nothing, this just tells the compiler this value is still referenced so it doesn't eagerly garbage collect it before/during the external call. Though I probably should just use that new "Scope" feature I made which keeps multiple things alive.

src/Temporalio/Bridge/src/client.rs
cretz: I believe it does cancel the request. I believe the `call_future` will be dropped which will cancel the future which I would hope tonic translates into gRPC request cancel. @Sushisource - can you confirm?

src/Temporalio/Bridge/src/client.rs
cretz: Yes for sure (and on RPC errors too). I need to use smarter bridge errors here. I'll end up throwing https://learn.microsoft.com/en-us/dotnet/api/system.operationcanceledexception from lang side. I'll also write a test for this (you'll note many things like proper exceptions are not included in this initial version).

cretz:
> I got all the way to the Rust bridge client and had to go.

@bergundy - Thanks. Will look forward to the rest.

src/Temporalio/Bridge/src/lib.rs
```diff
@@ -0,0 +1,173 @@
...
+    size: libc::size_t,
+}
+
+impl ByteArrayRef {
+    fn to_slice(&self) -> &[u8] {
+        unsafe { std::slice::from_raw_parts(self.data, self.size) }
+    }
+
+    fn to_vec(&self) -> Vec<u8> {
+        unsafe { std::slice::from_raw_parts(self.data, self.size).to_vec() }
```
bergundy: nit: can reuse `to_slice`

src/Temporalio/Bridge/src/lib.rs
bergundy: This sounds like a nice comment to embed in the code.

src/Temporalio/Bridge/src/lib.rs
```diff
@@ -0,0 +1,173 @@
...
+            Some(self.to_str())
+        }
+    }
+
+    fn to_option_string(&self) -> Option<String> {
+        self.to_option_str().map(str::to_string)
+    }
+
+    fn to_str_map_on_newlines(&self) -> HashMap<&str, &str> {
+        let strs: Vec<&str> = self.to_str().split('\n').collect();
```
bergundy, line 73: Hmm.. what if the keys or values contain newlines?
There's probably a safer way to encode this, I'd probably just use JSON or proto.

src/Temporalio/Bridge/ByteArrayRef.cs
```diff
@@ -0,0 +1,112 @@
...
+        {
+            var stream = new MemoryStream();
+            using (var writer = new StreamWriter(stream, StrictUTF8))
+            {
+                foreach (var pair in metadata)
+                {
+                    // If either have a newline, we error since it would make an invalid set
+                    if (pair.Key.IndexOf('\n') >= 0 || pair.Value.IndexOf('\n') >= 0)
+                    {
+                        throw new ArgumentException("Metadata keys/values cannot have newlines");
```
bergundy, line 52: Why have this limitation?

src/Temporalio/Bridge/src/lib.rs
bergundy: Also added a comment in the C# code, asking why we're adding the newline limitation.

src/Temporalio/Bridge/src/lib.rs
```diff
@@ -0,0 +1,173 @@
...
+}
+
+impl ByteArray {
+    fn from_utf8(str: String) -> ByteArray {
+        ByteArray::from_vec(str.into_bytes())
+    }
+
+    fn from_vec(vec: Vec<u8>) -> ByteArray {
+        // Mimics Vec::into_raw_parts that's only available in nightly
+        let mut vec = std::mem::ManuallyDrop::new(vec);
```
bergundy, line 104: Is there anything that `drop` does for `Vec` that we should manually do other than free the data?

src/Temporalio/Bridge/src/lib.rs
```diff
@@ -0,0 +1,173 @@
...
+unsafe impl Sync for ByteArray {}
+
+impl Drop for ByteArray {
+    fn drop(&mut self) {
+        // In cases where freeing is disabled (or technically some other
+        // drop-but-not-freed situation though we don't expect any), the bytes
+        // remain non-null so we re-own them here
+        if !self.data.is_null() {
+            unsafe { Vec::from_raw_parts(self.data as *mut u8, self.size, self.cap) };
+        }
```
bergundy: Doesn't this effectively free the data?
Should we bail in case `disable_free` is set here?

src/Temporalio/Client/ITemporalConnection.cs
```diff
@@ -0,0 +1,28 @@
+namespace Temporalio.Client
+{
+    /// <summary>
+    /// Interface to a connection to Temporal.
+    /// </summary>
+    /// <seealso cref="TemporalConnection" />
+    public interface ITemporalConnection : IBridgeClientProvider
```
bergundy, line 9: You forgot to expose the HealthService

src/Temporalio/Client/OperatorService.cs
```diff
@@ -0,0 +1,41 @@
+using System.Threading.Tasks;
+using Google.Protobuf;
+
+namespace Temporalio.Client
+{
+    /// <summary>
+    /// Raw gRPC operator service.
+    /// </summary>
+    public abstract partial class OperatorService : RpcService
```
bergundy, line 9: Should there be an `IOperatorService` too in the sprit of the language?

src/Temporalio/Client/RpcOptions.cs
```diff
@@ -0,0 +1,48 @@
...
+    /// </summary>
+    public class RpcOptions : ICloneable
+    {
+        /// <summary>
+        /// Gets or sets whether the call will retry.
+        /// </summary>
+        /// <remarks>
+        /// High-level client calls retry by default, low-level calls do not.
+        /// </remarks>
+        public bool? Retry { get; set; }
```
bergundy, line 18: Is this meant to be exposed to users? (pardon me for not checking where this is used).

src/Temporalio/Client/RpcOptions.cs
```diff
@@ -0,0 +1,48 @@
...
+
+        /// <summary>
+        /// Gets or sets the gRPC metadata for the call (i.e. the headers).
```
bergundy, line 21: I'd note that the connection / client has globally associated metadata and that these values are merged with and override the global ones.

src/Temporalio/Client/RpcRetryOptions.cs
```diff
@@ -0,0 +1,49 @@
+using System;
+
+namespace Temporalio.Client
+{
+    /// <summary>
+    /// Retry options that can be set at the connection level to apply on calls that are retried.
+    /// </summary>
+    public class RpcRetryOptions : ICloneable
```
bergundy, line 8: Do we expose this publicly? I don't think we do in other SDKs.

src/Temporalio/Client/TemporalConnection.cs
```diff
@@ -0,0 +1,79 @@
...
+                resp,
+                options?.Retry ?? false,
+                options?.Metadata,
+                options?.Timeout,
+                options?.CancellationToken
+            );
+        }
+
+        /// <inheritdoc />
+        public SafeHandle BridgeClient => this.client;
```
bergundy, line 71: Why is this public?

src/Temporalio/Client/TemporalConnection.cs
bergundy: I would put it on the interface for completeness.

src/Temporalio/Client/ITemporalConnection.cs
bergundy: OIC that you exposed a `CheckHealthAsync` method below, that's a valid alternative.

src/Temporalio/Client/TemporalConnection.cs
bergundy: When reading this PR top-down, I was missing the health check method on the interface.

src/Temporalio/Client/TemporalConnectionOptions.cs
```diff
@@ -0,0 +1,122 @@
...
+        public TlsOptions? TlsOptions { get; set; }
+
+        /// <summary>
+        /// Gets or sets retry options for this connection.
+        /// </summary>
+        /// <remarks>
+        /// This only applies if the call is being retried, which by default is usually all
+        /// high-level client calls but no raw gRPC calls.
+        /// </remarks>
+        public RpcRetryOptions? RpcRetryOptions { get; set; }
```
bergundy, line 50: (As noted elsewhere) I don't think we expose this in other SDKs and I tend to think that we should use sane defaults and introduce this configuration if there's demand.

src/Temporalio/Client/TemporalConnectionOptions.cs
```diff
@@ -0,0 +1,122 @@
...
+        /// By default this uses <see cref="Runtime.Default" /> which is lazily created when first
+        /// needed.
+        /// </remarks>
+        public Runtime? Runtime { get; set; }
+
+        internal void ParseTargetHost(out string? ip, out int? port)
+        {
+            ip = null;
+            port = null;
+            if (TargetHost != null)
```
bergundy, line 79: nit: or empty

bergundy: What about an `else` clause?

src/Temporalio/Client/TemporalConnectionOptions.cs
```diff
@@ -0,0 +1,122 @@
...
+        /// Create a shallow copy of these options.
+        /// </summary>
+        /// <returns>A shallow copy of these options.</returns>
+        /// <remarks>Does not create copies of RPC metadata or runtime.</remarks>
+        public virtual object Clone()
+        {
+            var copy = (TemporalConnectionOptions)this.MemberwiseClone();
+            if (TlsOptions != null)
+            {
+                copy.TlsOptions = (TlsOptions)TlsOptions.Clone();
```
bergundy, line 113: Is this not shallowly copied by `MemberwiseClone`?

src/Temporalio/Client/TestService.cs
```diff
@@ -0,0 +1,41 @@
...
+            /// Create an implementation of the test service.
+            /// </summary>
+            /// <param name="connection">Connection to use.</param>
+            public Impl(TemporalConnection connection)
+            {
+                this.connection = connection;
+            }
+
+            /// <inheritdoc />
+            protected async override Task<T> InvokeRpcAsync<T>(
```
bergundy, line 30: Can't this method be defined on `RpcService`? Is there ever a case where a service won't have a `TemporalConnection`?

src/Temporalio/Client/TestService.cs
```diff
@@ -0,0 +1,41 @@
...
+    /// Raw gRPC test service.
+    /// </summary>
+    public abstract partial class TestService : RpcService
+    {
+        internal override Bridge.Interop.RpcService Service => Bridge.Interop.RpcService.Test;
+
+        /// <summary>
+        /// Implementation of the test service.
+        /// </summary>
+        public class Impl : TestService
```
bergundy, line 16: Why is this inner `Impl` class required?

src/Temporalio/Runtime.cs
```diff
@@ -0,0 +1,41 @@
...
+    /// This runtime carries the internal core engine and telemetry options for Temporal. All
+    /// connections/clients created using it, and any workers created from them, will be associated
+    /// with the runtime.
+    /// </remarks>
+    public sealed class Runtime
+    {
+        /// <summary>
+        /// Current version of this SDK.
+        /// </summary>
+        public const string Version = "0.1.0-alpha1";
```
bergundy, line 18: Is this intentionally hardcoded here?
Is there any duplicate definition of this version string?

src/Temporalio/Runtime.cs
```diff
@@ -0,0 +1,41 @@
...
+{
+    /// <summary>
+    /// Runtime for the Temporal SDK.
+    /// </summary>
+    /// <remarks>
+    /// This runtime carries the internal core engine and telemetry options for Temporal. All
+    /// connections/clients created using it, and any workers created from them, will be associated
+    /// with the runtime.
+    /// </remarks>
+    public sealed class Runtime
```
bergundy, line 13: Where are the runtime options supposed to be provided?

src/Temporalio/Testing/TestServerOptions.cs
```diff
@@ -0,0 +1,45 @@
...
+    /// </remarks>
+    public class TestServerOptions : ICloneable
+    {
+        /// <summary>
+        /// Gets or sets the existing executable path for the test server.
+        /// </summary>
+        public string? ExistingPath { get; set; }
+
+        /// <summary>
+        /// Gets or sets the version to version of the test server to download.
```
bergundy, line 20: Might want to mention which values are valid here (`vN.N.N` and `latest`)

bergundy (commented):
Finished going over this.
I don't think I had anything critical or blocking, mostly questions.

.gitignore
Sushisource: +1

src/Temporalio/Bridge/ByteArrayRef.cs
Sushisource: Not like a public API, just some high level overview on how memory passing is handled probably makes sense, though it seems fairly straightforward

src/Temporalio/Bridge/src/client.rs
Sushisource: Yeah, that should work properly

src/Temporalio/Bridge/src/client.rs
```diff
@@ -0,0 +1,453 @@
...
+    ($client:ident, $call:ident, $call_name:ident) => {
+        if $call.retry {
+            rpc_resp($client.$call_name(rpc_req($call)?).await)
+        } else {
+            rpc_resp($client.into_inner().$call_name(rpc_req($call)?).await)
+        }
+    };
+}
+
+async fn call_workflow_service(
```
Sushisource, line 232: This is redone in pretty much every core consumer with slight variations. We should make it slightly more generic and upstream it. Doesn't need to block this PR of course.

src/Temporalio/Bridge/ByteArrayRef.cs
```diff
@@ -0,0 +1,112 @@
...
+            }
+            return new ByteArrayRef(StrictUTF8.GetBytes(s));
+        }
+
+        public static ByteArrayRef FromProto(IMessage p)
+        {
+            return new ByteArrayRef(p.ToByteArray());
+        }
+
+        public static ByteArrayRef FromMetadata(IEnumerable<KeyValuePair<string, string>> metadata)
```
Sushisource, line 42: The way metadata is done is a bit odd to me. I see two simpler options:

* Extra simple: Just use a proto rather than this custom text based format
* Slightly less wasteful, but more annoying to build/free: Make a specific byte ref type in C like this:
```C
struct MetadataRef {
  const uint8_t* key_then_data;
  size_t key_size;
  size_t data_size;
  struct MetadataRef* next;
}
```

Lets you get rid of the special newline handling.

In general, I think it probably makes sense to represent these different "kinds" of byte arrays as different C structs, even if they maye have the exact same representation (like string & proto). It makes the C definitions read more clearly and expresses intent a bit better.

src/Temporalio/Bridge/ByteArray.cs
Sushisource: Can you expand a bit on the choice of `SafeHandle` here and what that's providing (that'd be the kind of thing that'd be good in a docstring somewhere).

I think this makes good sense, but the .NET docs kinda imply it's largely about windows stuff. Just a sentence or two about how it can be used to deal with externally managed resources generally would be great

src/Temporalio/Bridge/src/client.rs
```diff
@@ -0,0 +1,453 @@
...
+impl TryFrom<&ClientOptions> for CoreClientOptions {
+    type Error = anyhow::Error;
+
+    fn try_from(opts: &ClientOptions) -> anyhow::Result<Self> {
+        let mut opts_builder = ClientOptionsBuilder::default();
+        opts_builder
+            .target_url(Url::parse(opts.target_url.to_str())?)
+            .client_name(opts.client_name.to_str())
+            .client_version(opts.client_version.to_str())
+            .identity(opts.identity.to_string())
```
Sushisource: It's a bit odd that two of these are `str` and one is `String`. FYI the first two are just going to get converted into `String`s straight away.

src/Temporalio/Bridge/src/lib.rs
Sushisource: +1

src/Temporalio/Bridge/src/lib.rs
Sushisource: This is the right way

src/Temporalio/Bridge/src/lib.rs
Sushisource: This is fine (though clarification on the `disable_free` not being checked would be good).

But, this method of freeing is only safe iff the `ByteArray`'s memory was built allocated by creating a `Vec`. That's the case, which is great, but `ByteArray` should have a big fat safety docstring indicating that no one should add constructors that allocate memory any other way than via a `Vec`

src/Temporalio/Bridge/src/runtime.rs
```diff
@@ -0,0 +1,223 @@
...
+pub struct PrometheusOptions {
+    bind_address: ByteArrayRef,
+}
+
+#[derive(Clone)]
+pub struct Runtime {
+    pub(crate) core: Arc<CoreRuntime>,
+}
+
+/// If success or fail are not null, they must be manually freed when done.
```
Sushisource: References non-existent `success` field

src/Temporalio/Bridge/src/runtime.rs
```diff
@@ -0,0 +1,223 @@
...
+}
+
+#[no_mangle]
+pub extern "C" fn byte_array_free(runtime: *mut Runtime, bytes: *const ByteArray) {
+    // Bail if freeing is disabled
+    unsafe {
+        if bytes.is_null() || (*bytes).disable_free {
+            return;
+        }
+    }
```
Sushisource, line 115: Oh. Here it is, yeah, would be good to reference this from `ByteArray`'s `Drop`

src/Temporalio/Bridge/src/runtime.rs
```diff
@@ -0,0 +1,223 @@
...
+    let bytes = bytes as *mut ByteArray;
+    // Return vec back to core before dropping bytes
+    let vec = unsafe { Vec::from_raw_parts((*bytes).data as *mut u8, (*bytes).size, (*bytes).cap) };
+    // Set to null so the byte dropper doesn't try to free it
+    unsafe { (*bytes).data = std::ptr::null_mut() };
```
Sushisource, line 120: IMO this would be better accomplished by dereffing the pointer into the `ByteArray` and just dropping it. Avoids dupe and a bit more idiomatic.

src/Temporalio/Client/TemporalConnection.cs
Sushisource: Agree

Sushisource (approved):
Cool. Looking good to me!

Nothing blocking, but I think my comments on the bridge code are worth addressing, whether in this PR or the next.

src/Temporalio/Bridge/ByteArray.cs
cretz: Yeah most of the olds docs say it's for win32 handles, but it provides benefits to all unmanaged IntPtr. There was also https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.handleref which a note says is now SafeHandle. I could just implement most of https://referencesource.microsoft.com/#mscorlib/system/runtime/interopservices/safehandle.cs myself, though no value.

I will add docs to this one and I suppose tell other classes to read this class's docs. Or add a README at the root of this dir or something.

src/Temporalio/Bridge/ByteArrayRef.cs
cretz: :+1: I will sprinkle some throughout

src/Temporalio/Bridge/src/lib.rs
cretz: :+1: Will fix

src/Temporalio/Bridge/ByteArrayRef.cs
cretz: Because that limitation already exists for gRPC metadata and HTTP headers (e.g. https://github.com/grpc/grpc-go/issues/1888 and other places) and I needed some reasonable delimiter.

src/Temporalio/Bridge/src/lib.rs
cretz: See that comment

src/Temporalio/Bridge/src/lib.rs
cretz: I can add that warning. Yes, essentially `ByteArray` is Rust-owned, Vec-alloc'd byte array whereas `ByteArrayRef` is lang-owned byte array. I also left open the possibility of us reusing/pooling these at some point in the future, but this is fine for now.

src/Temporalio/Client/ITemporalConnection.cs
cretz: Yeah, I did the same in Python, there was no value in exposing raw health service.

src/Temporalio/Client/OperatorService.cs
cretz: I think it's unnecessary at this point. I can come back and add it later. These lower-level raw gRPC services do not need to be abstracted out to injectable interfaces yet IMO. I could technically make an interface for every class.

src/Temporalio/Client/RpcOptions.cs
cretz: Yes. This is how they can opt-in to retry-ing raw gRPC calls. Our high-level calls like `StartWorkflowAsync` always set this to true.

src/Temporalio/Client/RpcOptions.cs
cretz: Definitely. Will add.

src/Temporalio/Client/RpcRetryOptions.cs
cretz: We do in Python as `retry_config` at https://python.temporal.io/temporalio.client.Client.html#connect which accepts https://python.temporal.io/temporalio.service.RetryConfig.html. If a user wants to customize their retry approach for raw gRPC calls, how do they?

src/Temporalio/Client/TemporalConnection.cs
cretz: Because `IBridgeClientProvider` interface requires it which this class implements (indirectly via `ITemporalConnection`). When creating a worker, I need this `BridgeClient` handle and and a `Namespace` and that's basically it. So I have an `IWorkerClient` that returns an `IBridgeClientProvider` (which this is) and a `Namespace`, and `ITemporalClient` implements `IWorkerClient`.

I think this is a great separation between workers and clients while still allowing them to work easily together. The docs on `IBridgeClientProvider` mention that this is for internal use only. Without this, you could not have a worker accept an interface because it needs the bridge client somehow on the interface.

src/Temporalio/Client/TemporalConnectionOptions.cs
cretz: We do in Python and I think it's reasonable here. If we used the lang-native gRPC in either of those, I'd tell the user to go use the retry config for that, but since both of those SDKs use ours, this gives the flexibility needed for their retry needs on raw gRPC calls.

src/Temporalio/Client/TemporalConnectionOptions.cs
cretz: > nit: or empty

I'm gonna error on empty I think then. I don't want to accept two forms of unset strings throughout the codebase. .NET has null for this.

> What about an `else` clause?

Can you clarify? I could return early if this _is_ null, is that what you're saying?

src/Temporalio/Client/TemporalConnectionOptions.cs
cretz: `MemberwiseClone()` shallowly copies, so it won't deeply copy this field.

src/Temporalio/Client/TestService.cs
cretz: I built it this way so if you wanted to provide your own `WorkflowService` implementation with your own `InvokeRpcAsync` call that maybe wrapped the real one, you could without a connection. This is also why this call `protected` and not `internal` and why the service class is abstract and this is in a nested `Impl` class. Otherwise I wouldn't have any of this exposed, but this is reasonable from an OO POV.

src/Temporalio/Client/TestService.cs
cretz: Because we need an implementation of the abstract service. The service is abstract so that users could have their own impls of these services if they wanted to for mocks or whatever.

src/Temporalio/Runtime.cs
cretz: Yes it is intentionally hardcoded and is the only place it is. I think it also may be duplicated in the `csproj` file at publish time, but I don't think it's reliable to retrieve that at runtime from the DLL itself so I put it here. Many other SDKs have this too (needed for creating Rust client). There should be no other place in this project where this string would be, everyone should reference it from here.

src/Temporalio/Runtime.cs
cretz: They are not implemented yet. I may move this to a `Temporalio.Runtime` package so that those options can be provided.

src/Temporalio/Testing/TestServerOptions.cs
cretz: And `default`. Was ok not to mention it in Python :-) I will mention it here.

src/Temporalio/Bridge/src/client.rs
cretz: Definitely. Tired of typing these strings. Would definitely like to see a client call that accepts a string RPC call name and proto bytes and returns a proto message or its bytes.

src/Temporalio/Bridge/ByteArrayRef.cs
cretz: > Extra simple: Just use a proto rather than this custom text based format

It's not necessarily extra simple to maintain lang-specific protos for crossing the bridge. Note we use structs in many places for Rust and C# that we could just replace with protos but we choose not to.

> Slightly less wasteful, but more annoying to build/free: Make a specific byte ref type in C like this:

Yes, I could make a linked list of pairs. And if I need a `map<string, string>` that is not gRPC metadata I almost surely will. But this is specifically for metadata which has a nice no-newline property.

> Lets you get rid of the special newline handling.

But this is basically how HTTP headers and gRPC metadata are handled on the wire (though they use `:` for delimiting between key and value while I use newline, but they still use newline for pair delimiting).

> In general, I think it probably makes sense to represent these different "kinds" of byte arrays as different C structs

Usually agree, but after I typed it out, I discovered metadata was the only use for string maps and that newlines worked well as delimiters just like they do for HTTP headers. If we need general purpose string maps, I will definitely go make them. I only use newlines here because I can, not because I'm so lazy I impose unreasonable burdens on the caller.

src/Temporalio/Bridge/src/client.rs
cretz: `client_name` and `client_version` here take an `Into<String>` and `identity` here takes a `String` only so I figured I'd give each they lowest level they can accept.

src/Temporalio/Bridge/src/runtime.rs
cretz: Yup, invalid docs, will fix.

src/Temporalio/Bridge/src/runtime.rs
cretz: As in via comment? I can, sure. But yeah I don't want this logic inside of drop because "disable free" is a construct of the runtime use of this not the actual data. I still want drop to drop everything it has when it is dropped.

src/Temporalio/Bridge/src/runtime.rs
cretz: Not following exactly. Are you suggesting I `Box::from_raw` higher? I am using/updating all of the unsafe pointers before I return ownership back to Rust at the bottom.

src/Temporalio/Bridge/src/client.rs
Sushisource: It's not really about what's "lowest-level", it's about these lines implying that this is going to get borrowed when really it's getting cloned. The builder accepts into for `&'static str`, mainly, for convenience. 

It's not a big deeal, but just letting you know about a best practice.

src/Temporalio/Bridge/ByteArrayRef.cs
Sushisource: Even if it's the only use, I think the separated types add a lot of clarity while reading. They don't need to actually *do* anything, per se, they just make what each field is being used for more clear to a reader and future-proof the code a bit at a very low cost. I really think it's worth it.

We can keep the newline parsing if you don't think the other change is worth it, but I'd really advocate for the different types.

src/Temporalio/Bridge/src/runtime.rs
Sushisource: Yes, just to make it clear. I understand why it needs to happen outside.

src/Temporalio/Bridge/src/runtime.rs
Sushisource: Did a little more digging and there's a function exactly for this: 

https://doc.rust-lang.org/std/ptr/fn.drop_in_place.html

src/Temporalio/Bridge/ByteArrayRef.cs
cretz: I try to take care saying what is "metadata". To confirm, you're saying a struct with a single byte array ref in it instead of just the byte array ref directly? Byte array refs are the only way I have to pass byte arrays (and therefore strings) from C# to Rust. I'd hate to add a new byte array struct, but I can wrap it if needed.

src/Temporalio/Bridge/src/runtime.rs
cretz: Not following the change exactly. I have to extract-and-remove the vec separately from the `ByteArray` so that I can give it back to the "runtime" (just in case we pool bytes later). Can you give a more concrete suggested change?

Sushisource: Oh man, somehow I just totally missed the vec being passed back for re-use. Nevermind. :facepalm:

src/Temporalio/Bridge/ByteArrayRef.cs
Sushisource: Wrapping in C would be annoying because of the extra pointer. I think, if you're gonna keep metadata the way it is, I'd just make type aliases in C, and then little wrapper structs in Rust that use https://doc.rust-lang.org/nomicon/other-reprs.html#reprtransparent

This way you get some type safety on the Rust side in terms of using the right "type" of ref in each place, and it doesn't really cost you anything. They can implement `Deref` into a `ByteArrayRef` to avoid annoying `.0` stuff.

Alternatively, if that's too annoying, just type alias in Rust too, which at least signals some intent even if it doesn't provide safety.

src/Temporalio/Bridge/src/client.rs
bergundy: I'm curious how you know that.
I couldn't find that anywhere in the codebase of either tonic or tower:

https://github.com/search?q=repo%3Ahyperium%2Ftonic+cancel&type=code
https://github.com/search?q=repo%3Atower-rs%2Ftower+cancel&type=code

According to @Sushisource it's all the way down in [hyper]) https://github.com/hyperium/hyper/blob/bb609b227236fc960d35ea31eac0d5a31cff728b/src/client/dispatch.rs#L206) but it would be nice to confirm somehow.

src/Temporalio/Client/RpcRetryOptions.cs
bergundy: I see, this is for the client not for the worker connection.
The worker retry config should not be customizable.

src/Temporalio/Client/TemporalConnectionOptions.cs
bergundy: I thought you might want to error instead of ending up with nulls.

src/Temporalio/Client/TemporalConnectionOptions.cs
bergundy: So probably worth fixing the docstring.
```
        /// <returns>A shallow copy of these options.</returns>
```

src/Temporalio/Testing/TestServerOptions.cs
bergundy: Not sure I mentioned it in TS either TBH, only noticed it here.

src/Temporalio/Bridge/src/client.rs
cretz: I think it'd have to be an integration test in core that confirms tonic cancellation cancels the gRPC request. Not sure it's a reasonable test in .NET because I'd have to create a new gRPC server or at least a proxy to know whether the request _truly_ got cancelled.

I could do a bunch of hacky things I guess around metrics, but really you're asking about a tonic feature, so assertion should be there or nearer to there.

src/Temporalio/Client/RpcRetryOptions.cs
cretz: I am exposing what core exposes here. I would hope, like core probably does, that you wouldn't customize retry for worker clients. But if we want to make sure core worker clients don't use user-provided retry options, we can do that in core I think.

src/Temporalio/Client/TemporalConnectionOptions.cs
cretz: That's up to the caller I think. This is only used for ephemeral server creation where we're ok with no host and/or ip.

src/Temporalio/Client/TemporalConnectionOptions.cs
cretz: It is a shallow copy, just that it does copy this. For example, I don't clone the byte arrays used for client certs or the runtime or the RPC metadata (maybe I should on that one though?). This is meant to be used to set a different field value for any option. So I only clone options, not other fields. I could have just buried all of the TLS options into this object but I chose to make separate objects for those. I can try to describe that on every options class if I must.

src/Temporalio/Bridge/ByteArrayRef.cs
cretz: Taken to its logical conclusion, I should never use `ByteArrayRef`, but always have wrappers such as `ProtoByteArrayRef`, `MetadataByteArrayRef`, `Utf8ByteArrayRef`, `RawByteArrayRef`, etc. I think the documented contract between these things for internal use and living next to each other is clearer than adding types for the different kinds of byte arrays. And it doesn't help C# side anyways. I can type alias here and there I suppose (hopefully cbindgen doesn't mind).

Sushisource: Yeah I mean that's pretty much what I was suggesting, it just happens to be a bit cumbersome to do in C in a meaningfully type safe way.

The aliases at least signal intent nicely though, and cost nothing. I'd do it.

src/Temporalio/Bridge/OptionsExtensions.cs
cretz: Actually I just realized, empty is an error here. We don't want to treat empty and null the same.

src/Temporalio/Api/Dependencies/Gogo.cs
cretz: Fixed

src/Temporalio/Bridge/ByteArray.cs
cretz: Added minor docs

src/Temporalio/Bridge/src/lib.rs
cretz: Added

src/Temporalio/Client/TemporalConnection.cs
cretz: Put on interface

.gitignore
cretz: Fixed

src/Temporalio/Bridge/ByteArrayRef.cs
cretz: Done

src/Temporalio/Bridge/OptionsExtensions.cs
cretz: Fixed

src/Temporalio/Bridge/OptionsExtensions.cs
cretz: Fixed

src/Temporalio/Bridge/src/lib.rs
cretz: Fixed

src/Temporalio/Client/TemporalConnectionOptions.cs
cretz: Changed doc string a bit here.

src/Temporalio/Testing/TestServerOptions.cs
cretz: Fixed

src/Temporalio/Bridge/ByteArrayRef.cs
cretz: Added alias

src/Temporalio/Bridge/src/client.rs
cretz: Fixed

src/Temporalio/Bridge/src/runtime.rs
cretz: Fixed

src/Temporalio/Bridge/src/runtime.rs
cretz: Fixed

src/Temporalio/Client/ITemporalClient.cs
mathobbs: On this topic, recommend implementation classes (e.g. TemporalClient) are scoped to internal instead of public. A smaller public footprint means a smaller domain to learn for sdk consumers. E.g. when to use ITemporalClient vs. TemporalClient - easier to only see ITemporalClient & then the sdk can evolve the internals with less risk of breaking sdk api. Essentially interfaces and/or abstract classes & supporting types etc. would be the only public types an sdk consumer should have to see.

PR #11, opened 2023-01-11
Converters and Failure Exceptions

cretz:
## What was changed

* Add data converters, payload converters, failure converters, and payload codecs
* Add common exception types that represent proto failures

Note, it is known that not every line is covered in tests as future functional tests will cover them. Also more exceptions will be added later (e.g. workflow already started), this is just for failure exceptions.


src/Temporalio/Converters/BinaryNullConverter.cs
```diff
@@ -0,0 +1,46 @@
...
+    /// Encoding converter for null values.
+    /// </summary>
+    public class BinaryNullConverter : IEncodingConverter
+    {
+        private static readonly ByteString EncodingByteString = ByteString.CopyFromUtf8(
+            "binary/null"
+        );
+
+        /// <inheritdoc />
+        public string Encoding => "binary/null";
```
bergundy, line 17: I wish we didn't need both of these but I get that it doesn't seem worth abstracting for a handful of implementaions.

src/Temporalio/Converters/FailureConverter.cs
```diff
@@ -0,0 +1,197 @@
...
+using System.Linq;
+using Temporalio.Api.Failure.V1;
+using Temporalio.Exceptions;
+
+namespace Temporalio.Converters
+{
+    /// <summary>
+    /// Default implementation of <see cref="IFailureConverter" />.
+    /// </summary>
+    public class FailureConverter : IFailureConverter
```
bergundy: Should this be called `DefaultFailureConverter` to match the other SDKs?

src/Temporalio/Converters/FailureConverterOptions.cs
```diff
@@ -0,0 +1,25 @@
+using System;
+
+namespace Temporalio.Converters
+{
+    /// <summary>
+    /// Options for a failure converter.
+    /// </summary>
+    public class FailureConverterOptions : ICloneable
```
bergundy: This would need to be renamed `DefaultFailureConverterOptions` if you choose to rename the converter class.

src/Temporalio/Converters/IPayloadConverter.cs
```diff
@@ -0,0 +1,32 @@
...
+    /// <summary>
+    /// Representation of a converter from a value to/from a payload.
+    /// </summary>
+    /// <remarks>
+    /// This converter should be deterministic since it is used for workflows. For the same reason,
+    /// this converter should be immediate and avoid any network calls or any asynchronous/slow code
+    /// paths.
+    /// </remarks>
+    /// <seealso cref="PayloadConverter" />
+    public interface IPayloadConverter
```
bergundy, line 15: In Python (and I think Go) you chose to have PayloadConverter converter multi to multi, what made you change your mind?
I like this (.NET) way better FTR.

cretz:
I am converting this to draft while I debug native issues...

src/Temporalio/Converters/PayloadConverter.cs
```diff
@@ -0,0 +1,131 @@
...
+{
+    /// <summary>
+    /// Default implementation of a payload converter which iterates over a collection of
+    /// <see cref="IEncodingConverter" />.
+    /// </summary>
+    /// <remarks>
+    /// See <see cref="PayloadConverter()" /> for the default set of encoding converters used. To
+    /// create a custom converter, a new class should extend this one.
+    /// </remarks>
+    public class PayloadConverter : IPayloadConverter
```
bergundy: Same here, should we call it `DefaultPayloadConverter`.

src/Temporalio/Exceptions/ApplicationFailureException.cs
```diff
@@ -0,0 +1,92 @@
...
+        /// Cause of the exception (can use other constructor if no cause).
+        /// </param>
+        /// <param name="type">Optional string type name of the exception</param>
+        /// <param name="nonRetryable">If true, marks the exception as non-retryable.</param>
+        /// <param name="details">Collection of details to serialize into the exception.</param>
+        public ApplicationFailureException(
+            string message,
+            Exception? inner,
+            string? type = null,
+            bool nonRetryable = false,
```
bergundy, line 49: I'm tempted to leverage the fact that we can have default param values to use the positive form `retryable` which is much easier to parse.
The downside here would be inconsistency with UI (proto) and other SDKs.

bergundy (approved):
Mostly went over the public API, didn't go too much into the implementation or docstrings.
Overall LGTM.

src/Temporalio/Converters/BinaryNullConverter.cs
cretz: I assume by both of these you mean the static readonly and the encoding here? Yeah I struggled with this one. Technically the only way this encoding field is used by `PayloadConverter` is as a `ByteString` anyways, but I couldn't bring myself to make a `Google.Protobuf.ByteString` a type in the public API. I don't convert this to `ByteString` on each need of course, only in the `PayloadConverter` constructor. That's why I have the static here, so I can use that in the actual payload each time and avoid an alloc.

src/Temporalio/Converters/FailureConverter.cs
cretz: I don't think so. I only used `Default` in those others because of the name clash with the abstract/interface. In .NET the default impl of an interface being the interface name sans leading `I` makes the most sense.

But if desired, I can change it to match other SDKs. No big reason not to.

src/Temporalio/Converters/IPayloadConverter.cs
cretz: I brought this up on internal Slack. Here's the thought progression that happened:

Go and Java had single and multi-payload methods for its data converter. But we end up leveraging multiple for remote data converters which made multiple seem wise. We chose multiple in Python because type hints were optional and it matched Go. It just seemed right at the time to limit/batch the invocations to a user-defined thing like a payload converter.

In .NET I originally had collection to collection here with optional types. But it became clear that .NET doesn't allow optional types for JSON conversion. And really, .NET doesn't allow optional types anywhere. So like Go and Java we are going to require types. Once we made that decision, it became clear in the interest of generics that one at a time was clearest (see `ConverterExtensions` for the `ToValue` I expect most will use). After discussion, it also became clear that the arity was never allowed to change and since we don't have remote data converters or any need to have multiple at a time, we just went to one for code clarity.

I think we'd do one at a time in all SDKs if we had it to do over, but at least TypeScript, Ruby, and .NET will.

src/Temporalio/Converters/FailureConverterOptions.cs
cretz: Commented on `DefaultFailureConverter`

src/Temporalio/Converters/PayloadConverter.cs
cretz: Commented on `DefaultFailureConverter`

src/Temporalio/Exceptions/ApplicationFailureException.cs
cretz: I don't think the inconsistency here is worth it

src/Temporalio/Converters/FailureConverter.cs
cretz: After some more thought, I think I will rename these

cretz: Fixed

src/Temporalio/Converters/FailureConverterOptions.cs
cretz: Fixed

src/Temporalio/Converters/PayloadConverter.cs
cretz: Fixed

src/Temporalio/Bridge/Scope.cs
```diff
@@ -77,8 +77,12 @@ public Interop.ByteArrayRef NewlineDelimited(IEnumerable<string>? values)
...
         public IntPtr FunctionPointer<T>(T func) where T : Delegate
         {
-            toKeepAlive.Add(func);
-            return Marshal.GetFunctionPointerForDelegate(func);
+            // The delegate seems to get collected before called sometimes even if we add "func" to
+            // the keep alive list. Delegates are supposed to be reference types, but their pointers
+            // seem unstable. So we're going to alloc a handle for it. We can't pin it though.
+            var handle = GCHandle.Alloc(func);
+            toKeepAlive.Add(handle);
+            return Marshal.GetFunctionPointerForDelegate(handle.Target!);
```
cretz, line 85: This was the fix to broken CI. Apparently keeping "func" variable alive wasn't enough (though it is enough if it's a local variable).

PR #14, opened 2023-01-24
Initial workflow client

cretz:
## What was changed

* Start workflow and get workflow result
* Workflow attributes
* Tests and scaffolding for the above
* EDIT: added all the rest of the client features, this is now a full-fledged SDK client

~Note, this does not yet include all tests or all features. I am trying to get this in front of people early. I may continue to work on parts for this PR while it remains unapproved.~

EDIT: This thing is full featured and ready to go.


src/Temporalio/Client/Interceptors/ClientOutboundInterceptor.cs
```diff
@@ -0,0 +1,68 @@
...
+        /// <param name="input">Input details of the call.</param>
+        /// <returns>Handle for the workflow.</returns>
+        public virtual Task<WorkflowHandle<TResult>> StartWorkflowAsync<TResult>(
+            StartWorkflowInput input)
+        {
+            return Next.StartWorkflowAsync<TResult>(input);
+        }
+
+        // TODO(cretz): Document that this never returns an empty page while with a next page token
+        // is set (i.e. it immediately uses that token internally)
```
cretz: Can remove this, I already documented it

src/Temporalio/Client/TemporalClient.Workflow.cs
```diff
@@ -0,0 +1,266 @@
...
+                    {
+                        throw new InvalidOperationException("Unexpected raw history returned");
+                    }
+                    // Complete if we got any events or if there is no next page token
+                    if (resp.History.Events.Count > 0 || resp.NextPageToken.IsEmpty)
+                    {
+                        return new WorkflowHistoryEventPage(
+                            resp.History.Events,
+                            resp.NextPageToken.IsEmpty ? null : resp.NextPageToken.ToByteArray());
+                    }
```
cretz:
```suggestion
                    }
                    req.NextPageToken = resp.NextPageToken;
```

src/Temporalio/Converters/ConverterExtensions.cs
```diff
@@ -7,16 +12,142 @@ namespace Temporalio.Converters
...
+            this DefaultPayloadConverter converter, string key, object value)
         {
-            return (T?)converter.ToValue(payload, typeof(T));
+            if (value == null)
+            {
+                throw new ArgumentException($"Search attribute {key} has null value");
+            }
+            if (value is DateTime dateTimeValue)
+            {
+                // 8601 with convert with timezone
```
cretz:
```suggestion
                // 8601 with timezone
```

src/Temporalio/Client/Interceptors/ClientOutboundInterceptor.cs
cretz: Fixed

src/Temporalio/Client/TemporalClient.Workflow.cs
cretz: Fixed

src/Temporalio/Converters/ConverterExtensions.cs
cretz: Fixed

src/Temporalio/Bridge/src/client.rs
```diff
@@ -206,15 +211,36 @@ pub extern "C" fn client_rpc_call(
...
-            Err(err) => (
+        let (success, status_code, failure_message, failure_details) = match res {
+            Ok(b) => (
+                ByteArray::from_vec(b).into_raw(),
+                0,
+                std::ptr::null_mut(),
                 std::ptr::null_mut(),
-                ByteArray::from_utf8(format!("{}", err)).into_raw(),
             ),
+            Err(err) => match err.downcast::<tonic::Status>() {
```
Sushisource, line 221: I think the downcast could be fairly easily avoided by changing the various call methods/macros to return enum errors (use `thiserror` or w/e) rather than anyhow errors. They could even be `tonic | other`. Makes this match a bit cleaner and the code a bit more maintainable.

src/Temporalio/Client/WorkflowHandle.cs
```diff
@@ -0,0 +1,377 @@
...
+
+        /// <summary>
+        /// Get the result of a workflow, deserializing into the given return type.
+        /// </summary>
+        /// <typeparam name="TResult">Return type to deserialize result into.</typeparam>
+        /// <param name="followRuns">
+        /// Whether to follow runs until the latest workflow is reached.
+        /// </param>
+        /// <param name="rpcOptions">RPC options.</param>
+        /// <returns>Untyped task for waiting on result.</returns>
```
Sushisource: This returns a typed task, no?

src/Temporalio/Client/WorkflowHandle.cs
```diff
@@ -0,0 +1,377 @@
...
+        /// <exception cref="WorkflowFailureException">
+        /// Exception thrown for unsuccessful workflow result. The cause can be
+        /// <see cref="CancelledFailureException" />, <see cref="TerminatedFailureException" />,
+        /// <see cref="TimeoutFailureException" />, or any exception deserialized that was thrown in
+        /// the workflow (usually an <see cref="ApplicationFailureException" />).
+        /// </exception>
+        public async Task<TResult> GetResultAsync<TResult>(
+            bool followRuns = true, RpcOptions? rpcOptions = null)
+        {
+            // Continually get pages
```
Sushisource, line 82: It's a little sad that we reimplement pagination in every SDK. This could all be encapsulated in Core if we want (the code is there already), but I suppose we lose out on some low level interception (which I doubt anyone cares much about).

WDYT?

src/Temporalio/Client/WorkflowHandle.cs
```diff
@@ -0,0 +1,377 @@
...
+            WorkflowHandle(Client, ID, RunID, ResultRunID, FirstExecutionRunID)
+    {
+        /// <summary>
+        /// Get the result of a workflow, deserializing into the known result type.
+        /// </summary>
+        /// <param name="followRuns">
+        /// Whether to follow runs until the latest workflow is reached.
+        /// </param>
+        /// <param name="rpcOptions">RPC options.</param>
+        /// <returns>Untyped task for waiting on result.</returns>
```
Sushisource: Also typed

src/Temporalio/Exceptions/RpcException.cs
```diff
@@ -0,0 +1,152 @@
...
+                }
+                return Api.Common.V1.GrpcStatus.Parser.ParseFrom(rawStatus);
+            });
+        }
+
+        /// <summary>
+        /// gRPC status code taken from
+        /// https://github.com/grpc/grpc-dotnet/blob/master/src/Grpc.Core.Api/StatusCode.cs.
+        /// </summary>
+        public enum StatusCode
```
Sushisource, line 36: Why copied rather than re-used?

src/Temporalio/Exceptions/WorkflowQueryRejectedException.cs
```diff
@@ -0,0 +1,25 @@
+using Temporalio.Api.Enums.V1;
+
+namespace Temporalio.Exceptions
+{
+    /// <summary>
+    /// Thrown when a query is rejected by the worker due to bad workflow status.
```
Sushisource: Worth explicitly distinguishing the difference between this and `WorkflowQueryFailedException` in the docstring I think

src/Temporalio/Workflow/WorkflowAttribute.cs
```diff
@@ -0,0 +1,246 @@
...
+                if (runMethod == null)
+                {
+                    errs.Add($"{type} does not have a valid WorkflowRun method");
+                }
+
+                // If there are any errors, throw
+                if (errs.Count > 0)
+                {
+                    // TODO(cretz): Ok to use aggregate exception here or should I just
+                    // comma-delimit into a single message or something?
```
Sushisource, line 219: Seems like that's what it's there for

Sushisource (approved):
Looking pretty good! Nothing blocking

src/Temporalio/Bridge/src/client.rs
cretz: While it does make the match a bit cleaner, that my macros can return any error like inability to decode protos does make that code cleaner (though yes I could have a `NotTonic(anyhow::Error)`).

src/Temporalio/Client/WorkflowHandle.cs
cretz: It does, will fix

src/Temporalio/Client/WorkflowHandle.cs
cretz: Usually I would concur, however I am about to write the history event async iterator which needs to control paging anyways so there'd be no value here. I wrote the interceptor and single-page fetching the way I did here with the expectation of reuse.

src/Temporalio/Client/WorkflowHandle.cs
cretz: Thanks, will fix.

src/Temporalio/Exceptions/RpcException.cs
cretz: I don't want to add a dependency on the .NET gRPC library just for this

src/Temporalio/Exceptions/WorkflowQueryRejectedException.cs
cretz: Will do! Also noticed a typo here that may help explain the difference too

```suggestion
    /// Thrown when a query is rejected by the server due to bad workflow status.
```

src/Temporalio/Workflow/WorkflowAttribute.cs
cretz: Kinda, but it's not "AggregateMessageException", it has separate exceptions carrying their own stack traces, causes, etc. But I agree, probably good enough, and bonus is the outputters know how to format this one well.

src/Temporalio/Bridge/src/client.rs
Sushisource: The macros should end up with the ability to work the same if you use the auto-From implementations generated by `thiserror`

tests/Temporalio.Tests/Client/WorkflowHandleTests.cs
```diff
@@ -0,0 +1,126 @@
...
+
+        // Describe
+        var desc = await handle.DescribeAsync();
+
+        // Check values
+        var beforeNow = DateTime.UtcNow.AddSeconds(-30);
+        var afterNow = DateTime.UtcNow.AddSeconds(30);
+        var temp = desc.CloseTime!.Value;
+        Assert.InRange(desc.CloseTime!.Value, beforeNow, afterNow);
+        Assert.InRange(desc.ExecutionTime!.Value, beforeNow, afterNow);
```
jakejscott, line 175: @cretz  Not sure if you want to add another dependency, but Fluent Assertions library is alot nicer to use and gives better output when assertions fail.

It has a helpers that let you test a date is `CloseTo` some time and you pass it a TimeSpan of how close
https://fluentassertions.com/datetimespans/ 

The object graph comparision also comes in handy
https://fluentassertions.com/objectgraphs/

Also useful is the ability to async retry something and still use assertions in the body (which gives nice errors when it fails)...
```
Func<Task> asyncRetry = async () =>
{
    var describeExecutionResponse = await _stepFunctions.DescribeExecutionAsync(new DescribeExecutionRequest
    {
        ExecutionArn = executionArn
    });
    describeExecutionResponse.Should().NotBeNull();
    describeExecutionResponse.Status.Should().Be(ExecutionStatus.SUCCEEDED);
};
await asyncRetry.Should().NotThrowAfterAsync(waitTime: 60.Seconds(), pollInterval: 2.Seconds());
```

cretz: As the tests grow I very well may add that dependency or other test helpers (I don't mind adding dependencies to our test project). I haven't needed too much yet (and I have started an `AssertMore` package at the root of the project for a couple of them).

src/Temporalio/Client/WorkflowHandle.cs
cretz: Fixed

src/Temporalio/Client/WorkflowHandle.cs
cretz: Fixed

cretz:
This PR has had as much review as it's going to receive I'm afraid. Merging.

PR #18, opened 2023-02-06
Activity worker

cretz:
## What was changed

* Added activity worker and tests
* Added async activity client
* Added activity test environment
* Added new code analysis analyzers and made several fixes based on what they found
* README updates


src/Temporalio/Activity/ActivityCancelReason.cs
```diff
@@ -0,0 +1,40 @@
...
+        /// </summary>
+        CancelRequested,
+
+        /// <summary>
+        /// Activity timeout caused activity to be marked cancelled.
+        /// </summary>
+        Timeout,
+
+        /// <summary>
+        /// Worker the activity is running is shutting down.
```
cretz:
```suggestion
        /// Worker the activity is running on is shutting down.
```

src/Temporalio/Activity/ActivityContext.cs
```diff
@@ -0,0 +1,122 @@
...
+        /// <summary>
+        /// Gets a value indicating whether the current code is running in an activity.
+        /// </summary>
+        public static bool HasCurrent => AsyncLocalCurrent.Value != null;
+
+        /// <summary>
+        /// Gets the current activity context.
+        /// </summary>
+        /// <exception cref="InvalidOperationException">If no context is available.</exception>
+        public static ActivityContext Current => AsyncLocalCurrent.Value ??
```
cretz, line 48: Getting kind of tired typing `ActivityContext.Current` everywhere in my activity code. I am worried it will get worse with workflows typing `WorkflowContext.Current` everywhere. I have considered changing everything on here to be static (e.g. `ActivityContext.Heartbeat()` instead of `ActivityContext.Current.Heartbeat`) but it harms people's ability to pass the activity context around if they wanted (maybe `IAsyncLocal` isn't enough for them?). Other situations will allow you to have an attributed param like `[ActivityContext] IActivityContext` and have it injected, but that would mess with our param typing.

So this is kinda ugly but it's the .NET way. That's why people have to type `MyAsyncLocalVar.Value.Whatever()` and `MyLazyVar.Value.Whatever()`.

src/Temporalio/Activity/CompleteAsyncException.cs
```diff
@@ -0,0 +1,16 @@
+namespace Temporalio.Activity
+{
+    /// <summary>
+    /// Exception the user should throw to inform the worker that this activity will be completed
+    /// asynchronously.
+    /// </summary>
+    public class CompleteAsyncException : Exceptions.TemporalException
```
cretz, line 7: I considered putting this in `Temporalio.Exceptions`, but it's a control-flow exception, so it is the only one I am justifying leaving out of that namespace.

src/Temporalio/Worker/ActivityWorker.cs
```diff
@@ -0,0 +1,703 @@
...
+                    () => ExecuteActivityAsync(act, tsk)).Unwrap();
+#pragma warning restore CA2008
+            }
+        }
+
+        private async Task ExecuteActivityAsync(
+            RunningActivity act,
+            Bridge.Api.ActivityTask.ActivityTask tsk)
+        {
+            // Try to build a completion, but if it fails the manually build one with the failure
```
cretz:
```suggestion
            // Try to build a completion, but if it fails then manually build one with the failure
```

src/Temporalio/Worker/ActivityWorker.cs
```diff
@@ -0,0 +1,703 @@
...
+                    },
+                };
+            }
+
+            // Try to finish heartbeats and send completion
+            try
+            {
+                // We have to wait on any outstanding heartbeats to finish. This will not throw. We
+                // accept that in a rare scenario, this heartbeat can fail to encode but it is too
+                // late to cancel the activity. Like other SDKs, we current drop this heartbeat.
```
cretz:
```suggestion
                // late to cancel the activity. Like other SDKs, we currently drop this heartbeat.
```

src/Temporalio/Worker/ActivityWorker.cs
```diff
@@ -0,0 +1,703 @@
...
+                                    details).ConfigureAwait(false));
+                        }
+                        worker.BridgeWorker.RecordActivityHeartbeat(heartbeat);
+                    }
+                }
+                catch (Exception e)
+                {
+                    using (Context.Logger.BeginScope(Context.Info.LoggerScope))
+                    {
+                        // Log exception on done (nowhere to can propagate), warning and cancel if
```
cretz:
```suggestion
                        // Log exception on done (nowhere to propagate), warning + cancel if
```

src/Temporalio/Worker/TemporalWorker.cs
```diff
@@ -0,0 +1,287 @@
...
+        /// Run this worker until failure, cancelled, or task from given function completes.
+        /// </summary>
+        /// <remarks>
+        /// When shutting down, the worker will cancel and wait for completion for all executing
+        /// activities. If an activity does not properly respond to cancellation, this may never
+        /// return.
+        /// </remarks>
+        /// <param name="untilComplete">
+        /// If the task returned from this function completes, the worker will shutdown
+        /// (propagating) exception as necessary.
```
cretz:
```suggestion
        /// propagating exception as necessary.
```

src/Temporalio/Worker/TemporalWorker.cs
```diff
@@ -0,0 +1,287 @@
...
+                    logger.LogInformation("User task completed, shutting down");
+                }
+                else if (task.Exception != null)
+                {
+                    logger.LogError(task.Exception, "Worker failed, shutting down");
+                }
+            }
+
+            // Now wait for all of the tasks to complete. This will collect exceptions from them all
+            // and throw them as aggregate.
```
cretz: Move this to just above the `try` below

src/Temporalio/Worker/TemporalWorker.cs
```diff
@@ -0,0 +1,287 @@
...
+
+            // If there is an activity worker, we want to add a graceful shutdown task for it
+            if (activityWorker != null)
+            {
+                tasks.Add(activityWorker.GracefulShutdownAsync());
+            }
+            try
+            {
+                // This is essentially guaranteed to throw since it will contain one of the tasks
+                // from the WhenAny that threw
```
cretz: > guaranteed to throw

Not true since it could be a user-defined task that just completed successfully

src/Temporalio/Activity/ActivityCancelReason.cs
cretz: Fixed

src/Temporalio/Worker/ActivityWorker.cs
cretz: Fixed

src/Temporalio/Worker/ActivityWorker.cs
cretz: Fixed

src/Temporalio/Worker/ActivityWorker.cs
cretz: Fixed

src/Temporalio/Worker/TemporalWorker.cs
cretz: Fixed

src/Temporalio/Worker/TemporalWorker.cs
cretz: Fixed

src/Temporalio/Worker/TemporalWorker.cs
cretz: Fixed

.editorconfig
```diff
@@ -4,6 +4,45 @@
 
 # Please keep in alphabetical order by field.
 
+# We cannot reasonably mark ourselves CLSCompliant one way or another if
```
bergundy, line 7: General comments for this file. Are there rules that we care about and would rather disable locally?

README.md
```diff
@@ -136,7 +136,7 @@ public record GreetingInfo(string Salutation = "Hello", string Name = "<unknown>
 [Workflow]
 public interface IGreetingWorkflow
 {
-    static readonly IGreetingWorkflow Ref = Refs<IGreetingWorkflow>.Instance;
+    static readonly IGreetingWorkflow Ref = Refs.Create<IGreetingWorkflow>();
```
bergundy, line 296: I like this, it's reads nicer.

src/Temporalio/Activity/ActivityAttribute.cs
```diff
@@ -0,0 +1,111 @@
+using System;
+using System.Collections.Concurrent;
+using System.Reflection;
+using System.Threading.Tasks;
+
+namespace Temporalio.Activity
+{
+    /// <summary>
+    /// Designate a method as a workflow.
```
bergundy:
```suggestion
    /// Designate a method as an activity.
```

src/Temporalio/Activity/ActivityCancelReason.cs
```diff
@@ -0,0 +1,40 @@
...
+{
+    /// <summary>
+    /// Reason for why an activity may be cancelled.
+    /// </summary>
+    public enum ActivityCancelReason
+    {
+        /// <summary>
+        /// The activity has not been cancelled or was cancelled for an unknown reason.
+        /// </summary>
+        Unknown,
```
bergundy: I don't think you need this, when will the reason be unknown?

src/Temporalio/Activity/ActivityContext.cs
```diff
@@ -0,0 +1,122 @@
...
+        /// <summary>
+        /// Gets the logger scoped to this activity.
+        /// </summary>
+        public ILogger Logger { get; private init; }
+
+        /// <summary>
+        /// Gets why the activity was cancelled. This value is inaccurate until
+        /// <see cref="CancellationToken" /> is cancelled.
+        /// </summary>
+        public ActivityCancelReason CancelReason => CancelReasonRef.CancelReason;
```
bergundy, line 65: I'd prefer if this was nullable or we renamed `Unknown` to `NotSet` or `Unset`.

src/Temporalio/Activity/ActivityCancelReason.cs
bergundy: See below where I mentioned renaming this variant.

src/Temporalio/Activity/ActivityContext.cs
```diff
@@ -0,0 +1,122 @@
...
+        /// </summary>
+        /// <remarks>
+        /// Heartbeats should be used for all non-immediately-returning, non-local activities and
+        /// they are required to receive cancellation. Heartbeats are queued and processed
+        /// asynchronously, so this will not error if the details cannot be converted. Rather any
+        /// error converting heartbeat details will result in activity cancellation then activity
+        /// failure.
+        /// </remarks>
+        /// <param name="details">Details to record with the heartbeat if any.</param>
+        public void Heartbeat(params object?[] details)
```
bergundy, line 118: Could we use varargs here to avoid the need for wrapping details in an array?

bergundy: I generally (in all SDKs) don't get why this is an array in the first place but I guess it's fine since it matches our loose requirements for method signatures.

src/Temporalio/Activity/ActivityContext.cs
```diff
@@ -0,0 +1,122 @@
...
+        {
+            if (Heartbeater == null)
+            {
+                throw new InvalidOperationException("Heartbeater not set yet");
+            }
```
bergundy, line 123: Can't you set the heartbeater on init?

src/Temporalio/Activity/ActivityContext.cs
```diff
@@ -0,0 +1,122 @@
...
+        /// </summary>
+        internal ByteString TaskToken { get; private init; }
+
+        /// <summary>
+        /// Record a heartbeat on the activity.
+        /// </summary>
+        /// <remarks>
+        /// Heartbeats should be used for all non-immediately-returning, non-local activities and
+        /// they are required to receive cancellation. Heartbeats are queued and processed
+        /// asynchronously, so this will not error if the details cannot be converted. Rather any
```
bergundy, line 108: Mention throttling too please.

src/Temporalio/Bridge/Worker.cs
```diff
@@ -0,0 +1,335 @@
...
+
+        /// <summary>
+        /// Shutdown this worker.
+        /// </summary>
+        /// <returns>Completion task.</returns>
+        public async Task ShutdownAsync()
+        {
+            using (var scope = new Scope())
+            {
+                var completion = new TaskCompletionSource<bool>();
```
bergundy, line 283: Why does this need to be a bool instead of empty (void) task?

src/Temporalio/Bridge/Worker.cs
```diff
@@ -0,0 +1,335 @@
...
+                unsafe
+                {
+                    Interop.Methods.worker_shutdown(
```
bergundy, line 286: No need to call wait for shutdown, that's done in `finalize_shutdown` anyways. There a sync `initiate_shutdown` method that's more suitable here.

src/Temporalio/Bridge/src/worker.rs
```diff
@@ -0,0 +1,402 @@
...
+        Ok(config) => match temporal_sdk_core::init_worker(
+            &client.runtime.core,
+            config,
+            client.core.clone().into_inner(),
+        ) {
+            Err(err) => (
+                std::ptr::null_mut(),
+                client
+                    .runtime
+                    .alloc_utf8(&format!("Worker start failed: {}", err))
```
bergundy, line 84: How are you going to distinguish between invalid options and worker start failed later?
Looks like worker creation failure is always an invalid operation exception.

src/Temporalio/Client/AsyncActivityHandle.cs
```diff
@@ -0,0 +1,94 @@
+using System;
+using System.Threading.Tasks;
+
+namespace Temporalio.Client
+{
+    /// <summary>
+    /// Async activity handle to perform activity actions for activities that will complete
+    /// asynchronously.
```
bergundy:
```suggestion
    /// Handle to perform activity actions for activities that will complete asynchronously.
```

src/Temporalio/Client/AsyncActivityHandle.cs
```diff
@@ -0,0 +1,94 @@
...
+    public record AsyncActivityHandle(
+        ITemporalClient Client, AsyncActivityHandle.Reference Activity)
+    {
+        /// <summary>
+        /// Issue a heartbeat for this activity.
+        /// </summary>
+        /// <param name="options">Heartbeat options.</param>
+        /// <returns>Completion task.</returns>
+        /// <exception cref="Exceptions.AsyncActivityCancelledException">
+        /// If the server has requested that this activity be cancelled. Users should catch this and
```
bergundy, line 20:
```suggestion
        /// If the workflow has requested that this activity be cancelled. Users should catch this and
```

src/Temporalio/Client/AsyncActivityHandle.cs
```diff
@@ -0,0 +1,94 @@
...
+            private protected Reference()
+            {
+            }
+        }
+
+        /// <summary>
+        /// Reference to an activity by its workflow ID, workflow run ID, and activity ID.
+        /// </summary>
+        /// <param name="WorkflowID">ID for the activity's workflow.</param>
+        /// <param name="RunID">Run ID for the activity's workflow.</param>
```
bergundy, line 82: This is optional.

src/Temporalio/Client/Interceptors/IClientInterceptor.cs
```diff
@@ -8,15 +8,15 @@ public interface IClientInterceptor
         /// <summary>
         /// Create a client outbound interceptor to intercept calls.
         /// </summary>
-        /// <param name="next">The next interceptor in the chain to call.</param>
+        /// <param name="nextInterceptor">The next interceptor in the chain to call.</param>
```
bergundy, line 11: nit: Why?

src/Temporalio/Exceptions/AsyncActivityCancelledException.cs
```diff
@@ -0,0 +1,17 @@
+namespace Temporalio.Exceptions
+{
+    /// <summary>
+    /// Exception thrown from <see cref="Client.AsyncActivityHandle.HeartbeatAsync" /> if server has
+    /// requested that an activity be cancelled.
```
bergundy:
```suggestion
    /// Exception thrown from <see cref="Client.AsyncActivityHandle.HeartbeatAsync" /> if workflow has
    /// requested that an activity be cancelled.
```

bergundy: Or is this for any reason that activity should stop heartbeating?

src/Temporalio/Testing/ActivityEnvironment.cs
```diff
@@ -0,0 +1,159 @@
...
+        /// </summary>
+        public ActivityInfo Info { get; init; } = DefaultInfo;
+
+        /// <summary>
+        /// Gets or inits the logger. Default is a no-op logger.
+        /// </summary>
+        public ILogger Logger { get; init; } = NullLogger.Instance;
+
+        /// <summary>
+        /// Gets or sets the cancel reason. Callers may prefer <see cref="Cancel" /> instead.
```
bergundy, line 52: Why expose this in the first place instead of exposing `Cancel`?

src/Temporalio/Testing/ActivityEnvironment.cs
```diff
@@ -0,0 +1,159 @@
...
+        /// If cancellation not already requested, set the cancel reason and cancel the token
+        /// source.
+        /// </summary>
+        /// <param name="reason">Cancel reason.</param>
+        public void Cancel(ActivityCancelReason reason = ActivityCancelReason.CancelRequested)
+        {
+            // This is intentionally not an atomic operation same as it's not in the real worker.
+            // It is documented for callers not to expect reason to be valid until cancellation
+            // token is set.
+            if (!CancellationTokenSource.IsCancellationRequested)
```
bergundy, line 154: I'm fine with this method being reentrant and having subsequent calls ignored but it should be documented behavior.

src/Temporalio/Worker/TemporalWorker.cs
```diff
@@ -0,0 +1,288 @@
...
+        /// </param>
+        /// <param name="stoppingToken">Cancellation token to stop the worker.</param>
+        /// <returns>
+        /// When the task is complete, the worker has completed shutdown.
+        /// </returns>
+        /// <exception cref="InvalidOperationException">Already started.</exception>
+        /// <exception cref="OperationCanceledException">Cancellation requested.</exception>
+        /// <exception cref="Exception">Fatal worker failure.</exception>
+        public Task ExecuteAsync(
+            Func<Task> untilComplete, CancellationToken stoppingToken = default)
```
bergundy, line 133: I find accepting both of stopping token and untilComplete function confusing.
If a user wants to use a token they can use it in the `untilComplete` function.

src/Temporalio/Worker/TemporalWorker.cs
```diff
@@ -0,0 +1,288 @@
...
+            {
+#pragma warning disable CA1031 // Intentionally swallow all exceptions to finalizing shutdown
+                try
+                {
+                    await BridgeWorker.FinalizeShutdownAsync().ConfigureAwait(false);
+                }
+                catch (Exception e)
+                {
+                    // Ignore finalization errors, worker will be dropped Rust side anyways
+                    logger.LogDebug(e, "Worker finalization failed");
```
bergundy: I'd increase the level at least to warning.

src/Temporalio/Worker/TemporalWorkerOptions.cs
```diff
@@ -0,0 +1,131 @@
...
+        /// <param name="taskQueue">Task queue for the worker.</param>
+        public TemporalWorkerOptions(string taskQueue)
+        {
+            TaskQueue = taskQueue;
+        }
+
+        /// <summary>
+        /// Gets or sets the task queue for the worker.
+        /// </summary>
+        public string? TaskQueue { get; set; }
```
bergundy, line 32: This shouldn't be optional.

src/Temporalio/Worker/TemporalWorkerOptions.cs
```diff
@@ -0,0 +1,131 @@
...
+        /// Gets or sets the maximum number of activities that will ever be given to this worker
+        /// concurrently.
+        /// </summary>
+        public int MaxConcurrentActivities { get; set; } = 100;
+
+        /// <summary>
+        /// Gets or sets the maximum number of concurrent poll activity task requests we will
+        /// perform at a time on this worker's task queue.
+        /// </summary>
+        public int MaxConcurrentActivityTaskPolls { get; set; } = 5;
```
bergundy: There's only going to be one with the current implementation, you might as we not expose this at all until this is fixed or we implement poller autoscaling which would make this obsolete.

tests/Temporalio.Tests/Worker/ActivityWorkerTests.cs
```diff
@@ -0,0 +1,718 @@
...
+        var actErr = Assert.IsType<ActivityFailureException>(wfErr.InnerException);
+        Assert.IsType<ApplicationFailureException>(actErr.InnerException);
+    }
+
+    [Fact]
+    public async Task ExecuteActivityAsync_ThrowsOperationCancelled_ReportsFailure()
+    {
+        // Just to confirm there's not an issue with this exception
+        [Activity]
+        static void Throws() => throw new OperationCanceledException();
```
bergundy, line 257: Why is this test required, isn't it a duplicate of `ExecuteActivityAsync_ThrowCommonException_ReportsFailure`?
Are we testing what happens when there's no error message?

tests/Temporalio.Tests/Worker/ActivityWorkerTests.cs
```diff
@@ -0,0 +1,718 @@
...
+        static string SomeActivity2() => string.Empty;
+        var err = Assert.Throws<ArgumentException>(() =>
+        {
+            using var worker = new TemporalWorker(Client, new()
+            {
+                TaskQueue = $"tq-{Guid.NewGuid()}",
+                Activities = { SomeActivity1, SomeActivity2 },
+            });
+        });
+        Assert.Contains("same key has already been added", err.Message);
```
bergundy: Not sure what the entire message looks like but can we check that it mentions that it's a duplicate activity registration error and it says which activities are conflicting?

bergundy (commented):
Submitting already even though I haven't read the ActivityWorker implementation yet.
Awesome work, I didn't have too many comments.

.editorconfig
cretz: You mean like a developer disabling rules on their workstation that we'd want on for the repo? No, these rules apply to the code itself and need to match for everyone so they can know what we/CI expect.

Locally I go and comment out `<TreatWarningsAsErrors>true</TreatWarningsAsErrors>` in `Directory.Build.props` sometimes while I work just so I can tell compile failures from these static analyzer ones.

README.md
cretz: Concur. I was also getting https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1000 when I added these new analyzers which made it clear that the previous approach didn't read right.

src/Temporalio/Activity/ActivityCancelReason.cs
cretz: > when will the reason be unknown?

Unknown when either given unknown from core or when the activity is never cancelled (but we cancel the token anyways after the activity returns so if they sent this context off somewhere else, this will remain unknown post activity completion though the cancel token will be done).

src/Temporalio/Activity/ActivityContext.cs
cretz: :+1: Will change to `None` per https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/enum. I originally didn't want to use `None` because it implies that it wasn't cancelled, but rereading it just says there's no reason.

src/Temporalio/Activity/ActivityContext.cs
cretz: > Could we use varargs here to avoid the need for wrapping details in an array?

This is varargs (that's what the `params` keyword means there)

src/Temporalio/Activity/ActivityContext.cs
cretz: No, from the docs on `Heartbeater`:

> Gets or sets the heartbeater. This is late-bound since interceptors that need this context may be invoked before the interceptor is created that ends up providing the heartbeater.

So basically on "intercept activity" and "init" outbound calls I make this context available without heartbeating. Then I set the outbound interceptor's heartbeat as the heartbeater before running the activity.

src/Temporalio/Bridge/Worker.cs
cretz: There is no such thing as `void` type arg in .NET so `TaskCompletionSource<void>` isn't a thing. `TaskCompletionSource` sans type is a thing and is what I would use, but it is newer than our oldest supported .NET versions, ref https://learn.microsoft.com/en-us/dotnet/api/system.threading.tasks.taskcompletionsource#applies-to.

src/Temporalio/Bridge/Worker.cs
cretz: For the way I have things ordered, I don't want this to return until pollers are shutdown. Finalize I have as a separate thing just for reclaiming the arc.

src/Temporalio/Bridge/src/worker.rs
cretz: I didn't know there was a difference. If it can start it can't start form the user POV. Regardless, like client RPC call exceptions that I recently made more robust, when a need arises I'll abstract and differentiate here too.

src/Temporalio/Client/AsyncActivityHandle.cs
cretz: This is to differentiate from server and worker-shutdown cancellation. The server issues the cancellation e.g. on timeout. Saying that's done by the workflow may make users think it's explicitly done from the workflow.

src/Temporalio/Client/AsyncActivityHandle.cs
cretz: TIL (the API protos don't make this clear). I will mark optional, though I hope they always set it.

src/Temporalio/Client/Interceptors/IClientInterceptor.cs
cretz: This is a fun one. This tripped https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1716 and after digging I found https://github.com/dotnet/roslyn-analyzers/blob/1dc431ec828e9cc816fc69f89ad9b8286e74d707/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldNotMatchKeywords.cs#L597 and realized it's a keyword in VB.Net for https://learn.microsoft.com/en-us/dotnet/visual-basic/language-reference/statements/for-next-statement.

src/Temporalio/Exceptions/AsyncActivityCancelledException.cs
cretz: This is for when https://github.com/temporalio/api/blob/e4246bbd59fd1f850bdd5be6a59d6d2f8e532d76/temporal/api/workflowservice/v1/request_response.proto#L415-L417 comes back true. But it's the server that is telling us it's cancelled, e.g. for timeouts, not the workflow. Not sure why the proto docs say "the activity has been asked to cancel itself" though.

src/Temporalio/Testing/ActivityEnvironment.cs
cretz: This is a test environment and since they can be read independently I allow them to be set independently too.

.editorconfig
bergundy: I meant locally as in per location exception of these rules.

src/Temporalio/Bridge/Worker.cs
bergundy: Yeah, I know there's no `void` type arg, I was referring to `TaskCompletionSource` but I understand that it's not supported in all .NET versions.

.editorconfig
cretz: Ah, you'll see I do that with pragmas sometimes. These I globally disable are specifically because I disagree with them at the project level for all code. For cases where I need local overrides (i.e. it's a good rule but I have to violate for some special reason), I do that very specifically. I comment here and in those places and can provide details on any specific rule decision if needed.

src/Temporalio/Testing/ActivityEnvironment.cs
cretz: The comment is not about reentrancy as much as just why I don't do these steps atomically. The user should not care. In .NET it's normal for a second cancel not to do anything and the user facing doc on this call says it won't.

src/Temporalio/Worker/TemporalWorker.cs
cretz: Cancellation is different than a task. I'd be more likely to remove until complete than I'd remove stopping token. Stopping token matches what .NET does in many other cases and I'll always support that to match https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.hosting.backgroundservice.executeasync?view=dotnet-plat-ext-7.0#microsoft-extensions-hosting-backgroundservice-executeasync(system-threading-cancellationtoken) (can't take dependency on that though). I only have these overloads as convenience and regardless of task convenience we always want to allow a cancellation token. But if they are that confusing I can remove them.

src/Temporalio/Worker/TemporalWorkerOptions.cs
cretz: It has to be, as is every required option without a default. This is the same for host port for client connection even though I will fail at runtime if not set. It is the .NET options pattern approach to allow all of these to be late bound instead of required on init (and using placeholders like an empty string is unclear in that it makes one think it has a default and that default is acceptable).

src/Temporalio/Worker/TemporalWorkerOptions.cs
cretz: If there a general purpose issue where we're deprecating this or setting it to one in other core SDKs?

src/Temporalio/Bridge/Worker.cs
bergundy: I'll read the activity worker implementation and reply

tests/Temporalio.Tests/Worker/ActivityWorkerTests.cs
cretz: From the comment:

> // Just to confirm there's not an issue with this exception

In Python I had an issue where I caught the language native cancelled on a non cancelling scenario and tried to send cancel back to server only to have server fail saying cancel was never requested. This is just a sanity test to confirm I don't have that mistake here.

tests/Temporalio.Tests/Worker/ActivityWorkerTests.cs
cretz: I admit being lazy and just letting https://learn.microsoft.com/en-us/dotnet/api/system.linq.enumerable.todictionary?view=net-7.0 throw which does say what is duplicated but not that it's activities. Can improve.

src/Temporalio/Bridge/src/worker.rs
bergundy: I'd throw the proper exception type here.
Generally you're going to need a mechanism to pass around the error types from Rust to C#, it's something worth doing.

src/Temporalio/Client/AsyncActivityHandle.cs
bergundy: Oh, so you'll get an exception for timeout too, makes sense.

src/Temporalio/Client/AsyncActivityHandle.cs
bergundy: I just learned this lately too but it makes sense since both workflow ID and activity ID can be set by the user and run ID can't.

src/Temporalio/Exceptions/AsyncActivityCancelledException.cs
bergundy: Oh so that's only if the workflow has requested activity cancellation.

bergundy: So the docstring change still applies IMO.

src/Temporalio/Testing/ActivityEnvironment.cs
bergundy: I think it's unnecessary, having a single way to cancel is enough IMHO.
But I leave the decision up to you.

src/Temporalio/Bridge/src/worker.rs
cretz: Haven't needed this in Python. What case does the user need specific exceptions here? Invalid operation covers all cases until there is a compelling reason for a more specific exception type for a more specific situation.

src/Temporalio/Worker/TemporalWorkerOptions.cs
bergundy: I didn't expose this in TS: https://typescript.temporal.io/api/interfaces/worker.WorkerOptions

tests/Temporalio.Tests/Worker/ActivityWorkerTests.cs
bergundy: Okay, maybe worth a more detailed comment if you want.

src/Temporalio/Testing/ActivityEnvironment.cs
bergundy: True

src/Temporalio/Worker/TemporalWorker.cs
bergundy: I mean for this specific overload.
I'd have `ExecuteAsync(untilComplete)` and `ExecuteAsync(stoppingToken)` but not `ExecuteAsync(untilComplete, stoppingToken)`.

src/Temporalio/Worker/ActivityWorker.cs
```diff
@@ -0,0 +1,703 @@
...
+        /// </returns>
+        public async Task GracefulShutdownAsync()
+        {
+            using (logger.BeginScope(new Dictionary<string, object>()
+            {
+                ["TaskQueue"] = worker.Options.TaskQueue!,
+            }))
+            {
+                workerShutdownTokenSource.Cancel();
+                logger.LogInformation(
```
bergundy, line 121: Maybe log before cancel?

src/Temporalio/Worker/ActivityWorker.cs
```diff
@@ -0,0 +1,703 @@
...
+                };
+            }
+
+            // Try to finish heartbeats and send completion
+            try
+            {
+                // We have to wait on any outstanding heartbeats to finish. This will not throw. We
+                // accept that in a rare scenario, this heartbeat can fail to encode but it is too
+                // late to cancel the activity. Like other SDKs, we currently drop this heartbeat.
+                await act.FinishHeartbeatsAsync().ConfigureAwait(false);
```
bergundy: There's no need to wait for heartbeats to complete to send successful completion, only failure.

bergundy: And not for cancellation either.

src/Temporalio/Worker/ActivityWorker.cs
```diff
@@ -0,0 +1,703 @@
...
+                // Remove from running activities
+                runningActivities.TryRemove(tsk.TaskToken, out _);
+            }
+        }
+
+        private async Task<Bridge.Api.ActivityTaskCompletion> ExecuteActivityInternalAsync(
+            RunningActivity act,
+            Bridge.Api.ActivityTask.ActivityTask tsk)
+        {
+            act.Context.Logger.LogDebug(
```
bergundy, line 282: Doesn't this run in the activity info logger scope? If so, you should have the activity type in the scope.

src/Temporalio/Worker/ActivityWorker.cs
```diff
@@ -0,0 +1,703 @@
...
+                        " but more than that are required by the signature");
+                }
+                // Zip the params and input and then decode each. It is intentional that we discard
+                // extra input arguments that the signature doesn't accept.
+                var paramVals = new List<object?>(tsk.Start.Input.Count);
+                try
+                {
+                    foreach (var (input, paramInfo) in tsk.Start.Input.Zip(paramInfos, (a, b) => (a, b)))
+                    {
+                        var paramVal = await worker.Client.Options.DataConverter.ToValueAsync(
```
bergundy: Prefer to do this concurrently for all params.

src/Temporalio/Worker/ActivityWorker.cs
```diff
@@ -0,0 +1,703 @@
...
+                var result = await inbound.ExecuteActivityAsync(new(
+                    Delegate: defn.Delegate,
+                    Parameters: paramVals.ToArray(),
+                    Headers: tsk.Start.HeaderFields)).ConfigureAwait(false);
+
+                completion.Result.Completed = new();
+                // As a special case, ValueTuple is considered "void"
+                if (result is not ValueTuple)
+                {
+                    completion.Result.Completed.Result =
```
bergundy, line 356: AFAIR, we decided that "null" result should end up as a null payload even though the behavior is inconsistent in the other SDKs.

src/Temporalio/Worker/ActivityWorker.cs
```diff
@@ -0,0 +1,703 @@
...
+                act.Context.Logger.LogWarning(
+                    "Completing activity {ActivityType} as failure during heartbeat",
+                    act.Context.Info.ActivityType);
+                completion.Result.Failed = new()
+                {
+                    Failure_ = await worker.Client.Options.DataConverter.ToFailureAsync(
+                        act.HeartbeatFailureException).ConfigureAwait(false),
+                };
+            }
+            catch (Exception e)
```
bergundy, line 390: Should we have a case for when activity throws due to worker shutting down? At least for a nicer log message and possibly info log level.

src/Temporalio/Worker/ActivityWorker.cs
```diff
@@ -0,0 +1,703 @@
...
+            {
+                workerShutdownTokenSource.Cancel();
+                logger.LogInformation(
+                    "Beginning activity worker shutdown, will wait {GracefulShutdownTimeout} before " +
+                    "cancelling {ActivityCount} activity instance(s)",
+                    worker.Options.GracefulShutdownTimeout,
+                    runningActivities.Count);
+                await Task.Delay(worker.Options.GracefulShutdownTimeout).ConfigureAwait(false);
+                // Issue cancel-all and wait for all running activities to complete
+                await Task.WhenAll(runningActivities.Values.Select(act =>
```
bergundy, line 129: You should wait for poll to return shutdown error to ensure that no activities are added to this dictionary before this line.

Looks like there's no guarantee here: https://github.com/temporalio/sdk-dotnet/pull/18/files/b1eb2174cda30c4110f0dde22a800fe4d7636535#diff-49d90da0465bf6446c00276cd2c7e7f5e4832c01ddd1c426a6af81b1b689ff53R260

src/Temporalio/Worker/TemporalWorker.cs
cretz: Every one of these async calls needs to accept a cancellation token. It's the .NET way. There are plenty of .NET APIs that accept a task and a cancellation token. I've tried to allow a cancellation token on every async call I expose to the user throughout this SDK (I may have missed one or two on test environments). Cancellation token is general threading and not trivially converted to a task anyways, though you can do it via a wait handle or infinite delay. By this logic https://learn.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.waitany?view=net-7.0#system-threading-tasks-task-waitany(system-threading-tasks-task()-system-threading-cancellationtoken) overload wouldn't exist.

src/Temporalio/Bridge/Worker.cs
bergundy: I read it.
I like that the activity shutdown sequence is contained in the activity worker and doesn't make assumptions about when it is called.
I'd prefer to initiate shutdown externally in TemporalWorker and wait for poller shutdown and completion of all activities in the ActivityWorker.

cretz: >  I'd prefer to initiate shutdown externally in TemporalWorker and wait for poller shutdown and completion of all activities in the ActivityWorker.

I do initiate shutdown in the TemporalWorker, but yes it is the activity worker caller that waits for poller shutdown and activity completion.

src/Temporalio/Exceptions/AsyncActivityCancelledException.cs
cretz: That boolean would not be set to true for an activity start-to-close or heartbeat timeout?

src/Temporalio/Worker/TemporalWorkerOptions.cs
cretz: Ok, will remove here (EDIT: Removed). Should I make an issue to deprecate and set this to 1 in Python by default?

src/Temporalio/Worker/ActivityWorker.cs
cretz: Cancel is a synchronous call so probably doesn't _really_ matter, but will update

src/Temporalio/Worker/ActivityWorker.cs
cretz: To be a proper steward of .NET and to properly ensure my Rust instance only has 1 worker Arc reference remaining, I need to wait for all tasks to finish before finalizing shutdown of the worker. This is just waiting for the converters to finish and make the Rust call, so it's mostly harmless.

src/Temporalio/Worker/ActivityWorker.cs
cretz: Showing scope in .NET is opt-in and many loggers don't even support it. That's why for non-internal-error type logs of an activity, I also qualify the activity a bit in the log string.

src/Temporalio/Worker/ActivityWorker.cs
cretz: A null result will end up as a null payload, but this is a "void" which while it can't be differentiated in TS and Python from null/undefined/None, it can in Java and .NET. This does remind me that I need to fix this to explicitly check `ValueTuple` and not any derivatives.

src/Temporalio/Worker/ActivityWorker.cs
cretz: I do say how many activities I am force cancelling on worker shutdown in that log message, but I don't think we need a separate catch clause here for "cancelled failure + cancel reason is worker shutdown". These catch clauses are not for specific error messages (though they have them), they are for different behavior.

src/Temporalio/Worker/ActivityWorker.cs
cretz: Yes, I mention at the top of this call that it should not be called until "ExecuteAsync" is done but I didn't ensure that from the caller. Fixing.

src/Temporalio/Activity/ActivityAttribute.cs
cretz: Fixed

src/Temporalio/Activity/ActivityContext.cs
cretz: Fixed

src/Temporalio/Activity/ActivityContext.cs
cretz: Fixed

src/Temporalio/Client/AsyncActivityHandle.cs
cretz: Fixed

src/Temporalio/Client/AsyncActivityHandle.cs
cretz: Fixed

src/Temporalio/Worker/TemporalWorker.cs
cretz: Fixed

tests/Temporalio.Tests/Worker/ActivityWorkerTests.cs
cretz: More detail added

tests/Temporalio.Tests/Worker/ActivityWorkerTests.cs
cretz: Fixed

src/Temporalio/Worker/ActivityWorker.cs
cretz: Fixed

src/Temporalio/Worker/ActivityWorker.cs
cretz: Fixed

src/Temporalio/Worker/ActivityWorker.cs
cretz: Fixed

src/Temporalio/Exceptions/AsyncActivityCancelledException.cs
bergundy: AFAIK you'll get a NotFound error from the server in those cases.

bergundy: Or if the workflow has already completed.
Please test and confirm.

src/Temporalio/Worker/TemporalWorker.cs
bergundy: I guess it doesn't hurt to accept an optional token.
Just the exact behavior needs to be very clear, will untilComplete block the task returned by ExecuteAsync if the cancellation token is triggered first?

src/Temporalio/Worker/TemporalWorkerOptions.cs
bergundy: I would be nice for anyone accessing this property to know that it has to be set but I figure that this probably isn't doable in .NET.

src/Temporalio/Worker/TemporalWorkerOptions.cs
bergundy: Just let Core do its thing.
Don't deprecate for now, I hope to revisit the core poller implementation soon.

src/Temporalio/Worker/ActivityWorker.cs
bergundy: You're blocking completion from reaching the server which is more valuable than a heartbeat.

src/Temporalio/Worker/ActivityWorker.cs
bergundy: Got it

src/Temporalio/Worker/ActivityWorker.cs
bergundy: I think it's still worth changing the logging in this case to reduce the chance of user confusion.

src/Temporalio/Worker/TemporalWorker.cs
cretz: No, it's whichever happens first - workflow poll fail, activity poll fail, cancel token cancelled, or user task completes. Whichever happens causes shutdown process to start.

src/Temporalio/Worker/TemporalWorkerOptions.cs
cretz: Only via docs and strong hints by having it in a constructor if I'm following options pattern. I really wish I could use records and required init properties. The .NET options guidelines don't say much on this, see https://learn.microsoft.com/en-us/dotnet/core/extensions/options-library-authors. But the ASP.NET guidelines at https://learn.microsoft.com/en-us/aspnet/core/fundamentals/configuration/options do state:

> Must be non-abstract with a public parameterless constructor.
> All public read-write properties of the type are bound.

Mostly due to JSON deserialization that needs to be able to construct the object before deserializing.

src/Temporalio/Worker/ActivityWorker.cs
cretz: It's valuable for tracking of async tasks and worker refs. This same thing exists for Python. We ensure every Rust call for an activity is completed before activity completion so I don't have unaccounted for Rust work holding a ref to the worker. I know activity complete means no more Rust calls. I do shortcut pending heartbeat though from running after done.

src/Temporalio/Worker/ActivityWorker.cs
cretz: Not following. It says why it failed in the log. The cancellation token was cancelled.

bergundy: Yeah, warning is fine here.

src/Temporalio/Worker/ActivityWorker.cs
bergundy: I'm not sure Java makes that distinction but I'm fine with this behavior for .NET.

src/Temporalio/Worker/ActivityWorker.cs
bergundy: Sure, but you can await this task after flushing the completion and prevent the call into core if the activity is already done.

bergundy: If you want to take it a step further (which I don't think is necessary), you can cancel the heartbeat task too.

src/Temporalio/Worker/TemporalWorkerOptions.cs
Sushisource: @bergundy I've mentioned a few times this isn't actually the case. It does, in fact, spin up multiple pollers. I just ran a test and looked at metrics:

```
temporal_num_pollers{namespace="default",poller_type="sticky_workflow_task",service_name="temporal-core-sdk",task_queue="workflow_load_YfXY5vL2"} 8
temporal_num_pollers{namespace="default",poller_type="workflow_task",service_name="temporal-core-sdk",task_queue="workflow_load_YfXY5vL2"} 2
```

There's something wrong with activity pollers always reporting 0 for the metric value right now, not sure that that's about, but I verified just throwing in a debug statement that activity pollers also got as high as 8 with a 10 value.

So, this _does in fact do something_. We might not expose it for now anyway, but there's not always 1 poller.

src/Temporalio/Bridge/Worker.cs
cretz: When I changed from `tasks.Add(Task.Run(BridgeWorker.ShutdownAsync, CancellationToken.None));` to `BridgeWorker.InitiateShutdown();` (which calls worker `initiate_shutdown`), my `ExecuteActivityAsync_MaxConcurrent_TimesOutIfMore` test fails. This is because once I hit max concurrent and the workflow fails timing out while scheduling the next activity, I cancel the running of the worker (i.e. user-defined task completes). But for whatever reason, Core is not sending a `PollActivityError::ShutDown` to my poller. But with `shutdown` it properly does. `initiate_shutdown` also seems to properly send that error to my poller in most other cases, but not all.

Can you test and confirm this scenario? I want the poller to be told it is done by Core on shutdown.

src/Temporalio/Worker/ActivityWorker.cs
cretz: > Sure, but you can await this task after flushing the completion and prevent the call into core if the activity is already done.

Only in cases of success which complicates the code. I have added this complication (glad I didn't have to in Python). I question whether it makes a real difference (only in slow codecs on heartbeat + complete).

src/Temporalio/Exceptions/AsyncActivityCancelledException.cs
cretz: Added test and confirmed it is not-found RPC error. Changed class doc.

src/Temporalio/Worker/TemporalWorkerOptions.cs
bergundy: IIUC, we may see more than 1 concurrent activity poller under certain circumstances but this isn't really intentional and can only happen if there are local activities or activity cancellations to be dispatched while there's an outstanding poll request.

The poll buffer guarantees that we do not poll unless a poll is requested:

https://github.com/temporalio/sdk-core/blob/17ceab7952fc67674b624183b62eae16ecd12824/core/src/pollers/poll_buffer.rs#L130

https://github.com/temporalio/sdk-core/blob/17ceab7952fc67674b624183b62eae16ecd12824/core/src/pollers/poll_buffer.rs#L82

We add a poll permit every time we reach this line:

https://github.com/temporalio/sdk-core/blob/17ceab7952fc67674b624183b62eae16ecd12824/core/src/worker/activities.rs#L197

So if we break out of the select statement after adding a poll permit, the next poll request might add an additional permit.

https://github.com/temporalio/sdk-core/blob/17ceab7952fc67674b624183b62eae16ecd12824/core/src/worker/activities.rs#L200

My point is that this mechanism needs a better design and the fact that we sometimes get more than one poller is not the way it should work.

I hope I'm not terribly misinterpreting the code.

bergundy (approved):

src/Temporalio/Worker/TemporalWorkerOptions.cs
Sushisource: Yes it's totally true that it's nonobvious why it works, but essentially it works because a permit is added to the polls requested sempahore every time lang calls, but it is not used every time, because the poll may return early because of buffered work.

So, yeah, it should be changed for sure, but let's stop saying it only ever polls one at a time b/c that's not accurate either.

src/Temporalio/Bridge/src/worker.rs
bergundy: I just noticed I never replied to this.
My ask here was that if worker init fails for example because it failed to connect, this would result in something other than invalid operation because it's not.

cretz: In .NET invalid operation is general purpose for most things (https://learn.microsoft.com/en-us/dotnet/standard/exceptions/best-practices-for-exceptions#use-the-predefined-net-exception-types). Also, workflow construction is synchronous, so is it really connecting? We use the same general purpose exception to represent all workflow construction failures in Python and I figure we should here too unless there's a good reason not to.

bergundy: You're good, I double checked. I think this used to connect which is what had me confused.

PR #29, opened 2023-03-02
Initial workflow runtime

cretz:
## What was changed

Contains the initial workflow runtime with no features. Features will be added in the next branch after this. This is only open early for reviewers. Most suggestions will be added in the next branch.

### DIFF
--- a/src/Temporalio/Worker/TemporalWorker.cs
+++ b/src/Temporalio/Worker/TemporalWorker.cs
@@ -15,6 +15,8 @@ namespace Temporalio.Worker
     public class TemporalWorker : IDisposable
     {
         private readonly ActivityWorker? activityWorker;
+        private readonly WorkflowWorker? workflowWorker;
+        private readonly bool workflowTaskTracingEnabled;
         private int started;
 
         /// <summary>
@@ -38,10 +40,13 @@ namespace Temporalio.Worker
             {
                 activityWorker = new(this);
             }
-            else
+            else if (options.Workflows.Count == 0)
             {
-                // TODO(cretz): Remove requirement when workflows are impl'd
-                throw new ArgumentException("No activities present");
+                throw new ArgumentException("Must have at least one workflow and/or activity");
+            }
+            if (options.Workflows.Count > 0)
+            {
+                workflowWorker = new(this);
             }
 
             // Interceptors are the client interceptors that implement IWorkerInterceptor followed
@@ -51,6 +56,29 @@ namespace Temporalio.Worker
             if (Options.Interceptors != null)
             {
                 Interceptors = Interceptors.Concat(Options.Interceptors);
+            }
+            // Extract workflow interceptor constructors out
+            var expectedTypes = new Type[] { typeof(WorkflowInboundInterceptor) };
+            WorkflowInboundInterceptorTypes = Interceptors.Select(
+                i =>
+                {
+                    var type = i.WorkflowInboundInterceptorType;
+                    if (type == null)
+                    {
+                        return null;
+                    }
+                    else if (type.GetConstructor(expectedTypes) == null)
+                    {
+                        throw new InvalidOperationException($"Workflow interceptor {type} missing constructor accepting inbound");
+                    }
+                    return type;
+                }).OfType<Type>();
+
+            // Enable workflow task tracing if needed
+            workflowTaskTracingEnabled = !options.DisableWorkflowTaskTracing && options.Workflows.Count > 0;
+            if (workflowTaskTracingEnabled)
+            {
+                WorkflowTaskEventListener.Instance.Register();
             }
         }
 
@@ -84,6 +112,11 @@ namespace Temporalio.Worker
         /// Gets the set of interceptors in the order they should be applied.
         /// </summary>
         internal IEnumerable<IWorkerInterceptor> Interceptors { get; private init; }
+
+        /// <summary>
+        /// Gets the set of workflow inbound interceptor types to create for each workflow instance.
+        /// </summary>
+        internal IEnumerable<Type> WorkflowInboundInterceptorTypes { get; private init; }
 
         /// <summary>
         /// Run this worker until failure or cancelled.
@@ -182,6 +215,11 @@ namespace Temporalio.Worker
             {
                 activityWorker?.Dispose();
                 BridgeWorker.Dispose();
+                // Remove task tracing if not disabled and there are workflows present
+                if (workflowTaskTracingEnabled)
+                {
+                    WorkflowTaskEventListener.Instance.Unregister();
+                }
             }
         }
 
@@ -216,7 +254,12 @@ namespace Temporalio.Worker
                 tasks.Add(activityWorkerTask);
                 pollTasks.Add(activityWorkerTask);
             }
-            // TODO(cretz): Workflows
+            if (workflowWorker != null)
+            {
+                var workflowWorkerTask = workflowWorker.ExecuteAsync();
+                tasks.Add(workflowWorkerTask);
+                pollTasks.Add(workflowWorkerTask);
+            }
 
             // Wait until any of the tasks complete including cancellation
             var task = await Task.WhenAny(tasks).ConfigureAwait(false);
--- a/src/Temporalio/Worker/TemporalWorkerOptions.cs
+++ b/src/Temporalio/Worker/TemporalWorkerOptions.cs
@@ -41,6 +41,11 @@ namespace Temporalio.Worker
         /// Gets or sets the activity delegates.
         /// </summary>
         public IList<Delegate> Activities { get; set; } = new List<Delegate>();
+
+        /// <summary>
+        /// Gets or sets the workflow types.
+        /// </summary>
+        public IList<Type> Workflows { get; set; } = new List<Type>();
 #pragma warning restore CA2227
 
         /// <summary>
@@ -103,6 +108,70 @@ namespace Temporalio.Worker
         public TimeSpan GracefulShutdownTimeout { get; set; } = TimeSpan.Zero;
 
         /// <summary>
+        /// Gets or sets the number of workflows cached for sticky task queue use. If this is 0,
+        /// sticky task queues are disabled and no caching occurs.
+        /// </summary>
+        public int MaxCachedWorkflows { get; set; } = 10000;
+
+        /// <summary>
+        /// Gets or sets the maximum allowed number of workflow tasks that will ever be given to
+        /// the worker at one time.
+        /// </summary>
+        public int MaxConcurrentWorkflowTasks { get; set; } = 100;
+
+        /// <summary>
+        /// Gets or sets the maximum number of local activities that will ever be given to this
+        /// worker concurrently.
+        /// </summary>
+        public int MaxConcurrentLocalActivities { get; set; } = 100;
+
+        /// <summary>
+        /// Gets or sets a value indicating whether the worker will only handle workflows and local
+        /// activities.
+        /// </summary>
+        public bool DisableRemoteActivities { get; set; }
+
+        /// <summary>
+        /// Gets or sets how long a workflow task is allowed to sit on the sticky queue before it is
+        /// timed out and moved to the non-sticky queue where it may be picked up by any worker.
+        /// </summary>
+        public TimeSpan StickyQueueScheduleToStartTimeout { get; set; } = TimeSpan.FromSeconds(10);
+
+        /// <summary>
+        /// Gets or sets a value indicating whether deadlock detection will be disabled for all
+        /// workflows. If unset, this value defaults to true only if the <c>TEMPORAL_DEBUG</c>
+        /// environment variable is <c>true</c> or <c>1</c>.
+        /// </summary>
+        /// <remarks>
+        /// When false, the default, deadlock detection prevents workflow tasks from taking too long
+        /// before yielding back to Temporal. This is undesirable when stepping through code, so
+        /// this should be set to true in those cases.
+        /// </remarks>
+        public bool DebugMode { get; set; } =
+            string.Equals(DebugModeEnvironmentVariable, "true", StringComparison.OrdinalIgnoreCase) ||
+            DebugModeEnvironmentVariable == "1";
+
+        /// <summary>
+        /// Gets or sets a value indicating whether workflow task tracing will be disabled for all
+        /// workflows.
+        /// </summary>
+        /// <remarks>
+        /// When false, the default, a <see cref="System.Diagnostics.Tracing.EventListener" /> is
+        /// used to catch improper use of tasks outside of the built-in task scheduler.
+        /// </remarks>
+        public bool DisableWorkflowTaskTracing { get; set; }
+
+        /// <summary>
+        /// Gets or sets a function to create workflow instances.
+        /// </summary>
+        /// <remarks>
+        /// Don't expose this until there's a use case.
+        /// </remarks>
+        internal Func<WorkflowInstanceDetails, IWorkflowInstance>? WorkflowInstanceFactory { get; set; }
+
+        private static string? DebugModeEnvironmentVariable { get; } = Environment.GetEnvironmentVariable("TEMPORAL_DEBUG");
+
+        /// <summary>
         /// Add the given delegate as an activity.
         /// </summary>
         /// <param name="del">Delegate to add.</param>
@@ -110,6 +179,17 @@ namespace Temporalio.Worker
         public TemporalWorkerOptions AddActivity(Delegate del)
         {
             Activities.Add(del);
+            return this;
+        }
+
+        /// <summary>
+        /// Add the given type as a workflow.
+        /// </summary>
+        /// <param name="type">Type to add.</param>
+        /// <returns>This options instance for chaining.</returns>
+        public TemporalWorkerOptions AddWorkflow(Type type)
+        {
+            Workflows.Add(type);
             return this;
         }
 
--- a/src/Temporalio/Worker/WorkflowCodecHelper.cs
+++ b/src/Temporalio/Worker/WorkflowCodecHelper.cs
@@ -1,0 +1,279 @@
+using System.Linq;
+using System.Threading.Tasks;
+using Google.Protobuf.Collections;
+using Temporalio.Api.Common.V1;
+using Temporalio.Bridge.Api.ActivityResult;
+using Temporalio.Bridge.Api.ChildWorkflow;
+using Temporalio.Bridge.Api.WorkflowActivation;
+using Temporalio.Bridge.Api.WorkflowCommands;
+using Temporalio.Bridge.Api.WorkflowCompletion;
+using Temporalio.Converters;
+
+namespace Temporalio.Worker
+{
+    /// <summary>
+    /// Utilities to encode/decode workflow protobufs.
+    /// </summary>
+    internal static class WorkflowCodecHelper
+    {
+        /// <summary>
+        /// Encode the completion.
+        /// </summary>
+        /// <param name="codec">Codec to use.</param>
+        /// <param name="comp">Completion to encode.</param>
+        /// <returns>Task for completion.</returns>
+        internal static async Task EncodeAsync(
+            IPayloadCodec codec, WorkflowActivationCompletion comp)
+        {
+            switch (comp.StatusCase)
+            {
+                case WorkflowActivationCompletion.StatusOneofCase.Failed:
+                    if (comp.Failed.Failure_ != null)
+                    {
+                        await codec.EncodeFailureAsync(comp.Failed.Failure_).ConfigureAwait(false);
+                    }
+                    break;
+                case WorkflowActivationCompletion.StatusOneofCase.Successful:
+                    foreach (var cmd in comp.Successful.Commands)
+                    {
+                        await EncodeAsync(codec, cmd).ConfigureAwait(false);
+                    }
+                    break;
+            }
+        }
+
+        /// <summary>
+        /// Decode the activation.
+        /// </summary>
+        /// <param name="codec">Codec to use.</param>
+        /// <param name="act">Activation to decode.</param>
+        /// <returns>Task for completion.</returns>
+        internal static async Task DecodeAsync(IPayloadCodec codec, WorkflowActivation act)
+        {
+            foreach (var job in act.Jobs)
+            {
+                switch (job.VariantCase)
+                {
+                    case WorkflowActivationJob.VariantOneofCase.CancelWorkflow:
+                        await DecodeAsync(codec, job.CancelWorkflow.Details).ConfigureAwait(false);
+                        break;
+                    case WorkflowActivationJob.VariantOneofCase.QueryWorkflow:
+                        await DecodeAsync(codec, job.QueryWorkflow.Arguments).ConfigureAwait(false);
+                        break;
+                    case WorkflowActivationJob.VariantOneofCase.ResolveActivity:
+                        await DecodeAsync(codec, job.ResolveActivity.Result).ConfigureAwait(false);
+                        break;
+                    case WorkflowActivationJob.VariantOneofCase.ResolveChildWorkflowExecution:
+                        await DecodeAsync(
+                            codec, job.ResolveChildWorkflowExecution.Result).ConfigureAwait(false);
+                        break;
+                    case WorkflowActivationJob.VariantOneofCase.ResolveChildWorkflowExecutionStart:
+                        if (job.ResolveChildWorkflowExecutionStart.Cancelled != null
+                            && job.ResolveChildWorkflowExecutionStart.Cancelled.Failure != null)
+                        {
+                            await codec.DecodeFailureAsync(
+                                job.ResolveChildWorkflowExecutionStart.Cancelled.Failure).
+                                ConfigureAwait(false);
+                        }
+                        break;
+                    case WorkflowActivationJob.VariantOneofCase.ResolveRequestCancelExternalWorkflow:
+                        if (job.ResolveRequestCancelExternalWorkflow.Failure != null)
+                        {
+                            await codec.DecodeFailureAsync(
+                                job.ResolveRequestCancelExternalWorkflow.Failure).
+                                ConfigureAwait(false);
+                        }
+                        break;
+                    case WorkflowActivationJob.VariantOneofCase.ResolveSignalExternalWorkflow:
+                        if (job.ResolveSignalExternalWorkflow.Failure != null)
+                        {
+                            await codec.DecodeFailureAsync(
+                                job.ResolveSignalExternalWorkflow.Failure).
+                                ConfigureAwait(false);
+                        }
+                        break;
+                    case WorkflowActivationJob.VariantOneofCase.SignalWorkflow:
+                        await DecodeAsync(codec, job.SignalWorkflow.Input).ConfigureAwait(false);
+                        break;
+                    case WorkflowActivationJob.VariantOneofCase.StartWorkflow:
+                        await DecodeAsync(codec, job.StartWorkflow).ConfigureAwait(false);
+                        break;
+                }
+            }
+        }
+
+        private static async Task EncodeAsync(IPayloadCodec codec, WorkflowCommand cmd)
+        {
+            switch (cmd.VariantCase)
+            {
+                case WorkflowCommand.VariantOneofCase.CompleteWorkflowExecution:
+                    if (cmd.CompleteWorkflowExecution.Result != null)
+                    {
+                        await EncodeAsync(
+                            codec, cmd.CompleteWorkflowExecution.Result).ConfigureAwait(false);
+                    }
+                    break;
+                case WorkflowCommand.VariantOneofCase.ContinueAsNewWorkflowExecution:
+                    await EncodeAsync(
+                        codec, cmd.ContinueAsNewWorkflowExecution.Arguments).ConfigureAwait(false);
+                    foreach (var val in cmd.ContinueAsNewWorkflowExecution.Memo.Values)
+                    {
+                        if (val != null)
+                        {
+                            await EncodeAsync(codec, val).ConfigureAwait(false);
+                        }
+                    }
+                    break;
+                case WorkflowCommand.VariantOneofCase.FailWorkflowExecution:
+                    if (cmd.FailWorkflowExecution.Failure != null)
+                    {
+                        await codec.EncodeFailureAsync(
+                            cmd.FailWorkflowExecution.Failure).ConfigureAwait(false);
+                    }
+                    break;
+                case WorkflowCommand.VariantOneofCase.RespondToQuery:
+                    if (cmd.RespondToQuery.Failed != null)
+                    {
+                        await codec.EncodeFailureAsync(
+                            cmd.RespondToQuery.Failed).ConfigureAwait(false);
+                    }
+                    else if (cmd.RespondToQuery.Succeeded?.Response != null)
+                    {
+                        await EncodeAsync(
+                            codec, cmd.RespondToQuery.Succeeded.Response).ConfigureAwait(false);
+                    }
+                    break;
+                case WorkflowCommand.VariantOneofCase.ScheduleActivity:
+                    await EncodeAsync(codec, cmd.ScheduleActivity.Arguments).ConfigureAwait(false);
+                    break;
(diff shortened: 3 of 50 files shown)


src/Temporalio/Worker/WorkflowInstance.cs
```diff
@@ -0,0 +1,503 @@
...
+        }
+
+        /// <inheritdoc/>
+        public WorkflowActivationCompletion Activate(WorkflowActivation act)
+        {
+            using (Logger.BeginScope(Info.LoggerScope))
+            {
+                completion = new() { RunId = act.RunId, Successful = new() };
+                currentActivationException = null;
+                // We need to sort jobs
```
cretz, line 194: Clarify in comment that we intend to be in order of: patch, signal, non-query, query

src/Temporalio/Worker/WorkflowInstance.cs
```diff
@@ -0,0 +1,503 @@
...
+            {
+                // Have to unwrap and re-throw target invocation exception if present
+                Task resultTask;
+                try
+                {
+                    resultTask = (Task)input.RunMethod.Invoke(input.Instance, input.Parameters)!;
+                }
+                catch (TargetInvocationException e)
+                {
+                    throw e.InnerException!;
```
cretz, line 482:
```suggestion
                    ExceptionDispatchInfo.Capture(e.InnerException!).Throw();
```
Properly keep stack trace.

src/Temporalio/Worker/WorkflowWorker.cs
```diff
@@ -0,0 +1,254 @@
...
+using Temporalio.Workflows;
+
+namespace Temporalio.Worker
+{
+    /// <summary>
+    /// Worker for workflows.
+    /// </summary>
+    internal class WorkflowWorker
+    {
+        private static readonly TimeSpan DefaultDeadlockTimeout = TimeSpan.FromSeconds(2);
```
cretz, line 22: I set this to 2 in Python because it's just a slow language, but can probably set to 1 in .NET like Go/Java. Java technically allows this to be configured, but other SDKs don't and I like "on or off".

src/Temporalio/Worker/WorkflowWorker.cs
```diff
@@ -0,0 +1,254 @@
...
+
+        private readonly ILogger logger;
+        // Keyed by run ID
+        private readonly ConcurrentDictionary<string, IWorkflowInstance> runningWorkflows = new();
+        // Keyed by run ID
+        private readonly ConcurrentDictionary<string, Task> deadlockedWorkflows = new();
```
cretz, line 28: Document somewhere how we're handling deadlocked workflows.

What we're doing here to prevent deadlocked workflows from eating up more thread pool space is failing the WFT when deadlocked, and then adding here. So we still proceed with cache eviction, but any future WFTs are responded to immediately with a deadlock without even recreating the workflow. This entry is removed once the task finally completes (if ever).

@Sushisource, @bergundy suggested that instead of this approach (evicting and just eagerly failing new activations), I should just never send back completion for eviction activation while deadlocked. Think that's a better approach? It'll properly consume a cache task slot right? And is there any harm to core in an eviction not being handled immediately but hanging (possibly forever)? I assume shutdown will not send a poll shutdown error ending polling until all outstanding remove-from-cache's are handled?

src/Temporalio/Worker/WorkflowWorker.cs
```diff
@@ -0,0 +1,254 @@
...
+            {
+                var defn = WorkflowAttribute.Definition.FromType(workflow);
+                if (WorkflowDefinitions.ContainsKey(defn.Name))
+                {
+                    throw new ArgumentException($"Duplicate workflow named {defn.Name}");
+                }
+                WorkflowDefinitions[defn.Name] = defn;
+            }
+            deadlockTimeout = worker.Options.DebugMode ? DefaultDeadlockTimeout : Timeout.InfiniteTimeSpan;
+            instanceFactory = worker.Options.WorkflowInstanceFactory ??
```
cretz, line 51: We don't technically need this instance factory concept, but I am trying to keep the serializable/sandbox separation alive internally in .NET even though we'll never use a sandbox most likely.

src/Temporalio/Worker/WorkflowWorker.cs
```diff
@@ -0,0 +1,254 @@
...
+                    // If it's null, we're done
+                    if (act == null)
+                    {
+                        break;
+                    }
+                    // Otherwise handle and put on task set
+                    // TODO(cretz): Any reason for users to need to customize factory here?
+                    var task = Task.Run(() => HandleActivationAsync(codec, act)).
+                        ContinueWith(task => { tasks.TryRemove(task, out _); }, TaskScheduler.Current);
+                    tasks[task] = null;
```
cretz, line 90: This isn't right, the `ContinueWith` task is different than the one we're removing. We need to store the original task in the dictionary, _then_ add the continue with.

src/Temporalio/Worker/WorkflowWorker.cs
```diff
@@ -0,0 +1,254 @@
...
+                    // TODO(cretz): Any reason for users to need to customize factory here?
+                    // TODO(cretz): Since deadlocks can cause unreclaimed threads, we are setting
+                    // LongRunning here. That means they may oversubscribe and create a new thread
+                    // just for this, is that ok?
+                    var workflowTask = Task.Factory.StartNew(
+                        () => workflow.Activate(act),
+                        default,
+                        TaskCreationOptions.LongRunning | TaskCreationOptions.DenyChildAttach | TaskCreationOptions.PreferFairness,
+                        TaskScheduler.Current);
+                    var timeoutTask = Task.Delay(deadlockTimeout);
```
cretz, line 146: Create this with a cancel token that we cancel when it didn't timeout. Otherwise this task will complete later unnecessarily (see https://devblogs.microsoft.com/oldnewthing/20220505-00/?p=106585).

I can't find a much cheaper way to timeout a task than create a delayed task for every single task run :-( https://learn.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.waitasync is only supported in newer .NET versions. Looking [at the source](https://github.com/dotnet/runtime/blob/dfb3d24cec0a227817420de0ae95f45b210e7c25/src/libraries/System.Private.CoreLib/src/System/Threading/Tasks/Task.cs#L2798-L2816) it basically does what Delay does (timer queue timer) though maybe a couple fewer steps. I will only move to it if microbenchmarks confirm necessity.

I should at least at the above in a comment here.

tests/Temporalio.Tests/Worker/WorkflowWorkerTests.cs
```diff
@@ -0,0 +1,183 @@
...
+    {
+        using var worker = new TemporalWorker(Client, new()
+        {
+            TaskQueue = $"tq-{Guid.NewGuid()}",
+            Workflows = { typeof(TWf) },
+        });
+        await worker.ExecuteAsync(() => action(worker));
+    }
+
+    // TODO(cretz): Tests needed:
```
cretz, line 131: This is what I consider the full set of tests to confirm the workflow runtime is complete for this SDK. Feel free to suggest any I have missed (or ask for clarification on any if needed).

cretz (commented):
Ignore failing CI, it is due to https://github.com/dotnet/format/issues/1800 and I don't want to disable code style in build yet.

src/Temporalio/Worker/WorkflowInstance.cs
```diff
@@ -0,0 +1,503 @@
...
+                    var paramInfos = details.Definition.RunMethod.GetParameters();
+                    if (details.Start.Arguments.Count < paramInfos.Length &&
+                        !paramInfos[details.Start.Arguments.Count].HasDefaultValue)
+                    {
+                        throw new InvalidOperationException(
+                            $"Workflow {Info.WorkflowType} given {details.Start.Arguments.Count} parameter(s)," +
+                            " but more than that are required by the signature");
+                    }
+                    // Zip the params and input and then decode each. It is intentional that we discard
+                    // extra input arguments that the signature doesn't accept.
```
Sushisource, line 139: Probably worth at least logging something in that instance

src/Temporalio/Worker/WorkflowInstance.cs
```diff
@@ -0,0 +1,503 @@
...
+                        completion.Failed = new()
+                        {
+                            Failure_ = new() { Message = $"Failed converting activation exception: {inner}" },
+                        };
+                    }
+                }
+
+                // Remove any non-query commands after terminal commands
+                if (completion.Successful != null)
+                {
```
Sushisource, line 261: You shouldn't have to do this - why don't you open an issue in core and I can handle this there and then we can get rid of this code.

src/Temporalio/Worker/WorkflowInstance.cs
```diff
@@ -0,0 +1,503 @@
...
+                // TODO(cretz): Check conditions
+            }
+        }
+
+        private void AddCommand(WorkflowCommand cmd)
+        {
+            if (completion == null)
+            {
+                throw new InvalidOperationException("No completion available");
+            }
```
Sushisource, line 362: Seems like this state could be made unrepresentable by having completion be passed through the relevant functions instead of as a class field?

src/Temporalio/Worker/WorkflowWorker.cs
Sushisource: You're allowed to not complete a task forever, and the slot will be occupied. Shutdown will never resolve until all outstanding work is replied to.

Sushisource (commented):
Looks like it's making sense to me :+1:

src/Temporalio/Worker/TemporalWorker.cs
```diff
@@ -52,6 +57,29 @@ public TemporalWorker(IWorkerClient client, TemporalWorkerOptions options)
...
+                    }
+                    else if (type.GetConstructor(expectedTypes) == null)
+                    {
+                        throw new InvalidOperationException($"Workflow interceptor {type} missing constructor accepting inbound");
+                    }
+                    return type;
+                }).OfType<Type>();
+
+            // Enable workflow task tracing if needed
+            workflowTaskTracingEnabled = !options.DisableWorkflowTaskTracing && options.Workflows.Count > 0;
```
bergundy, line 78: Hmm.. might be cleaner to do this in the workflow worker instead of the the top level one.

src/Temporalio/Worker/TemporalWorkerOptions.cs
```diff
@@ -102,6 +107,70 @@ public TemporalWorkerOptions(string taskQueue)
...
+        /// Gets or sets the maximum number of local activities that will ever be given to this
+        /// worker concurrently.
+        /// </summary>
+        public int MaxConcurrentLocalActivities { get; set; } = 100;
+
+        /// <summary>
+        /// Gets or sets a value indicating whether the worker will only handle workflows and local
+        /// activities.
+        /// </summary>
+        public bool DisableRemoteActivities { get; set; }
```
bergundy, line 132: nit: there's not really a concept of remote activities, we've used it before but I think it'd be a bit clearer to use NonLocal instead.

src/Temporalio/Worker/WorkflowWorker.cs
cretz: @Sushisource - do you think that's a better approach than what I have today where I do evict but still kinda occupy a slot locally? I'm thinking so because I bet that task I'm holding has a reference to my task scheduler which means the whole workflow is probably still in memory.

src/Temporalio/Worker/WorkflowInstance.cs
cretz: I think we decided this is just an accepted Temporal feature in all SDKs. Not even a warn-based one.

src/Temporalio/Worker/WorkflowInstance.cs
cretz: Opened https://github.com/temporalio/sdk-core/issues/497

src/Temporalio/Worker/WorkflowInstance.cs
cretz: Correct for this PR, but the next one will have things on outbound interceptor (e.g. starting a timer) that is too far separated from this to have a completion passed through.

src/Temporalio/Worker/WorkflowCodecHelper.cs
```diff
@@ -0,0 +1,279 @@
...
+                case WorkflowActivationCompletion.StatusOneofCase.Failed:
+                    if (comp.Failed.Failure_ != null)
+                    {
+                        await codec.EncodeFailureAsync(comp.Failed.Failure_).ConfigureAwait(false);
+                    }
+                    break;
+                case WorkflowActivationCompletion.StatusOneofCase.Successful:
+                    foreach (var cmd in comp.Successful.Commands)
+                    {
+                        await EncodeAsync(codec, cmd).ConfigureAwait(false);
```
bergundy, line 39: Would be nice to parallelize this

src/Temporalio/Worker/WorkflowCodecHelper.cs
```diff
@@ -0,0 +1,279 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Decode the activation.
+        /// </summary>
+        /// <param name="codec">Codec to use.</param>
+        /// <param name="act">Activation to decode.</param>
+        /// <returns>Task for completion.</returns>
+        internal static async Task DecodeAsync(IPayloadCodec codec, WorkflowActivation act)
```
bergundy, line 51: Can we have some test that verifies every payload goes through these methods?
In TS we can catch this at compile time but the type system here obviously isn't complex enough for that.
I mostly worry about forgetting to handle new jobs and commands as we add those.

src/Temporalio/Worker/WorkflowCodecHelper.cs
```diff
@@ -0,0 +1,279 @@
...
+        {
+            foreach (var job in act.Jobs)
+            {
+                switch (job.VariantCase)
```
bergundy, line 55: What about headers? those should be decoded / encoded here.
Mostly for skipping the deadline detector in workflow.

src/Temporalio/Worker/WorkflowInstance.cs
```diff
@@ -0,0 +1,503 @@
...
+
+        /// <inheritdoc/>
+        public WorkflowActivationCompletion Activate(WorkflowActivation act)
+        {
+            using (Logger.BeginScope(Info.LoggerScope))
+            {
+                completion = new() { RunId = act.RunId, Successful = new() };
+                currentActivationException = null;
+                // We need to sort jobs
+                var jobs = act.Jobs.OrderBy(job =>
```
bergundy, line 195: Core should just do this for us, no need to repeat this in every lang.

src/Temporalio/Worker/WorkflowInstance.cs
```diff
@@ -0,0 +1,503 @@
...
+                        (e as ActivityFailureException)?.InnerException is CancelledFailureException ||
+                        (e as ChildWorkflowFailureException)?.InnerException is CancelledFailureException))
+                {
+                    // If cancel was ever requested and this is a cancellation or an activity/child
+                    // cancellation, we add a cancel command. Technically this means that a
+                    // swallowed cancel followed by, say, an activity cancel later on will show the
+                    // workflow as cancelled. But this is a Temporal limitation in that cancellation
+                    // is a state not an event.
+                    Logger.LogDebug(e, "Workflow raised cancel with run ID {RunID}", Info.RunID);
+                    AddCommand(new() { CancelWorkflowExecution = new() });
```
bergundy, line 400: This condition should be tested (when implemented)

src/Temporalio/Worker/WorkflowWorker.cs
Sushisource: I think that'd be swell - but you do still want to fail the activation so that the user ends up seeing WFT failure "because deadlocked" in their history. As long as that still happens the approach you describe makes sense to me.

src/Temporalio/Worker/WorkflowInstanceDetails.cs
```diff
@@ -0,0 +1,53 @@
...
+    /// <param name="DisableTaskTracing">Whether task tracing is disabled.</param>
+    /// <remarks>
+    /// This is built to be easily serializable in case we do want a sandbox one day.
+    /// </remarks>
+    internal record WorkflowInstanceDetails(
+        string Namespace,
+        string TaskQueue,
+        WorkflowAttribute.Definition Definition,
+        WorkflowActivation InitialActivation,
+        StartWorkflow Start,
```
bergundy, line 29: You don't need to save these, better to free up memory used by the payloads.

src/Temporalio/Worker/WorkflowWorker.cs
bergundy: > I assume shutdown will not send a poll shutdown error ending polling until all outstanding remove-from-cache's are handled?

That's the desired behavior but I'm not convinced that's what currently happens. We'll need to address that core side.

src/Temporalio/Worker/WorkflowWorker.cs
```diff
@@ -0,0 +1,254 @@
...
+                if (codec != null)
+                {
+                    await WorkflowCodecHelper.DecodeAsync(codec, act).ConfigureAwait(false);
+                }
+
+                // Log proto at trace level
+                logger.LogTrace("Received workflow activation: {Activation}", act);
+
+                // We only have to run if there are any non-remove jobs
+                removeJob = act.Jobs.Select(job => job.RemoveFromCache).FirstOrDefault(job => job != null);
```
bergundy, line 122: @Sushisource can / does core ensure that remove from cache would always be last?

bergundy: When would lang get remove from cache + other jobs?

src/Temporalio/Worker/WorkflowWorker.cs
```diff
@@ -0,0 +1,254 @@
...
+            if (codec != null)
+            {
+                try
+                {
+                    await WorkflowCodecHelper.EncodeAsync(codec, comp).ConfigureAwait(false);
+                }
+                catch (Exception e)
+                {
+                    logger.LogError(e, "Failed encoding completion on workflow with run ID {RunId}", act.RunId);
+                    comp.Failed = new() { Failure_ = new() { Message = $"Failed encoding completion: {e}" } };
```
bergundy, line 194: Hmm... looks like you might have both Failure and Successful set in the completion here.

src/Temporalio/Worker/WorkflowWorker.cs
```diff
@@ -0,0 +1,254 @@
...
+            // Send completion
+            logger.LogTrace("Sending workflow completion: {Completion}", comp);
+            try
+            {
+                await Worker.BridgeWorker.CompleteWorkflowActivationAsync(comp).ConfigureAwait(false);
+            }
+            catch (Exception e)
+            {
+                logger.LogError(e, "Failed completing activation on  {RunId}", act.RunId);
+                comp.Failed = new() { Failure_ = new() { Message = $"Failed encoding completion: {e}" } };
```
bergundy, line 217: This is redundant / inaccurate

src/Temporalio/Worker/TemporalWorker.cs
cretz: The workflow worker is not disposable and I want this registered/unregistered on construct/dispose, not start/shutdown. No particular reason, just feels better to do it on these boundaries.

src/Temporalio/Worker/TemporalWorkerOptions.cs
cretz: Hrmm, this is what it's called in Python. I assumed "remote" means "non-local". You're saying this should be called `DisableNonLocalActivities`? Maybe I should change to Go and Java wording of `LocalActivityWorkerOnly`?

src/Temporalio/Worker/WorkflowCodecHelper.cs
cretz: I am not convinced there is a ton of value in parallelizing. In fact, I think there's more value in batching. Shall we change our codecs to submit these things in batches? Or do we accept code complication is not enough benefit to contort the code to try to do multiple of these things at once?

src/Temporalio/Worker/WorkflowCodecHelper.cs
cretz: Yes. I was thinking about writing a custom codec test that recursively, reflectively iterates over every payload in the nested proto hierarchy (excluding search attributes), and creates a new object all the way from root (i.e. clones) and sets the payload and runs root through this and confirms the payload was converted. This should catch new payloads. We could also write a code generator like the Go proxy does, but I think this is fine for now.

src/Temporalio/Worker/WorkflowInstance.cs
cretz: Feel free to create issues in core when you come across these things. I won't let it slow down my progress.

src/Temporalio/Worker/WorkflowCodecHelper.cs
cretz: Hrmm, I think I missed headers in Python (and the aforementioned reflection test would catch these for .NET). Will fix here, but what should I do about Python? Leave headers unencoded for compat reasons?

src/Temporalio/Worker/WorkflowInstance.cs
cretz: Which "condition"? General workflow cancellation in its multiple bubbled forms? Yes, that will be part of tests.

src/Temporalio/Worker/WorkflowInstanceDetails.cs
cretz: True. I will keep them here but can avoid setting the whole details object as a field on the instance

src/Temporalio/Worker/WorkflowWorker.cs
cretz: Protobuf `oneof`s properly handle mutual exclusivity internally

src/Temporalio/Worker/WorkflowWorker.cs
cretz: Yes, will remove, bad copy/pasta

cretz:
@bergundy and @Sushisource - I plan on making required changes in next PR, see any issues with that? I can either merge this when that PR comes about or I can close this and have that PR be all this new stuff + a bunch of new things? Preference?

src/Temporalio/Worker/TemporalWorkerOptions.cs
bergundy: I think `DisableNonLocalActivities` or better yet `EnableNonLocalActivities` is the clearest.
`LocalActivityWorkerOnly` isn't technically correct since the worker runs workflows (which is why Spencer renamed it in Core).

src/Temporalio/Worker/WorkflowCodecHelper.cs
bergundy: Value is shorter task processing times and reduced chance of workflow task timeouts especially for remote codecs.
Not blocking the PR, we don't parallelize all invocations in other SDKs.

bergundy: Just thought it would be nice to consider since this is a fresh implementation.

src/Temporalio/Worker/WorkflowCodecHelper.cs
bergundy: Agree but could use an issue.

src/Temporalio/Worker/WorkflowCodecHelper.cs
bergundy: Headers should be encoded, they may contain sensitive information.

src/Temporalio/Worker/WorkflowInstance.cs
bergundy: 👍

bergundy (approved):
Any unaddressed comments can be addressed in a follow up PR.
Approving to maintain the momentum.

src/Temporalio/Worker/WorkflowWorker.cs
Sushisource: It's always last in the job list, but you can get other jobs if the WFT is about to be replied to and the cache is at the limit. So we finish the task, reply, then evict to make room for someone else

src/Temporalio/Worker/TemporalWorkerOptions.cs
cretz: We would prefer false to be the default IMO (like Go, proto, unset .NET boolean primitive, etc). So, what should we name it in the interest of clarity and consistency and default-false semantics? `DisableNonLocalActivities` is an inconsistent, ugly double negative.

In general, can you provide an adjective for "non-local" that Temporal can use company wide? Why can "remote" not be it since it's the opposite of local?

> LocalActivityWorkerOnly isn't technically correct since the worker runs workflows

It's only incorrect depending on how you read it. You read "local-activity worker only" and I read it as "local activity-worker only" completely unrelated to workflow workers. In fact, most of our worker options apply to either activity workers or workflow workers, not both. I mean I guess I can rename `GracefulShutdownTimeout` to `GracefulShutdownActivityWorkerTimeout` and `MaxHeartbeatThrottleInterval` to `MaxNonLocalActivityHeartbeatThrottleInterval` and such if we're concerned.

src/Temporalio/Worker/WorkflowCodecHelper.cs
cretz: You would be much shorter processing times for remote codecs if you batched. Should we batch instead?

src/Temporalio/Worker/WorkflowCodecHelper.cs
cretz: I have a lot of tests and things to verify that I don't have issues for

src/Temporalio/Worker/WorkflowCodecHelper.cs
cretz: Agreed on what they should be, there was never any argument there. I am asking about what to do about Python knowing it already exists unencoded today. There are compatibility concerns. I will fix in .NET.

src/Temporalio/Worker/WorkflowWorker.cs
cretz: I don't need to make that assumption I don't think unless a microbenchmark proves otherwise.

src/Temporalio/Worker/TemporalWorkerOptions.cs
bergundy: If you prefer false for default I would avoid the double negative.

I think I called non-local, remote activities a couple of years ago in the TS SDK but I haven't seen that concept anywhere in our docs.

`LocalActivityWorkerOnly` works for me or `DisableActivityPolling` (even though the latter doesn't account for eager activities), @Sushisource had issues with the local only wording.

cretz: :+1: We use the word "remote" in Rust and Python too. If @Sushisource doesn't have any objection, I'll switch to `LocalActivityWorkerOnly` even though a tad confusing because it at least matches Go and Java.

src/Temporalio/Worker/WorkflowWorker.cs
bergundy: Agree it doesn't matter for perf reasons, this was mostly for my understanding of the guarantees.

src/Temporalio/Worker/WorkflowCodecHelper.cs
bergundy: I meant for Python

src/Temporalio/Worker/WorkflowInstance.cs
bergundy: Done

src/Temporalio/Worker/WorkflowCodecHelper.cs
bergundy: As long as you're tracking this...

src/Temporalio/Worker/WorkflowCodecHelper.cs
bergundy: I agree that batching is better approach than parallelizing for both remote and non remote codecs.

We should just make sure that a single batch correlates to a single "context" (workflow / activity) to simplify the user API for when they need their codecs to be context aware.

src/Temporalio/Worker/WorkflowCodecHelper.cs
cretz: > Headers should be encoded, they may contain sensitive information.
> I meant for Python

We know that headers should be encoded. The question is what to do for Python when users are already using unencoded headers today. The question is about compatibility not about whether they should be encoded. I can take this out of this PR and we can discuss what we should do in Python since the question is not coming across here.

src/Temporalio/Worker/WorkflowCodecHelper.cs
cretz: For performance reasons, should we update SDKs to support batching? Should we update .NET to support batching? Or should we accept that performance benefit is negligible compared to separate requests.

bergundy: I don't know how negligible it is, I think there are definitely cases where it would matter.
In any case I don't think we want to take this on for all SDKs right now.
In the meantime, batching could easily be built on top of concurrency user side by waiting a tick before submitting any accumulated requests (similarly to how this is done in [dataloader](https://www.npmjs.com/package/dataloader)) .

I wouldn't necessarily change the API and wouldn't block this PR, I would just take this into consideration.
In TypeScript we try to run codecs concurrently as much as possible (but I admit that there's room for improvement).

src/Temporalio/Worker/WorkflowInstance.cs
cretz: Fixed in next PR

src/Temporalio/Worker/WorkflowInstance.cs
cretz: Fixed in next PR

src/Temporalio/Worker/WorkflowWorker.cs
cretz: Fixed in next PR to not complete the remove-job task until the hung workflow task is complete

src/Temporalio/Worker/WorkflowWorker.cs
cretz: Fixed in next PR

src/Temporalio/Worker/WorkflowWorker.cs
cretz: Fixed in next PR

src/Temporalio/Worker/TemporalWorkerOptions.cs
cretz: Fixed in next PR

src/Temporalio/Worker/WorkflowCodecHelper.cs
cretz: While attempting to parallelize this, two other reasons came up to potentially not do it: 1) this will aggregate exceptions, so you could throw 50 exceptions from here for 50 commands, and 2) This can explode remote/IO call count filling up the thread pool. Technically they can prevent this by debouncing, but it is not as clear.

If you have a codec that does CPU bound AES let's say, and I call this concurrently for all 50 commands, .NET will try to use as much of the thread pool as it has available. Scatter-gather makes less sense for codecs for potentially IO/CPU-bound tasks in multithreaded languages.

As the implementation sits right now, we're not gonna use more than a single thread for a single workflow task, which is nice and predictable.

src/Temporalio/Worker/WorkflowCodecHelper.cs
cretz: Fixed in next PR

src/Temporalio/Worker/WorkflowInstanceDetails.cs
cretz: Fixed in next PR

src/Temporalio/Worker/WorkflowWorker.cs
cretz: Fixed in next PR

Issue #27, opened 2023-02-27
[Feature Request] Support dynamic workflows, activities, signals, and queries


cretz:
### Describe the solution you'd like

Need to support all of these things dynamically as part of general Temporal feature https://github.com/temporalio/features/issues/201.

PR #34, opened 2023-03-09
Search attributes, memos, and other updates

cretz:
## What was changed

* Changed search attributes to a more type safe form conforming to https://github.com/temporalio/proposals/blob/master/sdk-typed-search-attributes.md
* Changed memos to a more type safe form
* Added `Workflow.TypedSearchAttributes` and `Workflow.UpsertTypedSearchAttributes`
* Added `Workflow.Memo` and `Workflow.UpsertMemo`
* Added `IEncodedRawValue` and `IRawValue` for converter + payload tuple needs. `IRawValue` will also be used a lot for #27
* Added to support to check for non-workflow-scheduled tasks while workflow scheduler is current
* Minor PR fixes from #31

NOTE: PR may continue to fail Windows CI while https://github.com/dotnet/format/issues/1800 remains a bug.


src/Temporalio/Client/WorkflowStartOptions.cs
```diff
@@ -69,14 +69,14 @@ public WorkflowStartOptions(string id, string taskQueue)
...
-        /// Gets or sets the memo for the workflow.
+        /// Gets or sets the memo for the workflow. Values for the memo cannot be null.
         /// </summary>
         public IReadOnlyCollection<KeyValuePair<string, object>>? Memo { get; set; }
 
         /// <summary>
         /// Gets or sets the search attributes for the workflow.
         /// </summary>
-        public IReadOnlyCollection<KeyValuePair<string, object>>? SearchAttributes { get; set; }
+        public SearchAttributeCollection? TypedSearchAttributes { get; set; }
```
bergundy, line 79: Do you want to call this `TypedSearchAttributes` for uniformity across SDKs? Since we never had untyped ones I can see an argument for just calling this `SearchAttributes`.

src/Temporalio/Converters/EncodedRawValue.cs
```diff
@@ -0,0 +1,20 @@
...
+using Temporalio.Api.Common.V1;
+
+namespace Temporalio.Converters
+{
+    /// <summary>
+    /// Implementation of a <see cref="IEncodedRawValue" />.
+    /// </summary>
+    /// <param name="DataConverter">Data converter to perform conversions.</param>
+    /// <param name="Payload">Raw payload.</param>
+    public record EncodedRawValue(DataConverter DataConverter, Payload Payload) : IEncodedRawValue
```
bergundy, line 12: Is this a .NET thing that everything needs both an interface and a class? Is there other reasoning behind this?

src/Temporalio/Client/WorkflowStartOptions.cs
cretz: Exactly. I accept that this could be better in .NET but is best to be named the same in every language.

src/Temporalio/Converters/EncodedRawValue.cs
cretz: I need an implementation of the interface somewhere, otherwise how do I use it? I do think there is technically value in making this implementation public because you can technically cast and get the payload and data converter back out. But technically I could make this `internal`. But yeah, I have to have an implementation because this is what I set as the `Memo` dictionary keys for workflow executions coming back from client calls.

README.md
```diff
@@ -356,6 +356,7 @@ TODO(cretz): Document and try to prevent:
   * .NET does not allow us to intercept the timers. It forces them on a platform-specific `TimerQueue` which is
     non-deterministic.
   * We will try to prevent this at runtime with TPL event source capturing and at compile time with analyzers.
+* Users creating dataflow constructs without providing the current scheduler
```
bergundy, line 359:
```suggestion
* Users creating dataflow constructs without providing the current scheduler.
```

bergundy: For consistency with the content above.

tests/Temporalio.Tests/WorkflowEnvironmentTestBase.cs
```diff
@@ -20,4 +42,33 @@ protected WorkflowEnvironmentTestBase(ITestOutputHelper output, WorkflowEnvironm
...
+
+    protected async Task EnsureSearchAttributesPresentAsync()
+    {
+        // Only add search attributes if not present
+        var resp = await Client.Connection.WorkflowService.GetSearchAttributesAsync(new());
+        if (resp.Keys.ContainsKey(AttrBool.Name))
+        {
+            return;
+        }
+        await Client.Connection.OperatorService.AddSearchAttributesAsync(new()
```
bergundy, line 54: Obviously not for this PR but we could have a high level API for this too now that we have typed attributes.

src/Temporalio/SearchAttributeCollection.cs
```diff
@@ -0,0 +1,363 @@
...
+        /// <param name="searchAttributes">Proto search attributes.</param>
+        /// <returns>Search attribute collection.</returns>
+        public static SearchAttributeCollection FromProto(SearchAttributes searchAttributes)
+        {
+            // We intentionally don't shortcut an empty set of search attributes as Empty here
+            // because _technically_ the internal value may be mutated
+            static object JsonElementToObject(JsonElement elem, IndexedValueType valueType) => elem.ValueKind switch
+            {
+                JsonValueKind.Array =>
+                    elem.EnumerateArray().Select(j => JsonElementToObject(j, valueType)).ToList(),
```
bergundy, line 71: If we don't flatten non-keyword lists here, clients from other SDKs which set attribute values as lists won't be accessible AFAICT.

IIRC, we said that single value lists are acceptable.

bergundy (approved):
LGTM

src/Temporalio/Converters/EncodedRawValue.cs
bergundy: Just wondering why you need both instead of just the implementation.

cretz: There was a request to hide payloads from users like https://pkg.go.dev/go.temporal.io/sdk/converter#EncodedValue does. If we're not gonna do that, why not just use payloads for these things and make the converter available for use (an option brought up on internal Slack, but this was preferred)?

README.md
cretz: This is a TODO comment you're scrutinizing. It will be removed in favor of full docs once impl is done

tests/Temporalio.Tests/WorkflowEnvironmentTestBase.cs
cretz: That can be another proposal or approved change. Registering search attributes is not part of this effort by intention here or other SDKs when I implement there.

src/Temporalio/SearchAttributeCollection.cs
cretz: They are acceptable but wrong, so they have to be accessed via the untyped form. Typed search attributes are supported when created via typed search attributes. They are meant to be strict in their typed form.

PR #39, opened 2023-03-23
Workflow support for activities, children, and continue as new

cretz:
## What was changed

* Activities and local activities
* Child workflows (with the exception of sending signals to them atm)
* Continue as new support w/ type safety
* Doc fixes to include defaults
* Code fixes to use expression bodied members where possible
* Other minor changes:
  * `Workflow.NewGuid()`
  * workflow-not-registered tests
  * Renamed `Temporalio.Client.WorkflowStartOptions` to `Temporalio.Client.WorkflowOptions`

Please do a full review regardless of CI state (known Windows bug in dotnet format that MS is fixing)

### DIFF
--- a/src/Temporalio/Converters/BinaryPlainConverter.cs
+++ b/src/Temporalio/Converters/BinaryPlainConverter.cs
@@ -18,8 +18,7 @@ namespace Temporalio.Converters
         /// <inheritdoc />
         public bool TryToPayload(object? value, out Payload? payload)
         {
-            var bytes = value as byte[];
-            if (bytes == null)
+            if (value is not byte[] bytes)
             {
                 payload = null;
                 return false;
--- a/src/Temporalio/Worker/Interceptors/CreateContinueAsNewExceptionInput.cs
+++ b/src/Temporalio/Worker/Interceptors/CreateContinueAsNewExceptionInput.cs
@@ -1,0 +1,19 @@
+using System.Collections.Generic;
+using Temporalio.Api.Common.V1;
+using Temporalio.Workflows;
+
+namespace Temporalio.Worker.Interceptors
+{
+    /// <summary>
+    /// Input for <see cref="WorkflowOutboundInterceptor.CreateContinueAsNewException" />.
+    /// </summary>
+    /// <param name="Workflow">Workflow to continue.</param>
+    /// <param name="Args">Arguments for the workflow.</param>
+    /// <param name="Options">Options passed in to continue as new.</param>
+    /// <param name="Headers">Headers to include.</param>
+    public record CreateContinueAsNewExceptionInput(
+        string Workflow,
+        IReadOnlyCollection<object?> Args,
+        ContinueAsNewOptions? Options,
+        IDictionary<string, Payload>? Headers);
+}
--- a/src/Temporalio/Worker/Interceptors/ScheduleActivityInput.cs
+++ b/src/Temporalio/Worker/Interceptors/ScheduleActivityInput.cs
@@ -1,0 +1,19 @@
+using System.Collections.Generic;
+using Temporalio.Api.Common.V1;
+using Temporalio.Workflows;
+
+namespace Temporalio.Worker.Interceptors
+{
+    /// <summary>
+    /// Input for <see cref="WorkflowOutboundInterceptor.ScheduleActivityAsync" />.
+    /// </summary>
+    /// <param name="Activity">Activity type name.</param>
+    /// <param name="Args">Activity args.</param>
+    /// <param name="Options">Activity options.</param>
+    /// <param name="Headers">Headers.</param>
+    public record ScheduleActivityInput(
+        string Activity,
+        IReadOnlyCollection<object?> Args,
+        ActivityOptions Options,
+        IDictionary<string, Payload>? Headers);
+}
--- a/src/Temporalio/Worker/WorkflowInstance.cs
+++ b/src/Temporalio/Worker/WorkflowInstance.cs
@@ -10,6 +10,8 @@ using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Logging;
 using Temporalio.Api.Common.V1;
+using Temporalio.Bridge.Api.ActivityResult;
+using Temporalio.Bridge.Api.ChildWorkflow;
 using Temporalio.Bridge.Api.WorkflowActivation;
 using Temporalio.Bridge.Api.WorkflowCommands;
 using Temporalio.Bridge.Api.WorkflowCompletion;
@@ -40,6 +42,9 @@ namespace Temporalio.Worker
         private readonly LinkedList<Task> scheduledTasks = new();
         private readonly Dictionary<Task, LinkedListNode<Task>> scheduledTaskNodes = new();
         private readonly Dictionary<uint, TaskCompletionSource<object?>> timersPending = new();
+        private readonly Dictionary<uint, TaskCompletionSource<ActivityResolution>> activitiesPending = new();
+        private readonly Dictionary<uint, TaskCompletionSource<ResolveChildWorkflowExecutionStart>> childWorkflowsPendingStart = new();
+        private readonly Dictionary<uint, TaskCompletionSource<ChildWorkflowResult>> childWorkflowsPendingCompletion = new();
         private readonly Dictionary<string, List<SignalWorkflow>> bufferedSignals = new();
         private readonly CancellationTokenSource cancellationTokenSource = new();
         private readonly LinkedList<Tuple<Func<bool>, TaskCompletionSource<object?>>> conditions = new();
@@ -49,6 +54,8 @@ namespace Temporalio.Worker
         private object? instance;
         private Exception? currentActivationException;
         private uint timerCounter;
+        private uint activityCounter;
+        private uint childWorkflowCounter;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="WorkflowInstance"/> class.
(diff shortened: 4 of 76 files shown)


src/Temporalio/Converters/BinaryPlainConverter.cs
```diff
@@ -18,8 +18,7 @@ public class BinaryPlainConverter : IEncodingConverter
         /// <inheritdoc />
         public bool TryToPayload(object? value, out Payload? payload)
         {
-            var bytes = value as byte[];
-            if (bytes == null)
+            if (value is not byte[] bytes)
```
bergundy, line 21: I'm starting to like this language more and more as I read your PRs.

src/Temporalio/Worker/Interceptors/CreateContinueAsNewExceptionInput.cs
```diff
@@ -0,0 +1,19 @@
...
+{
+    /// <summary>
+    /// Input for <see cref="WorkflowOutboundInterceptor.CreateContinueAsNewException" />.
+    /// </summary>
+    /// <param name="Workflow">Workflow to continue.</param>
+    /// <param name="Args">Arguments for the workflow.</param>
+    /// <param name="Options">Options passed in to continue as new.</param>
+    /// <param name="Headers">Headers to include.</param>
+    public record CreateContinueAsNewExceptionInput(
+        string Workflow,
```
bergundy, line 15: Could potentially be useful to pass a reference to the workflow class if you have it so the interceptor can extract the argument types.
Same goes for the activity input.

src/Temporalio/Worker/Interceptors/CreateContinueAsNewExceptionInput.cs
```diff
@@ -0,0 +1,19 @@
+using System.Collections.Generic;
+using Temporalio.Api.Common.V1;
+using Temporalio.Workflows;
+
+namespace Temporalio.Worker.Interceptors
+{
+    /// <summary>
+    /// Input for <see cref="WorkflowOutboundInterceptor.CreateContinueAsNewException" />.
+    /// </summary>
+    /// <param name="Workflow">Workflow to continue.</param>
```
bergundy, line 10: Should we name this `WorkflowType` to disambiguate with e.g. workflow ID?
I'd at least mention in the docstring.

bergundy: I see below that you describe `Activity` as Activity type name.

src/Temporalio/Worker/Interceptors/ScheduleActivityInput.cs
```diff
@@ -0,0 +1,19 @@
...
+namespace Temporalio.Worker.Interceptors
+{
+    /// <summary>
+    /// Input for <see cref="WorkflowOutboundInterceptor.ScheduleActivityAsync" />.
+    /// </summary>
+    /// <param name="Activity">Activity type name.</param>
+    /// <param name="Args">Activity args.</param>
+    /// <param name="Options">Activity options.</param>
+    /// <param name="Headers">Headers.</param>
+    public record ScheduleActivityInput(
```
bergundy, line 14: I thought you were going with ExecuteActivity because the interceptor wraps the entire execution.

src/Temporalio/Worker/WorkflowInstance.cs
```diff
@@ -534,17 +568,41 @@ private async Task RunTopLevelAsync(Func<Task> func)
...
-                catch (ContinueAsNewException)
+                catch (ContinueAsNewException e)
                 {
                     Logger.LogDebug("Workflow requested continue as new with run ID {RunID}", Info.RunID);
-                    // TODO(cretz): This
-                    throw new NotImplementedException();
+                    var cmd = new ContinueAsNewWorkflowExecution()
+                    {
+                        WorkflowType = e.Input.Workflow,
+                        TaskQueue = e.Input.Options?.TaskQueue ?? string.Empty,
```
bergundy, line 577: I wonder if task queue is optional here, we use the one from workflow info in TypeScript as default.

src/Temporalio/Worker/WorkflowInstance.cs
```diff
@@ -935,6 +1034,403 @@ public override Task DelayAsync(DelayAsyncInput input)
...
+                // We do not even want to schedule if the cancellation token is already cancelled.
+                // We choose to use cancelled failure instead of wrapping in child failure which is
+                // similar to what Java and TypeScript do, with the accepted tradeoff that it makes
+                // catch clauses more difficult (hence the presence of
+                // TemporalException.IsCancelledException helper).
+                if (token.IsCancellationRequested)
+                {
+                    return Task.FromException<ChildWorkflowHandle<TResult>>(
+                        new CancelledFailureException("Child cancelled before scheduled"));
+                }
```
bergundy, line 1158: Curious why you're not doing this eager cancellation check for activities and local activities. You should AFAICT.

bergundy (approved):
I mostly skimmed the doc strings and skipped the tests.
The rest LGTM.

src/Temporalio/Worker/Interceptors/CreateContinueAsNewExceptionInput.cs
cretz: I may wait until that day occurs. Landscape currently:

* Go - Does get `interface{}` so it _could_ be a function
* Java - Only supports string like we do here
* Python - Only supports string like we do here
* TypeScript - I _think_ it only supports string like we do here (from ContinueAsNewOptions), it's hard to follow

But this does remind me I need to put a doc warning on these records that the constructor is not backwards compatible and should never be called by users (I can move to an internal constructor and explicit properties, it's just kinda a pain).

src/Temporalio/Worker/Interceptors/CreateContinueAsNewExceptionInput.cs
cretz: I call this `Workflow` and `Activity` just about everywhere except on `WorkflowInfo` and `ActivityInfo` where the differentiation is needed so I use `WorkflowType` and `ActivityType`. The word "type" confuses users I've found. But I can update the docs to anything of course.

src/Temporalio/Worker/Interceptors/ScheduleActivityInput.cs
cretz: I am using `ExecuteActivity` from a caller POV (see the `Workflows` class). This is just the interceptor, and I have to disambiguate because `ExecuteActivityInput` is already used for activity interceptors (and I didn't want to break this namespace up). This was the subject of the internal Slack thread we had the other day (can link if needed).

src/Temporalio/Worker/WorkflowInstance.cs
cretz: Hrmm, this is working today so apparently core is defaulting this. However core does not document in their protos that they default this so I will do so explicitly here (like I do for activities and child workflows already).

src/Temporalio/Worker/WorkflowInstance.cs
cretz: I am. To share code, `ScheduleActivityAsync` and `ScheduleLocalActivityAsync` delegate to `ExecuteActivityInternalAsync` which performs this check at the top. I even have a test to ensure this behavior.

src/Temporalio/Worker/WorkflowInstance.cs
```diff
@@ -935,6 +1034,403 @@ public override Task DelayAsync(DelayAsyncInput input)
...
+                            TaskQueue = input.Options.TaskQueue ?? instance.Info.TaskQueue,
+                            Arguments = { instance.payloadConverter.ToPayloads(input.Args) },
+                            RetryPolicy = input.Options.RetryPolicy?.ToProto(),
+                            CancellationType = (Bridge.Api.WorkflowCommands.ActivityCancellationType)input.Options.CancellationType,
+                        };
+                        if (input.Headers is IDictionary<string, Payload> headers)
+                        {
+                            cmd.Headers.Add(headers);
+                        }
+                        if (input.Options.ScheduleToCloseTimeout is TimeSpan schedToClose)
```
Sushisource, line 1067: Why do these (and other places) have these `is` checks? Can't they only be of that type? Some particular reason to prefer this over a "not null" kinda check?

src/Temporalio/Worker/WorkflowInstance.cs
```diff
@@ -935,6 +1034,403 @@ public override Task DelayAsync(DelayAsyncInput input)
...
+            {
+                if (input.Options.StartToCloseTimeout == null &&
+                    input.Options.ScheduleToCloseTimeout == null)
+                {
+                    throw new ArgumentException("Activity options must have StartToCloseTimeout or ScheduleToCloseTimeout");
+                }
+
+                return ExecuteActivityInternalAsync<TResult>(
+                    local: false,
+                    doBackoff =>
```
Sushisource, line 1050: Backoffs cannot apply to non-local activities. You should just error here or log an error or whatever instead of having all this code.

I was tempted to create two kinds of resolution proto messages to make this unrepresentable, but that was a bit overkill.

Sushisource (approved):
Only one thing that matters around not handling backoff for nonlocal activities

src/Temporalio/Worker/WorkflowInstance.cs
cretz: > Some particular reason to prefer this over a "not null" kinda check?

Yes, it's only for nullness check. With this, `schedToClose` is a known non-null variable whereas `input.Options.ScheduleToCloseTimeout != null` does not provide that guarantee at compile time making things a bit ugly. But either works. This is the .NET equiv of `if let Some(schedToClose) = ...`.

src/Temporalio/Worker/WorkflowInstance.cs
cretz: Yeah, I can error if I ever get called with a non-null backoff. Here I was just ignoring it knowing that core will never call this a second time so the enforcement didn't seem necessary, but i can add it. But I still need a "create command" code to have the shared code be cleaner.

src/Temporalio/Converters/BinaryPlainConverter.cs
cretz: Wait until you see F#

cretz:
Merging (and changes will be in next PR)

PR #40, opened 2023-03-30
Misc workflow features

cretz:
## What was changed

* Added replay safe logger
* Opt-in stack trace query support (cannot be done by default due to perf concerns)
* Patch support
* Record/struct workflow support
* Catch/fail on more invalid workflow calls
* External workflow signal/cancel support
* Child workflow signal support

This basically rounds out all modern workflow features. I just need docs, maybe a few more tests, and maybe some refactoring before next alpha which will be a full-featured SDK.

### DIFF
--- a/src/Temporalio/Activities/ActivityRefs.cs
+++ b/src/Temporalio/Activities/ActivityRefs.cs
@@ -1,0 +1,16 @@
+namespace Temporalio.Activities
+{
+    /// <summary>
+    /// Static helper for creating an instance of a type to reference its methods.
+    /// </summary>
+    public static class ActivityRefs
+    {
+        /// <summary>
+        /// Create an instance of the given type. This should only be used to reference methods on,
+        /// not to make any calls on. Instances returned may not be instantiated.
+        /// </summary>
+        /// <typeparam name="T">Type to create an instance of.</typeparam>
+        /// <returns>Instance of this type.</returns>
+        public static T Create<T>() => Refs.Create<T>();
+    }
+}
--- a/src/Temporalio/Worker/ReplaySafeLogger.cs
+++ b/src/Temporalio/Worker/ReplaySafeLogger.cs
@@ -1,0 +1,39 @@
+using System;
+using Microsoft.Extensions.Logging;
+
+namespace Temporalio.Worker
+{
+    /// <summary>
+    /// Logger that will not log during replay.
+    /// </summary>
+    internal class ReplaySafeLogger : ILogger
+    {
+        private readonly ILogger underlying;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ReplaySafeLogger"/> class.
+        /// </summary>
+        /// <param name="underlying">Logger to wrap.</param>
+        public ReplaySafeLogger(ILogger underlying) => this.underlying = underlying;
+
+        /// <inheritdoc />
+        public IDisposable BeginScope<TState>(TState state) => underlying.BeginScope<TState>(state);
+
+        /// <inheritdoc />
+        public bool IsEnabled(LogLevel logLevel) => underlying.IsEnabled(logLevel);
+
+        /// <inheritdoc />
+        public void Log<TState>(
+            LogLevel logLevel,
+            EventId eventId,
+            TState state,
+            Exception exception,
+            Func<TState, Exception, string> formatter)
+        {
+            if (!Workflows.Workflow.Unsafe.IsReplaying)
+            {
+                underlying.Log<TState>(logLevel, eventId, state, exception, formatter);
+            }
+        }
+    }
+}
--- a/src/Temporalio/Worker/WorkflowInstance.cs
+++ b/src/Temporalio/Worker/WorkflowInstance.cs
@@ -48,6 +48,8 @@ namespace Temporalio.Worker
         private readonly Dictionary<string, List<SignalWorkflow>> bufferedSignals = new();
         private readonly CancellationTokenSource cancellationTokenSource = new();
         private readonly LinkedList<Tuple<Func<bool>, TaskCompletionSource<object?>>> conditions = new();
+        private readonly HashSet<string> patchesNotified = new();
+        private readonly Dictionary<string, bool> patchesMemoized = new();
         private readonly WorkflowStackTrace workflowStackTrace;
         // Only non-null if stack trace is not None
         private readonly LinkedList<System.Diagnostics.StackTrace>? pendingTaskStackTraces;
@@ -236,6 +238,27 @@ namespace Temporalio.Worker
             string activity, IReadOnlyCollection<object?> args, LocalActivityOptions options) =>
             outbound.Value.ScheduleLocalActivityAsync<TResult>(
                 new(Activity: activity, Args: args, Options: options, Headers: null));
+
+        /// <inheritdoc />
+        public bool Patch(string patchID, bool deprecated)
+        {
+            // Use memoized result if present. If this is being deprecated, we can still use
+            // memoized result and skip the command.
+            if (patchesMemoized.TryGetValue(patchID, out var patched))
+            {
+                return patched;
+            }
+            patched = !IsReplaying || patchesNotified.Contains(patchID);
+            patchesMemoized[patchID] = patched;
+            if (patched)
+            {
+                AddCommand(new()
+                {
+                    SetPatchMarker = new() { PatchId = patchID, Deprecated = deprecated },
+                });
+            }
+            return patched;
+        }
 
         /// <inheritdoc/>
         public Task<ChildWorkflowHandle<TResult>> StartChildWorkflowAsync<TResult>(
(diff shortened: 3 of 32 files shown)


src/Temporalio/Workflows/ChildWorkflowSignalOptions.cs
```diff
@@ -0,0 +1,24 @@
...
+    /// Options for child workflow signalling.
+    /// </summary>
+    public class ChildWorkflowSignalOptions : ICloneable
+    {
+        /// <summary>
+        /// Gets or sets the cancellation token to cancel the signal request. If the signal is
+        /// already sent, this does nothing. If unset, this defaults to the workflow cancellation
+        /// token.
+        /// </summary>
+        public CancellationToken? CancellationToken { get; set; }
```
cretz, line 16: @bergundy and @Sushisource - I need help knowing whether this (and the one for external) has any user-facing value. I am trying to understand the value of `CancelSignalWorkflow` in core as it relates to users.

src/Temporalio/Worker/WorkflowTracingEventListener.cs
cretz, line 1: This file was renamed from `WorkflowTaskEventListener` and had enough added (namely another event source that captures thread transfer on timers) that Git sees it as a completely new impl.

src/Temporalio/Workflows/ChildWorkflowSignalOptions.cs
Sushisource: Has _any_? Sure, I suppose there's some in theory. In practice it seems crazy unlikely. I would say you can leave it out.

src/Temporalio/Worker/WorkflowInstance.cs
```diff
@@ -555,9 +592,57 @@ private void AddCommand(WorkflowCommand cmd)
...
+                    var task = origFunc();
+                    return task.ContinueWith(
+                        _ =>
+                        {
+                            stackTraces.Remove(node);
+                            return task;
+                        },
+                        this).Unwrap();
+                };
+            }
```
Sushisource, line 642: may as well extract this identical bit to a helper

src/Temporalio/Worker/WorkflowInstance.cs
```diff
@@ -690,25 +787,35 @@ private void ApplyQueryWorkflow(QueryWorkflow query)
                 var origCmdCount = completion?.Successful?.Commands?.Count;
                 try
                 {
-                    // Find definition or fail
-                    var queries = mutableQueries.IsValueCreated ? mutableQueries.Value : defn.Queries;
-                    if (!queries.TryGetValue(query.QueryType, out var queryDefn))
+                    WorkflowQueryDefinition? queryDefn;
+                    // If it's a stack trace query, create definition
+                    if (query.QueryType == "__stack_trace")
```
Sushisource, line 792: Lift magic string up into a const

src/Temporalio/Worker/WorkflowTracingEventListener.cs
```diff
@@ -0,0 +1,206 @@
...
+        /// Unregister this as no longer needed by a worker.
+        /// </summary>
+        public void Unregister()
+        {
+            lock (eventSourceLock)
+            {
+                eventSourceListenerCount--;
+                // Disable if we're the last and there is a source
+                // TODO(cretz): Any perf concern with thrashing enable/disable if they are
+                // starting/stopping workers frequently?
```
Sushisource, line 77: No lol

tests/Temporalio.Tests/Worker/WorkflowWorkerTests.cs
```diff
@@ -1881,10 +2114,329 @@ public async Task ExecuteWorkflowAsync_AlreadyStartedChild_FailsProperly()
...
+            {
+                Assert.Equal("post-patch", await ExecuteWorkflowAsync(postPatchPatchID));
+                Assert.Equal("post-patch", await QueryWorkflowAsync(deprecatePatchID));
+                var exc = await Assert.ThrowsAsync<WorkflowQueryFailedException>(
+                    () => QueryWorkflowAsync(prePatchID));
+                Assert.Contains("Nondeterminism", exc.Message);
+                // TODO(cretz): This currently causes a core panic
+                // exc = await Assert.ThrowsAsync<WorkflowQueryFailedException>(
+                //     () => QueryWorkflowAsync(patchedID));
+                // Assert.Contains("Nondeterminism", exc.Message);
```
Sushisource, line 2430: What panic? Did I fix this already?

Sushisource (approved):

src/Temporalio/Worker/WorkflowInstance.cs
cretz: This code actually looks more similar than it is. I would love to combine it, but `var origFunc = func` actually means different things in each (can't be combined), and therefore this calls `ContinueWith` with a different signature, and therefore the `return` has a completely different signature. I tried, but there actually isn't much that could be made common and save code here.

src/Temporalio/Worker/WorkflowInstance.cs
cretz: Hard to justify for single private use in the entire codebase, but if that bothersome I can.

tests/Temporalio.Tests/Worker/WorkflowWorkerTests.cs
cretz: Internal Slack thread at https://temporaltechnologies.slack.com/archives/C01FG4BRQVB/p1680099870171289. I didn't want to create an issue until after I updated core and confirmed the issue persists. So I was gonna revisit this after next PR to update core.

cretz:
@bergundy - I went ahead and merged this, but you can still review and I will apply the suggestions on next PR.

src/Temporalio/Activities/ActivityRefs.cs
```diff
@@ -0,0 +1,16 @@
+namespace Temporalio.Activities
+{
+    /// <summary>
+    /// Static helper for creating an instance of a type to reference its methods.
+    /// </summary>
+    public static class ActivityRefs
```
bergundy, line 6: Why do we need this `ActivityRefs` class, I thought we were going with `Refs`?

src/Temporalio/Worker/ReplaySafeLogger.cs
```diff
@@ -0,0 +1,39 @@
+using System;
+using Microsoft.Extensions.Logging;
+
+namespace Temporalio.Worker
+{
+    /// <summary>
+    /// Logger that will not log during replay.
+    /// </summary>
+    internal class ReplaySafeLogger : ILogger
```
bergundy, line 9: Wondering why you chose to put `Safe` in the name of this class, it's not about safety, just noise reduction.

src/Temporalio/Activities/ActivityRefs.cs
cretz: This is a good topic of discussion.

I was going with `Temporalio.Refs` but I was blinded by my samples and tests always being in some child of `Temporalio` namespace where I didn't have to import `Temporalio` specifically. But if you're in another namespace, you have to import `Temporalio` just to access refs and our goal is to make `Temporalio.Activities` or `Temporalio.Workflows` _mostly_ all you need when working in there. However, we are putting common things like `RetryPolicy` and such in `Temporalio` and I've _almost_ completed convincing myself that needs to be in `Temporalio.Common`.

So you can either have:

```
namespace MyNamespace;
using Temporalio;
using Temporalio.Activities;

public class MyActivities
{
    public static readonly MyActivities Ref = Refs.Create<MyActivities>();

    // ...
}
```

or

```
namespace MyNamespace;
using Temporalio.Activities;

public class MyActivities
{
    public static readonly MyActivities Ref = ActivityRefs.Create<MyActivities>();

    // ...
}
```

I figured this was the clearest at least right now. Subject to discussion of course.

src/Temporalio/Worker/ReplaySafeLogger.cs
cretz: It's a common English approach for things like this. Could also be `ReplayAware`. What "X safe" would mean is that it's specifically built to be safe from the caveats that come with `X`, e.g. other English uses like "memory safe", "fail safe", etc.

Coincidentally durable tasks used similar naming. This name for us is only internal (this is not public API), but theirs is external: https://learn.microsoft.com/en-us/dotnet/api/microsoft.azure.webjobs.extensions.durabletask.durablecontextextensions.createreplaysafelogger?view=azure-dotnet.

src/Temporalio/Workflows/ChildWorkflowSignalOptions.cs
bergundy: I guess the only reason why you'd want this is to be able to resolve the task returned to the user in case the token is cancelled internally, which is a good enough reason to keep this IMHO.
It makes the cancellation experience for the SDK's async APIs consistent.

bergundy (commented):
Skimmed this, LGTM

src/Temporalio/Worker/ReplaySafeLogger.cs
bergundy: I like the term `Aware` better but 🤷

src/Temporalio/Activities/ActivityRefs.cs
bergundy: My preference would be to expose a single `Ref` class but it's such a minor point that I won't press it.

cretz: We would have to force people to add a new import just for this in many cases.

PR #48, opened 2023-04-17
Workflow replayer, docs, time-skipping test server, rename ActivityContext, core update, and more

### LINKED ISSUE #36: [Bug] NuGet package is built with too-new of a glibc version
### Describe the bug

We use glibc 2.32 in the GH `ubuntu-latest` which is too new for older distros (like `mcr.microsoft.com/dotnet/aspnet:7.0` which defaults to Debian 11 which is 2.31). Downgrade glibc as old as possible. Ideally manually, but can use just older ubuntu version during build if that helps.

### LINKED ISSUE #41: [Bug] DataConverter should not eagerly instantiate payload/failure converter from type
### Describe the bug

Using `DataConverter.Default with { PayloadConverterType = typeof(MyPayloadConverter) }` means internally .NET does a clone (shallow using internal copy constructor) _then_ sets the property at which point it's too late to set the `PayloadConverter`. Can probably just make the payload/failure converter properties `Lazy` (`PublicationOnly` because it is expected that multiple instantiations of payload/failure converter can occur).

cretz:
## What was changed

* Added workflow docs to README
* Updated core (and needed regenerated pieces)
* Update CI to build on older Linux (fixes #36)
* Fix construction of data converter (fixes #41)
* Properly implement time-skipping test server
* Implement workflow replayer
* Rename `ActivityContext` to `ActivityExecutionContext`

Sorry so much was included. It was originally small, but difficulties mounted during PR. Can ignore generated code of course.


tests/Temporalio.Tests/Worker/WorkflowWorkerTests.cs
```diff
@@ -2363,7 +2343,7 @@ public class PostPatchWorkflow : PatchWorkflowBase
         }
     }
 
-    [Fact]
+    [Fact(Skip = "TODO(cretz): This is currently not throwing expected non-determinism errors in core")]
```
cretz: @Sushisource - this is the test that, on the code below that runs:

```csharp
                var exc = await Assert.ThrowsAsync<WorkflowQueryFailedException>(
                    () => QueryWorkflowAsync(prePatchID));
```

This is not throwing the non-determinism error one might expect even with latest core. This is testing that using query to replay pre-patch code on a post-patch workflow will cause non-determinism (activity type change).

Sushisource: OK, weird. I'll check this out - what's the easiest way for me to run just this test? I haven't learned much of the tooling for this yet.

cretz: Download the [`dotnet` binary](https://dotnet.microsoft.com/en-us/download), clone this repo recursively, then in this dir run `dotnet build` (which runs Rust build, so you'll need `protoc` and such), remove the `Skip` message from the attribute, and then you can do something like:

    dotnet run --project tests/Temporalio.Tests -- -verbose -method "*.ExecuteWorkflowAsync_Patched_ProperlyHandled"

That should be enough for you to see. Feel free to comment out many parts of this multi-step test. Feel free to hit me up with any snags.

src/Temporalio/Converters/DataConverter.cs
```diff
@@ -42,51 +42,95 @@ public record DataConverter<TPayloadConverterType>(IPayloadCodec? PayloadCodec =
...
+        // /// Initializes a new instance of the <see cref="DataConverter"/> class with default
+        // /// payload and failure converters.
+        // /// </summary>
+        // /// <remarks>
+        // /// Non-inheriting users should use <see cref="Default" /> instead.
+        // /// </remarks>
+        // public DataConverter()
+        //     : this(typeof(DefaultPayloadConverter), typeof(DefaultFailureConverter))
+        // {
+        // }
```
cretz:
```suggestion
```

.github/workflows/package.yml
```diff
@@ -55,20 +62,22 @@ jobs:
         if: ${{ matrix.alternative-target }}
         run: rustup target add ${{ matrix.alternative-target }}
 
-      - name: Install protoc (x64)
-        # Does not work on arm
-        if: ${{ matrix.os != 'ubuntu-arm' }}
+      - name: Install protoc (non-Linux)
+        # Does not work on Linux
+        if: ${{ matrix.os != 'ubuntu-latest' && matrix.os != 'ubuntu-arm' }}
```
bergundy, line 67: nit: use `contains(matrix.os, 'ubuntu')`.

src/Temporalio/Converters/DataConverter.cs
```diff
@@ -42,51 +42,95 @@ public record DataConverter<TPayloadConverterType>(IPayloadCodec? PayloadCodec =
...
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="DataConverter"/> class.
+        /// </summary>
+        /// <param name="payloadConverterType">Payload converter type. This must be a type and not
+        /// an instance so it can be serialized across a sandbox boundary.</param>
+        /// <param name="failureConverterType">Failure converter type. This must be a type and not
+        /// an instance so it can be serialized across a sandbox boundary.</param>
```
bergundy, line 56: Do you want to keep using types instead of instances?
Would it make sense to require a factory here so users can pass in config to their converters (given that it's mostly codecs that need to be configurable)?

src/Temporalio/Converters/DataConverter.cs
```diff
@@ -42,51 +42,95 @@ public record DataConverter<TPayloadConverterType>(IPayloadCodec? PayloadCodec =
...
+        /// <param name="payloadCodec">Optional payload codec.</param>
+        public DataConverter(
+            Type payloadConverterType,
+            Type failureConverterType,
```
bergundy, line 60: Can we make the failure converter optional?

src/Temporalio/Worker/WorkflowReplayer.cs
```diff
@@ -0,0 +1,270 @@
...
+        /// <param name="histories">Histories to replay.</param>
+        /// <param name="throwOnReplayFailure">If true, which is not the default, this will throw
+        /// <see cref="InvalidWorkflowOperationException" /> on a workflow task failure
+        /// (e.g. non-determinism) as soon as it's encountered. This has no effect on workflow
+        /// failures which are not reported as part of replay.</param>
+        /// <param name="cancellationToken">Cancellation token to stop replaying.</param>
+        /// <returns>Results of the replay runs.</returns>
+        public async IAsyncEnumerable<WorkflowReplayResult> ReplayWorkflowsAsync(
+            IAsyncEnumerable<WorkflowHistory> histories,
+            bool throwOnReplayFailure = false,
```
bergundy, line 103: Why do you need this flag?
I removed it in TS since the user can decide to throw while iterating throw the results.

src/Temporalio/Worker/WorkflowReplayer.cs
```diff
@@ -0,0 +1,270 @@
...
+        /// <see cref="InvalidWorkflowOperationException" /> on a workflow task failure
+        /// (e.g. non-determinism). This has no effect on workflow failures which are not reported
+        /// as part of replay.</param>
+        /// <returns>Result of the replay run.</returns>
+        public async Task<WorkflowReplayResult> ReplayWorkflowAsync(
+            WorkflowHistory history, bool throwOnReplayFailure = true) =>
+            (await ReplayWorkflowsAsync(new[] { history }, throwOnReplayFailure).ConfigureAwait(false)).Single();
+
+        /// <summary>
+        /// Replay multiple workflows from the given histories.
```
bergundy, line 55: Might want to direct users to the async API if they're on a new enough version of .NET.

src/Temporalio/WorkflowHistory.cs
```diff
@@ -8,5 +17,310 @@ namespace Temporalio
...
+                            Fix(fix, fieldIndex + 1, dict);
+                        }
+                        return;
+                    default:
+                        throw new JsonException(
+                            $"Got type {prop.GetType()} unexpectedly and field index {fieldIndex} for fix {fix}");
+                }
+            }
+
+            private static void CollectPathsToFix(
```
bergundy, line 199: Nice!

tests/Temporalio.Tests/Testing/WorkflowEnvironmentTests.cs
```diff
@@ -0,0 +1,217 @@
...
+using Temporalio.Testing;
+using Temporalio.Worker;
+using Temporalio.Workflows;
+using Xunit;
+using Xunit.Abstractions;
+
+public class WorkflowEnvironmentTests : TestBase
+{
+    // Time-skipping test server only runs on x86/x64
+    public sealed class OnlyIntelFactAttribute : FactAttribute
```
bergundy, line 20: nit: it's not only intel, amd has this arch too but 🤷

tests/Temporalio.Tests/WorkflowEnvironment.cs
```diff
@@ -34,6 +34,7 @@ public async Task InitializeAsync()
                 },
             },
         });
+        // env = new(await Temporalio.Client.TemporalClient.ConnectAsync(new("localhost:7233")));
```
bergundy, line 38: leftover?

bergundy (approved):
Did not do an in-depth review but overall LGTM.

src/Temporalio/Converters/DataConverter.cs
cretz: It's kinda annoying for users, but I like that even though I know this won't ever be sandboxed probably, it still forces them to think about how these are isolated. Because they can't state share in their payload or failure converters which is a good thing anyways (well, unless they use statics).

src/Temporalio/Converters/DataConverter.cs
cretz: It is effectively optional if they do `DataConverter.Default with { PayloadConverterType = typeof(MyPayloadConverter) }` which the README suggests.

src/Temporalio/Worker/WorkflowReplayer.cs
cretz: @Sushisource advocated for it on the Python side. Same for single workflow replay. Are you suggesting remove it from both, or have single workflow replay throw but multi-workflow replay not throw? I don't mind either way, I just want to be consistent.

src/Temporalio/Worker/WorkflowReplayer.cs
cretz: I think it's clear enough when looking at API docs. And the iterable passed in is clear enough I think.

src/Temporalio/WorkflowHistory.cs
cretz: Yeah, but I should probably move it to compile time code generator. But I figured lazy on first use is ok to start with.

tests/Temporalio.Tests/Testing/WorkflowEnvironmentTests.cs
cretz: "Intel" is a common term for the instruction set, even though it's AMD's 64-bit extensions. I just used the word they are using to describe it at https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.architecture. This is test only code anyways. I could changed to `OnlyIntelBasedFactAttrbiute` or something.

tests/Temporalio.Tests/WorkflowEnvironment.cs
cretz: Intentionally so, because I come here and run against local server so I have a UI (I used it to download those JSONs for the replayer tests). But I can make a comment to that effect.

tests/Temporalio.Tests/Worker/WorkflowWorkerTests.cs
Sushisource: Turns out this is because the version of server this SDK is using for tests doesn't support the sdk metadata and hence internal flags feature

cretz: :+1: I will change the skip message to be more clear. I have opened https://github.com/temporalio/sdk-dotnet/issues/50. Semi-relatedly, I have also opened https://github.com/temporalio/sdk-java/issues/1746.

src/Temporalio/Converters/DataConverter.cs
cretz: Fixed

src/Temporalio/Worker/WorkflowReplayer.cs
bergundy: Suggesting that single replay throws and multiple (async iterator) doesn't.
We can give the option for the multi replay method that replays synchronously.

cretz: > Suggesting that single replay throws and multiple (async iterator) doesn't.

That's already the case by default. Are you saying both need to disallow changing the default? Or just one have the option of changing the default?

> We can give the option for the multi replay method that replays synchronously.

They all replay synchronously in this sense. There is no concurrent replay.

bergundy: I don't think we need the option in the async iterator method.

> They all replay synchronously in this sense. There is no concurrent replay.

I was talking about the multi replay method that returns `task<iterator>`.
I'm of course familiar with how the replayer works.

cretz: > I don't think we need the option in the async iterator method.

This argues for inconsistency. Why is a `foreach` ok to throw but not a `await foreach`?

bergundy: Because in the `task<iterator>` method you can't get the results one by one, and it's nice to be able to fail fast.
In the async iterator method, the user has full control and can abort whenever they like.

cretz: > Because in the `task<iterator>` method you can't get the results one by one

You can. The reason for the async iterator in this case is because I allow async iteration input. You can have either:

```csharp
foreach (var result in await replayer.ReplayWorkflowsAsync(syncHistIter))
{
    // Process 1 by t1
}
```
or
```csharp
await foreach (var result in replayer.ReplayWorkflowsAsync(asyncHyncHistIter))
{
    // Process 1 by 1
}
```

And technically the latter doesn't work on all .NET versions we support.

PR #67, opened 2023-05-16
Remove sandbox strictness and make more hackable (💥 BREAKING CHANGES)

cretz:
## What was changed

We have given up hope that we'll be able to sandbox .NET, so we made things easier for developers in some spots and more hackable in others. We also made it bit more hackable. Specifically:

* `TemporalWorkerOptions.Activities` and `TemporalWorkerOptions.Workflows` now are collections of definitions instead of delegates/types from before (same for workflow replayer options)
  * 💥BREAKING CHANGE - Users need to use `AddActivity` and `AddWorkflow` to add activities and workflows to options now instead of directly setting list
  * 💥BREAKING CHANGE - `AdditionalActivityDefinitions` and `AdditionalWorkflowDefinitions` removed
* `ActivityDefinition` creators reworked with an additional overload for custom invoker
  * 💥BREAKING CHANGE - Now called `Create` instead of `FromDelegate`
* `WorkflowDefinition` creators reworked with an additional overload for custom creator
  * 💥BREAKING CHANGE - Now called `Create` instead of `FromType`
* Allow workflow interceptors to be statefully created like activity interceptors
  * 💥BREAKING CHANGE - `IWorkerInterceptor` changed `WorkflowInboundInterceptorType` type to `InterceptWorkflow` function
  * 💥BREAKING CHANGE - `Interceptors` for client and worker options are now `IReadOnlyCollection` instead of  `IEnumerable`
* Data converter greatly simplified to just accept payload/failure converter instances instead of types
  * 💥BREAKING CHANGE - `DataConverter` no longer has generic version or accepts payload/failure converter types, only instances directly
* Updated README with needed changes


src/Temporalio/Activities/ActivityDefinition.cs
```diff
@@ -24,93 +37,232 @@ private ActivityDefinition(string name, Delegate del)
...
+        /// called on this delegate. The delegate must have an associated method and that method
+        /// must have <see cref="ActivityAttribute" /> set on it.
         /// </summary>
-        /// <param name="name">Activity name.</param>
-        /// <param name="del">Activity delegate.</param>
-        /// <returns>Activity definition.</returns>
-        /// <seealso cref="FromDelegate" />
-        public static ActivityDefinition CreateWithoutAttribute(string name, Delegate del)
+        /// <param name="del">Delegate to create definition from.</param>
+        /// <param name="cache">True if this should cache the result making successive invocations
```
Sushisource, line 63: Maybe elaborate on why anyone would set this false

src/Temporalio/Activities/ActivityDefinition.cs
```diff
@@ -24,93 +37,232 @@ private ActivityDefinition(string name, Delegate del)
...
+            var name = attr.Name;
+            if (name != null)
+            {
+                return name;
+            }
+            // Build name from method name
+            name = method.Name;
+            // Local functions are in the form <parent>g__name|other, so we will try to
+            // extract the name
+            var localBegin = name.IndexOf(">g__");
```
Sushisource, line 242: Oh boy this is nasty. Not that there's anything to be done I guess, but, eww.

src/Temporalio/Workflows/WorkflowDefinition.cs
```diff
@@ -265,35 +282,41 @@ public static WorkflowDefinition CreateWithoutAttribute(string name, Type type)
...
-                if (type.IsInterface && name.Length > 1 && name[0] == 'I' && char.IsUpper(name[1]))
-                {
-                    name = name.Substring(1);
-                }
+                throw new ArgumentException($"{runMethod} missing WorkflowRun attribute");
             }
+            // We intentionally use reflected type because we don't allow inheritance of run methods
+            // in any way, they must be explicitly defined on the type
+            return Create(runMethod.ReflectedType ??
+                throw new ArgumentException($"{runMethod} has no reflected type"));
```
Sushisource, line 305: Would be nice to have the error message make reference to how to fix this

Sushisource (approved):
Only small things

src/Temporalio/Activities/ActivityDefinition.cs
cretz: I almost didn't expose it, but technically I guess there can be a dynamic delegate but I don't want to even mention that. This parameter could be called "memoize" and therefore it is probably implied why a user may not want to store the result.

src/Temporalio/Activities/ActivityDefinition.cs
cretz: This is where I'm relying on unit tests. And note this is only for local functions which are very rare to use in .NET.

src/Temporalio/Workflows/WorkflowDefinition.cs
cretz: It's a very rare case, see https://learn.microsoft.com/en-us/dotnet/api/system.reflection.memberinfo.reflectedtype?view=net-7.0#remarks. It's only on module-level methods which you can't even write in C# (but the .NET IL technically supports it).

PR #69, opened 2023-05-16
Simple nightly benchmarker

### LINKED ISSUE #54: [Feature Request] Simple benchmarker that runs nightly

Need something like what is in Python that does simple benchmarks to see performance in certain situations

cretz:
## What was changed

Just a simple benchmarker similar to Python SDK until https://github.com/temporalio/omes is .NET ready and used more heavily.

Early results show that, without getting into concurrency details, Python can do ~80 workflows/sec on local Temporalite and a 10k workflow cache reaches ~630MiB process mem. .NET can do ~130 workflows/sec and a 10K workflow cache reaches ~390MiB. But not all workflows are cached at once (.NET evicts faster because it runs faster) and the real bottleneck is Temporalite.

The (accomplished) goal of this is to prove .NET is plenty fast, just not how fast (that will come with external load testing).

Closes #54 (ignore CI failures, #68 will help there).


.github/workflows/run-bench.yml
```diff
@@ -0,0 +1,56 @@
...
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Build
+        run: dotnet build --configuration Release
+
+      # Run a bunch of bench tests. We run multiple times since results vary.
+
+      - run: dotnet run --configuration Release --project tests/Temporalio.SimpleBench/Temporalio.SimpleBench.csproj -- --workflow-count 100 --max-cached-workflows 100 --max-concurrent 100
+      - run: dotnet run --configuration Release --project tests/Temporalio.SimpleBench/Temporalio.SimpleBench.csproj -- --workflow-count 100 --max-cached-workflows 100 --max-concurrent 100
+      - run: dotnet run --configuration Release --project tests/Temporalio.SimpleBench/Temporalio.SimpleBench.csproj -- --workflow-count 100 --max-cached-workflows 100 --max-concurrent 100
```
Sushisource, line 42: Not that it matters much but it might be nice to move the number of runs inside the bencher. Then you can also emit avg/variance

Sushisource (approved):

.github/workflows/run-bench.yml
cretz: Yeah, this just kinda matches the Python one. Both of which are very limited and need to be replaced with a more substantial option (e.g. omes).

PR #75, opened 2023-05-24
Use expression trees for all type-safe remote calls (💥 BREAKING CHANGES) 

### LINKED ISSUE #62: [Feature Request] Change Ref approach to expression-based

Akin to Hangfire and others, use LINQ expressions to let calls be made. This is more inline with how other .NET libraries work. We need to make sure status still work and we can extract return types in `Task` vs non-`Task`. This will also require updating docs and the blog post.

Also, as part of this effort, we need to modernize registration of activities and workflows. We probably need to:

* Make definitions the only collections on worker options instead of the "additional"
* Allow activities to also be registered by entire type/class that crawls looking for activity attributed 
* Abstract activity definition a tad to accept a form with arg types, return type, and a `Func<object?, Task<object?[]?>>` invoker. Put an `InvokeAsync` on it I guess. This is needed for dependency injection support. Would accept alternative activity class factory approach if we can think of it, though this is probably good enough. Can expose this all the way out to the `AddActivity` overload in the worker options.

cretz:
## What was changed

We have removed the concept of "Ref" and replaced with [expression trees](https://learn.microsoft.com/en-us/dotnet/csharp/advanced-topics/expression-trees/). Primary reasons:

* "Ref" approach was too unfamiliar to .NET devs
* "Ref" approach required implementers to add something when they implemented
* Expression trees are more familiar even if they have slight shortcomings (e.g. performance of interpreting complex arg expressions and not everything like optional/named parameters accepted)
* Expression trees allow for fewer overloads to support 0, 1, N-arg calls
* Expression trees allow us to store the type of workflow on the workflow handle, child workflow handle, and external workflow handle allowing us to enforce that type-safe signal and query calls are on the same workflow

💥 BREAKING CHANGES:

The following calls were changed to support expression trees instead of "Ref":

* `Temporalio.Client.ITemporalClient.ExecuteWorkflowAsync` (extension)
* `Temporalio.Client.ITemporalClient.StartWorkflowAsync`
* `Temporalio.Client.Schedules.ScheduleActionStartWorkflow.Create`
* `Temporalio.Client.WorkflowHandle.QueryAsync`
* `Temporalio.Client.WorkflowHandle.SignalAsync`
* `Temporalio.Workflows.ChildWorkflowHandle.SignalAsync`
* `Temporalio.Workflows.ExternalWorkflowHandle.SignalAsync`
* `Temporalio.Workflows.Workflow.CreateContinueAsNewException`
* `Temporalio.Workflows.Workflow.ExecuteActivityAsync`
* `Temporalio.Workflows.Workflow.ExecuteChildWorkflowAsync`
* `Temporalio.Workflows.Workflow.ExecuteLocalActivityAsync`
* `Temporalio.Workflows.Workflow.StartChildWorkflowAsync`

Any previous use of these is now likely as compiler error. To get an idea of what the changes look like from a caller POV, see the `tests/Temporalio.Tests/Worker/WorkflowWorkerTests.cs` file.

## Checklist

1. Closes #62


src/Temporalio/Common/ExpressionUtil.cs
cretz, line 1: Note, in a subsequent PR, I will be moving all `Temporalio` top-namespace-level types to `Temporalio.Common`, this just happens to be the first

bergundy (approved):
Skimmed this, overall LGTM

PR #81, opened 2023-06-01
Add Semgrep scanning

jackdawm:
## What was changed
Adding a Semgrep scanning workflow.

## Why?
To do static analysis for security vulnerabilities in our SDKs

## Checklist

How was this tested:
Ran this in private repos with good results and no problems


cretz:
Note, semgrep will not ever fail currently if bad code is found.

cretz (approved):
Approving with the guarantee that this check will not be able to fail without discussion with the repo owners first (there are several things to consider).

PR #83, opened 2023-06-05
OpenTelemetry Support

### LINKED ISSUE #56: [Feature Request] OpenTelemetry support

Need OpenTelemetry interceptor as separately published project

cretz:
## What was changed

Added a separate project for OpenTelemetry support. Since .NET has built-in tracing abstractions, I had to write a long README explaining why workflow tracing has problems with traditional .NET/OTel tracing.

## Checklist

1. Closes #56


src/Temporalio.Extensions.OpenTelemetry/README.md
```diff
@@ -0,0 +1,210 @@
...
+
+Diagnostic activities created for signals and queries are parented to the client-outbound diagnostic activity that
+started the workflow, and only _linked_ to the client-outbound diagnostic activity that invoked the signal/query.
+
+If a workflow fails or if a workflow task fails (i.e. workflow suspension due to workflow/signal exception that was not
+a Temporal exception), a _new_ diagnostic activity is created representing that failure. The diagnostic activity
+representing the run of the workflow is not only already completed (as they all are) but it may not even be created
+because this could be replaying on a different worker.
+
+Outbound calls from a workflow for scheduling activity, scheduling local activity, starting a child, signalling a child,
```
Sushisource:
```suggestion
Outbound calls from a workflow for scheduling an activity, scheduling a local activity, starting a child, signalling a child,
```

src/Temporalio.Extensions.OpenTelemetry/ActivitySourceExtensions.cs
```diff
@@ -0,0 +1,93 @@
...
+        /// <returns>The started workflow activity set as current.</returns>
+        /// <remarks>
+        /// This starts and immediately stops a <see cref="Activity" /> since workflows do not
+        /// support long-running .NET activities because workflows are interruptible/resumable. This
+        /// method should be used instead of any other call on <see cref="ActivitySource" /> inside
+        /// of workflows as it is replay safe. This will always return a value even if there is no
+        /// .NET activity created (e.g. no listener, during replay, etc). Users must call
+        /// <c>Dispose</c> on the activity to remove it from the current context.
+        /// </remarks>
+        public static WorkflowActivity StartWorkflowActivity(
```
Sushisource: The overlap between these diagnostic activities and our activities is kinda rough. It certainly confused me at first. Maybe we should call this `WorkflowDiagnosticActivity` or `WorkflowTracingActivity`? Bit of a mouthful but at least a bit more clear?

Sushisource (commented):
I didn't closely review the actual implementation code yet, but one question about the naming here that I'm worried could get confusing

src/Temporalio.Extensions.OpenTelemetry/ActivitySourceExtensions.cs
prom3theu5: I Agree - I think there should be some distinction between those methods, they are a little analogous. Perhaps something akin to the application insights library, where you have TrackEvent - maybe StartTrackedWorkflowActivity, or just TrackWorkflowActivity?

johnkattenhorn: I agree, you probably should avoid anything that starts with `Workflow`, for me `Track` or `Trace` works for me, I'm leaning toward `Trace`due to matching with the usual 4 categories of observability eg. `Logs, Metrics, Traces and Dependencies.`

cretz: While that overlapping is kinda rough I agree, this is an extension method on `ActivitySource`, not right next to the `Workflow` class. And this is only used by people that manually add spans to their workflows which we've found to be rare across our SDKs.

> It certainly confused me at first. Maybe we should call this WorkflowDiagnosticActivity or WorkflowTracingActivity? Bit of a mouthful but at least a bit more clear?

Concur. I like `WorkflowDiagnosticActivity` since it's a wrapper around `System.Diagnostic.ActivityContext` and `System.Diagnostic.Activity`.

> Perhaps something akin to the application insights library, where you have TrackEvent - maybe StartTrackedWorkflowActivity, or just TrackWorkflowActivity?

Concur. I never did like "Start" because it implies we are starting like an activity source starts, but really we start-then-stop immediately and just set it on context. I like `Track` as a verb over start.

> I agree, you probably should avoid anything that starts with Workflow

We need to make this clear it's only for workflows and not for other uses.

So I'm gonna change the method name to `TrackWorkflowDiagnosticActivity` and change the class name to `WorkflowDiagnosticActivity`.

src/Temporalio.Extensions.OpenTelemetry/README.md
cretz: Fixed.

src/Temporalio.Extensions.OpenTelemetry/ActivitySourceExtensions.cs
cretz: Updated

bergundy (approved):
I'm stamping this based on our discussions, did not read the code.

PR #87, opened 2023-06-13
Return collection instead of enumerable and fix/re-enable task scheduler check (💥 BREAKING CHANGES)

cretz:
## What was changed

* Change converter/codec/etc non-streaming `IEnumerable` return values to `IReadOnlyCollection`
  * Lazy enumerables for this case were confusing users and are against best practices
  * 💥 BREAKING CHANGE on some return types
* Fix issue that inadvertently disabled the task-on-workflow-scheduler checker (i.e. re-enable it)
  * 💥 BREAKING CHANGE on behavior because for the last couple of alphas this checker was inadvertently disabled
* Provided workaround for [`Task.WhenAny` sometimes using thread pool](https://github.com/dotnet/runtime/issues/87481)
  * Updated README and made `Workflow.WhenAnyAsync` that is workflow safe

### DIFF
--- a/README.md
+++ b/README.md
@@ -689,10 +689,15 @@ with .NET tasks inside of workflows:
 * Do not use `Task.Delay`, `Task.Wait`, timeout-based `CancellationTokenSource`, or anything that uses .NET built-in
   timers.
   * `Workflow.DelayAsync`, `Workflow.WaitConditionAsync`, or non-timeout-based cancellation token source is suggested.
+* Do not use `Task.WhenAny`.
+  * Use `Workflow.WhenAnyAsync` instead.
+  * Technically this only applies to an enumerable set of tasks with results or more than 2 tasks with results. Other
+    uses are safe. See [this issue](https://github.com/dotnet/runtime/issues/87481).
 * Be wary of additional libraries' implicit use of the default scheduler.
   * For example, while there are articles for `Dataflow` about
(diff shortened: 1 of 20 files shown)


README.md
```diff
@@ -689,10 +689,15 @@ with .NET tasks inside of workflows:
 * Do not use `Task.Delay`, `Task.Wait`, timeout-based `CancellationTokenSource`, or anything that uses .NET built-in
   timers.
   * `Workflow.DelayAsync`, `Workflow.WaitConditionAsync`, or non-timeout-based cancellation token source is suggested.
+* Do not use `Task.WhenAny`.
+  * Use `Workflow.WhenAnyAsync` instead.
```
bergundy, line 693: Does this method require the `Async` suffix? Wouldn't it be better to match the `Task` API naming?

bergundy (approved):

README.md
cretz: I chose `Workflow.DelayAsync` even though `Task.Delay`, so I did the same here. I think following [VSTHRD200](https://github.com/microsoft/vs-threading/blob/main/doc/analyzers/VSTHRD200.md) everywhere is better than trying to match `Task` name exactly for the couple of alternatives we're providing.

bergundy: I'd make exceptions here just because we're providing alternatives for known APIs.

cretz: I don't think we should (and didn't when we introduced `DelayAsync`). In fact, I suspect .NET regrets not suffixing these with `Async` back when they wrote it.

PR #88, opened 2023-06-13
Support query properties and disallow static signals/queries

### LINKED ISSUE #85: [Feature Request] Re-investigate signals/queries/activities as properties

We initially didn't allow signals or queries as properties because we required delegates to invoke them. But now that we're using expressions we probably can. Specifically, we may want to support:

* `[WorkflowSignal]` on properties with setters where the setter is invoked for the signal
* `[WorkflowQuery]` on properties where the getter is invoked for the query

When `[WorkflowUpdate]` comes around, setting it on a property will invoke the setter and return the getter

### LINKED ISSUE #86: [Feature Request] Disallow workflow attributes on anything static

Workflow methods (and the workflow class itself) should not be static.

cretz:
## What was changed

* Allow properties to be queries, flows very natural with .NET
* Disallow signals/queries as static (workflow run methods already can't be static)

## Checklist

1. Closes #85
2. Closes #86


Sushisource (approved):
Neato

PR #89, opened 2023-06-15
Dynamic workflows, activities, signals, and queries

### LINKED ISSUE #27: [Feature Request] Support dynamic workflows, activities, signals, and queries

cretz:
## What was changed

* Added support for `[Activity(Dynamic = true)]` which gets called on unmatched activity in worker with single `IRawValue[]` parameter
* Added support for `[Workflow(Dynamic = true)]` which gets called on unmatched workflow in worker with single `IRawValue[]` parameter
* Added support for `[WorkflowSignal(Dynamic = true)]` which gets called on unmatched signal in workflow with string signal name and `IRawValue[]` parameter
* Added support for `[WorkflowQuery(Dynamic = true)]` which gets called on unmatched query in workflow with string query name and `IRawValue[]` parameter
* Added `ActivityExecutionContext.PayloadConverter` for use by activities
* Added `Workflow.PayloadConverter` for use by workflows
* Fixed double-encoding issue for schedule workflow updates
* Added `IPayloadConverter.ToRawValue` and `IPayloadConverter.ToValue<T>` extensions to assist in conversion to/from raw values respectively
* Removed embedded payload converter from `RawValue`/`IRawValue` - we make the converters otherwise available and this was needed because people may need to construct these values and the converter was only there for convenience
* Added `Workflow.DynamicSignal` and `Workflow.DynamicQuery` as properties that can be set from within a workflow

## Checklist

1. Closes #27

### DIFF
--- a/src/Temporalio/Worker/ActivityWorker.cs
+++ b/src/Temporalio/Worker/ActivityWorker.cs
@@ -25,6 +25,7 @@ namespace Temporalio.Worker
         private readonly ILogger logger;
         // Keyed by name
         private readonly Dictionary<string, ActivityDefinition> activities;
+        private readonly ActivityDefinition? dynamicActivity;
         // Keyed by task token
         private readonly ConcurrentDictionary<ByteString, RunningActivity> runningActivities = new();
 
@@ -41,11 +42,22 @@ namespace Temporalio.Worker
             activities = new(worker.Options.Activities.Count);
             foreach (var defn in worker.Options.Activities)
             {
-                if (activities.ContainsKey(defn.Name))
+                if (defn.Name == null)
+                {
+                    if (dynamicActivity != null)
+                    {
+                        throw new ArgumentException("Multiple dynamic activities provided");
+                    }
+                    dynamicActivity = defn;
+                }
+                else if (activities.ContainsKey(defn.Name))
                 {
                     throw new ArgumentException($"Duplicate activity named {defn.Name}");
                 }
-                activities[defn.Name] = defn;
+                else
+                {
+                    activities[defn.Name] = defn;
+                }
             }
         }
 
@@ -280,32 +292,49 @@ namespace Temporalio.Worker
                 // Find activity or fail
                 if (!activities.TryGetValue(act.Context.Info.ActivityType, out var defn))
                 {
-                    var avail = activities.Keys.ToList();
-                    avail.Sort();
-                    var availStr = string.Join(", ", avail);
-                    throw new ApplicationFailureException(
-                        $"Activity {act.Context.Info.ActivityType} is not registered on this worker," +
-                        $" available activities: {availStr}",
-                        errorType: "NotFoundError");
+                    defn = dynamicActivity;
+                    if (defn == null)
+                    {
+                        var avail = activities.Keys.ToList();
+                        avail.Sort();
+                        var availStr = string.Join(", ", avail);
+                        throw new ApplicationFailureException(
+                            $"Activity {act.Context.Info.ActivityType} is not registered on this worker," +
+                            $" available activities: {availStr}",
+                            errorType: "NotFoundError");
+                    }
                 }
 
-                // Deserialize arguments. If the input is less than the required parameter count, we
-                // error.
-                if (tsk.Start.Input.Count < defn.RequiredParameterCount)
-                {
-                    throw new ApplicationFailureException(
-                        $"Activity {act.Context.Info.ActivityType} given {tsk.Start.Input.Count} parameter(s)," +
-                        " but more than that are required by the signature");
-                }
-                // Zip the params and input and then decode each. It is intentional that we discard
-                // extra input arguments that the signature doesn't accept.
                 object?[] paramVals;
                 try
                 {
-                    paramVals = await Task.WhenAll(
-                        tsk.Start.Input.Zip(defn.ParameterTypes, (input, paramType) =>
-                            worker.Client.Options.DataConverter.ToValueAsync(
-                                input, paramType))).ConfigureAwait(false);
+                    // Dynamic activities are just one param of a raw array, otherwise, deserialize
+                    if (defn.Dynamic)
+                    {
+                        paramVals = new[]
+                        {
+                            await Task.WhenAll(tsk.Start.Input.Select(p =>
+                                worker.Client.Options.DataConverter.ToValueAsync<IRawValue>(p))).
+                                ConfigureAwait(false),
+                        };
+                    }
+                    else
+                    {
+                        // Deserialize arguments. If the input is less than the required parameter
+                        // count, we error.
+                        if (tsk.Start.Input.Count < defn.RequiredParameterCount)
+                        {
+                            throw new ApplicationFailureException(
+                                $"Activity {act.Context.Info.ActivityType} given {tsk.Start.Input.Count} parameter(s)," +
+                                " but more than that are required by the signature");
+                        }
+                        // Zip the params and input and then decode each. It is intentional that we
+                        // discard extra input arguments that the signature doesn't accept.
+                        paramVals = await Task.WhenAll(
+                            tsk.Start.Input.Zip(defn.ParameterTypes, (input, paramType) =>
+                                worker.Client.Options.DataConverter.ToValueAsync(
+                                    input, paramType))).ConfigureAwait(false);
+                    }
                 }
                 catch (Exception e)
                 {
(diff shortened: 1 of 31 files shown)


src/Temporalio/Worker/ActivityWorker.cs
```diff
@@ -279,32 +292,49 @@ private void StartActivity(Bridge.Api.ActivityTask.ActivityTask tsk)
...
-                        errorType: "NotFoundError");
+                    defn = dynamicActivity;
+                    if (defn == null)
+                    {
+                        var avail = activities.Keys.ToList();
+                        avail.Sort();
+                        var availStr = string.Join(", ", avail);
+                        throw new ApplicationFailureException(
+                            $"Activity {act.Context.Info.ActivityType} is not registered on this worker," +
+                            $" available activities: {availStr}",
```
mjameswh, line 303: (Not related to this specific PR) Just a reminder that listing activities in failure details is now problematic. Server will report warning at 2KB, and truncate the Failure if exceeding 4KB ([see code](https://github.com/temporalio/temporal/pull/4338/files#diff-8b569e7d448b915cfc54cba06dfa7dfa8d639218eab08286169ede455af3dd10R3989-R4023).

src/Temporalio/Worker/WorkflowInstance.cs
```diff
@@ -169,6 +178,42 @@ public WorkflowInstance(WorkflowInstanceDetails details)
...
         public CancellationToken CancellationToken => cancellationTokenSource.Token;
 
+        /// <inheritdoc />
+        public WorkflowQueryDefinition? DynamicQuery
+        {
+            get => dynamicQuery;
+            set
+            {
+                if (value != null && !value.Dynamic)
+                {
```
mjameswh, line 188: Why require that the dynamic query function has the Dynamic attribute if they are explicitely assigning to `DynamicQuery`? Seems redundant...

src/Temporalio/Worker/WorkflowInstance.cs
```diff
@@ -169,6 +178,42 @@ public WorkflowInstance(WorkflowInstanceDetails details)
...
+                    throw new ArgumentException("Signal is not dynamic");
+                }
+                dynamicSignal = value;
+                if (value != null)
+                {
+                    // If it's not null, send _all_ buffered signals. We will copy all from
+                    // buffered, clear buffered, and send all to apply signal
+                    var signals = bufferedSignals.ToList();
```
mjameswh, line 210: This is technically incorrect, as the Workflow could change/unregister the DynamicSignal handler or register named signal handler from a signal handler, causing reentrant calls to either this loop or the one in named query handler registration. In these case, ApplySignalWorkflow might push back signals to the queue, which might scramble buffered signals relative order. See [this commit](https://github.com/temporalio/sdk-typescript/pull/1038/files#diff-edcae2d23017bc07e3a859b68e980d151947323c04128f2bc08147bda4a1d7deR1971) for example cases.

This should instead be something like:

```
    while (signal = bufferedSignals.tryTakeFirst()) {
        ApplySignalWorkflow(signal);
    }
```

src/Temporalio/Worker/ActivityWorker.cs
cretz: Ug, yeah that's the size of the entire failure proto - message, stack trace, recursive causes and all. What should we do for big stack traces? What should we do for large cause chains?

Seems it's not listing activities is problematic, but large failures are problematic and I wonder if we need a holistic solution. How am I supposed to know whether I can list more or less activities just because the stack trace happens to be smaller?

src/Temporalio/Worker/WorkflowInstance.cs
cretz: They are different things. `[WorkflowQuery]` is declaration-site method made into a query. But we _also_ offer (for DSLs and other use cases) a way to set handlers from code at runtime of the workflow. `[Workflow(Dynamic = true)]` is the recommended way, but `DynamicQuery` property is an optional runtime alternative.

src/Temporalio/Worker/WorkflowInstance.cs
```diff
@@ -938,21 +997,49 @@ private void OnQueryDefinitionAdded(string name, WorkflowQueryDefinition defn)
...
             if (name != defn.Name)
             {
                 throw new ArgumentException($"Signal name {name} doesn't match definition name {defn.Name}");
             }
-            // Send all buffered signals
-            if (bufferedSignals.TryGetValue(name, out var buffered))
+            // Send all buffered signals that apply to this one. We are going to collect these by
+            // iterating list in reverse matching predicate while removing, then iterating in
+            // reverse
+            var buffered = new List<SignalWorkflow>();
```
mjameswh, line 1011: See related comment above

src/Temporalio/Worker/ActivityWorker.cs
mjameswh: Indeed, the problem is not specifically with listing activity names. Yet, the fact that the server may penalize us on for sending the list of activity names if that exceeds some threshold means that this is not a reliable way to provide that kind of feedback to user. We'll need to find an alternative for listing registered activity/workflow/signal/queries types AND a better way to deal with activities failures containing large stack traces.

Also, I'd say that Failure should not contain stack trace for these kind of errors, if that's possible. There is no user code on the stack at this point, so nothing they should need to trace.

But really, that don't need to be resolved today.

src/Temporalio/Worker/WorkflowInstance.cs
mjameswh: Yeah, I got that. What I'm saying is that if the user is explicitly calling the setter to set handler at runtime, why is it required that provided function do have the `Dynamic` attribute?

mjameswh (approved):
A few minor comments. Only one thing that really need to be addressed (risk of mangling buffered signals if signal handlers are added/modified/removed from a signal handler (ie. reentrant case)), but overall LGTM.

src/Temporalio/Worker/ActivityWorker.cs
cretz: > server may penalize us on for sending the list of activity names if that exceeds some threshold

Yeah, I need to probably put some upper bound on this

> We'll need to find an alternative for listing registered activity/workflow/signal/queries types 

https://github.com/temporalio/features/issues/51

> I'd say that Failure should not contain stack trace for these kind of errors, if that's possible

It's possible, but not very transparent. As an open source library I like showing where all exceptions occur even in our own code.

src/Temporalio/Worker/WorkflowInstance.cs
cretz: It is not required to have that (or even have an attribute). This is a reference to a property getter that just checks the absence of the name, it is completed unrelated to the attribute. From the test for that setter:

```csharp
            Workflow.DynamicQuery = WorkflowQueryDefinition.CreateWithoutAttribute(
                null, (string queryName, IRawValue[] args) =>
                {
                    Events.Add($"query-{queryName}: " +
                        Workflow.PayloadConverter.ToValue<string>(args.Single()));
                    return "done";
                });
```

src/Temporalio/Worker/WorkflowInstance.cs
cretz: I think the approach you show there is more confusing but no more accurate. I intentionally take all of the signals off _at handler registration time_ and clear the list before I even start processing them. This is very clear logic to the user setting the handler. You set a signal handler, your handler gets all _buffered signals_ guaranteed. No maybe about it. Same for setting a named handler. It's the same problem that exists in many places when mutating a collection while iterating it and is exactly why we flush the collection _before_ iterating on it.

With my logic, it is clear what happens when, say, you have psuedo code:

```
SetDynamicSignalHandler(HandleDynamicSignals)
SetSpecificSignalHandler("Signal1", HandleSignal1Signals)
```

The first one gets all buffered signals and the second one gets 0 buffered signals. I have documented this behavior and it is clear. But if I had to change the documentation to "start the draining process from the beginning when registered" it would be much less clear.

If you change handlers inside of some code somewhere else, doesn't affect that you chose to have all your buffered signals delivered to the original one since you know that is a registration time behavior.

src/Temporalio/Worker/WorkflowInstance.cs
cretz: The way signals are drained is very much by intention. It is confusing behavior to remove from this one by one.

src/Temporalio/Worker/WorkflowInstance.cs
cretz: We also start all of these tasks immediately, just like if a workflow task came in with all of these signals present on it

PR #91, opened 2023-06-26
Remove build information from CI workflow

jackdawm:
## What was changed
Remove unused log line that contains an injection vulnerability.

## Why?
To remove the injection vulnerability (despite it not being an auto-run risk, humans are fallible)

## Checklist

How was this tested:
Running dotnet test locally and watching the github workflows (one linter for ubuntu-arm failed because it timed out due to no workers, but nothing should have changed to cause that to fail).


cretz (approved):

cretz:
Thanks!

PR #93, opened 2023-06-27
Random with 64-bit seed

### LINKED ISSUE #90: [Feature Request] Random with larger seed

Maybe use https://docs.rs/rand/latest/rand/trait.SeedableRng.html#method.seed_from_u64. Will be **💥 BREAKING CHANGE**.

cretz:
## What was changed

.NET PRNG only accepts 32-bit seeds. This changes the workflow random to use a Rust-based random which supports 64-bit seeds.

## Checklist

1. Closes #90

### DIFF
--- a/src/Temporalio/Bridge/Cargo.toml
+++ b/src/Temporalio/Bridge/Cargo.toml
@@ -15,6 +15,11 @@ log = "0.4"
 parking_lot = "0.12"
 prost = "0.11"
 prost-types = "0.11"
+# We rely on Cargo semver rules not updating a 0.x to 0.y. Per the rand
+# documentation, before 1.0, minor 0.x updates _can_ break portability which can
+# cause non-determinism.
+rand = "0.8.5"
+rand_pcg = "0.3.1"
 temporal-client = { version = "0.1.0", path = "./sdk-core/client" }
 temporal-sdk-core = { version = "0.1.0", path = "./sdk-core/core" }
 temporal-sdk-core-api = { version = "0.1.0", path = "./sdk-core/core-api" }
--- a/src/Temporalio/Worker/WorkflowInstance.cs
+++ b/src/Temporalio/Worker/WorkflowInstance.cs
@@ -157,8 +157,7 @@ namespace Temporalio.Worker
             pendingTaskStackTraces = workflowStackTrace == WorkflowStackTrace.None ? null : new();
             logger = details.LoggerFactory.CreateLogger($"Temporalio.Workflow:{start.WorkflowType}");
             replaySafeLogger = new(logger);
-            // We accept overflowing for seed (uint64 -> int32)
-            Random = new(unchecked((int)details.Start.RandomnessSeed));
+            Random = new(details.Start.RandomnessSeed);
             TracingEventsEnabled = !details.DisableTracingEvents;
         }
 
@@ -233,7 +232,7 @@ namespace Temporalio.Worker
         public IDictionary<string, WorkflowQueryDefinition> Queries => mutableQueries.Value;
 
         /// <inheritdoc />
-        public Random Random { get; private set; }
+        public DeterministicRandom Random { get; private set; }
 
         /// <inheritdoc />
         public IDictionary<string, WorkflowSignalDefinition> Signals => mutableSignals.Value;
@@ -981,7 +980,7 @@ namespace Temporalio.Worker
         }
 
         private void ApplyUpdateRandomSeed(UpdateRandomSeed update) =>
-            Random = new(unchecked((int)update.RandomnessSeed));
+            Random = new(update.RandomnessSeed);
 
         private void OnQueryDefinitionAdded(string name, WorkflowQueryDefinition defn)
         {
--- a/src/Temporalio/Bridge/Cargo.lock
+++ b/src/Temporalio/Bridge/Cargo.lock
@@ -1785,6 +1785,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "rand_pcg"
+version = "0.3.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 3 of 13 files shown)


src/Temporalio/Bridge/Cargo.toml
```diff
@@ -15,6 +15,11 @@ log = "0.4"
 parking_lot = "0.12"
 prost = "0.11"
 prost-types = "0.11"
+# We rely on Cargo semver rules not updating a 0.x to 0.y. Per the rand
+# documentation, before 1.0, minor 0.x updates _can_ break portability which can
+# cause non-determinism.
+rand = "0.8.5"
+rand_pcg = "0.3.1"
```
Sushisource, line 22: This shouldn't be an issue since you're explicitly choosing the algorithm, I believe.

bergundy (approved):

src/Temporalio/Worker/WorkflowInstance.cs
```diff
@@ -157,8 +157,7 @@ public WorkflowInstance(WorkflowInstanceDetails details)
             pendingTaskStackTraces = workflowStackTrace == WorkflowStackTrace.None ? null : new();
             logger = details.LoggerFactory.CreateLogger($"Temporalio.Workflow:{start.WorkflowType}");
             replaySafeLogger = new(logger);
-            // We accept overflowing for seed (uint64 -> int32)
-            Random = new(unchecked((int)details.Start.RandomnessSeed));
+            Random = new(details.Start.RandomnessSeed);
```
Sushisource, line 160: Why does this actually matter? What was the issue with using the truncated 32 bit seed?

src/Temporalio/Bridge/Cargo.toml
cretz: You're probably right, but I was just scared knowing some of the higher level traits are helpers on the lower level and they call out multiple times that the library in general is only _portable_ for the same 0.x series. So I figured I'd call it out here.

src/Temporalio/Worker/WorkflowInstance.cs
cretz: A user on Slack, who has helped me with other things and seems quite knowledgeable, was concerned about a 32-bit seed to generate UUIDs. I admittedly haven't done homework here to know the tradeoff, but it seemed to make sense on the surface.

Sushisource: I'm like 99% sure this isn't going to be important. It'd matter if we were trying to guarantee some level of cryptographic safety, but we aren't (probably we should be really explicit about that in the docs if we're not).

So, ignoring that concern, the UUIDs are going to be plenty sufficiently random. You could generate them until the heat death of the universe for your workflow and probably never have a collision.

My 2c would be to just toss this, since it's just not important - but perhaps if that user can lay out exactly what the concern is then that might change things. If they are saying it's for cryptographic security, then all of this is a huge no-no anyways because our numbers are predictable on purpose, lol.

cretz: I think it's just that we're cutting the randoms in half by truncating the seed 32 bit. Sure if you generated V4 UUIDs using a regular crypto randomizer you'll never collide. But our random seed comes from run ID right? That of course should never collide. But what are the odds you turn two run IDs to different 64 bit seeds but when truncated to 32 bit may clash? I will ask user if they can provide more details. It's definitely understood these are not crypto-safe PRNGs.

Having said that, I think you're probably right it's barely a concern. But I figured "why not have a full 64 bit seed if I can", and I don't see a downside.

Sushisource: The odds are unbelievably tiny. 

The downside is simply LOC, code that can't even exist with WASM Core. .NET apps seem to have been getting along fine making UUIDs with 32 bits worth of seed randomness without a native extension.

cretz: > The downside is simply LOC, code that can't even exist with WASM Core

It just becomes WASM-ified also I think, but yes, using native constructs has value if we can.

> .NET apps seem to have been getting along fine making UUIDs with 32 bits worth of seed randomness without a native extension.

Not sure how true this is. .NET uses crypto rand by default and I just checked and Azure durable functions [here](https://github.com/Azure/azure-functions-durable-extension/blob/dev/src/WebJobs.Extensions.DurableTask) seems to use a SHA1 or MD5 hash of some concatenated values including a counter [here](https://github.com/Azure/azure-functions-durable-extension/blob/29884cdb52da332d31caac565789b6c260c2d895/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs#L1221).

I just saw one other concern that makes me want to do this. From https://learn.microsoft.com/en-us/dotnet/api/system.random?view=net-7.0#notes-to-callers:

> The implementation of the random number generator in the [Random](https://learn.microsoft.com/en-us/dotnet/api/system.random?view=net-7.0) class isn't guaranteed to remain the same across major versions of the .NET Framework. As a result, you shouldn't assume that the same seed will result in the same pseudo-random sequence in different versions of the .NET Framework.

That's scary that seeded Random is not portable. And kinda surprising. But I think it only applies to ".NET Framework" and not all of ".NET". I know Java promises portability and surely Python would consider seeded algorithm change a breaking change. But this alone makes me not want to use it.

Sushisource: Yeah, if that's not guaranteed then that's enough reason

cretz: :+1: Will merge. When we WASM-ify core, we should probably be prepared for lang-specific Rust code pre-WASM compile (i.e. make sure the WASM tooling is easy for each lang to run with their code instead of assuming the core WASM blob is the same for each language).

PR #94, opened 2023-06-28
Switch from Temporalite to Temporal CLI for local server and other changes related to server update

### LINKED ISSUE #50: [Feature Request] Move from Temporalite to Temporal CLI in local testing environment

Temporalite is the old way, CLI is the new. Also, we need to unskip `ExecuteWorkflowAsync_Patched_ProperlyHandled` once we do.

cretz:
## What was changed

* Download/start CLI-based dev server instead of Temporalite server when using `WorkflowEnvironment.StartLocalAsync`
* Remove `TemporaliteOptions` in favor of `DevServerOptions` (:boom: breaking change, but they were marked unstable knowing this was gonna happen)
* Update `ScheduleRange` to support defaults as set in most recent server version to match https://github.com/temporalio/temporal/pull/3969 (slight :boom: breaking change here as we split into multiple constructors, but only for people that used named parameters and relied on defaults)
* Added `AssemblyVersion` to the projects since that is what is available at runtime and therefore sent via gRPC header (and documented caveats about how it should be updated for prerelease versions). Until now, version was reported as `0.1.0.0`, but starting with `0.1.0-alpha6` this will be `0.1.0.6` and better.
* Changed tests to use `OperatorService.ListSearchAttributes` since `WorkflowService.GetSearchAttributes` is no longer valid
* Unskip patch test that was broken on Temporalite

## Checklist

1. Closes #50


tests/Temporalio.Tests/Client/TemporalClientScheduleTests.cs
```diff
@@ -37,10 +37,7 @@ public async Task CreateScheduleAsync_Basics_Succeeds()
                 {
                     new()
                     {
-                        // Intentionally set step 1 though 0 and 1 are the same to prove that step
-                        // comes back as sent not as defaulted (other 0 values for step don't come
-                        // back as 1)
-                        Second = new List<ScheduleRange> { new(1, Step: 1) },
```
Sushisource, line 43: Why'd this change?

Sushisource (approved):

tests/Temporalio.Tests/Client/TemporalClientScheduleTests.cs
cretz: Good question! https://github.com/temporalio/temporal/pull/3969 in the most recent server release explicitly put defaulted values (`End` defaults to `Start` and `Step` defaults to `1`) back on the ranges when sending back where it didn't before. I use expected input/output symmetry of ranges in my equality checks here. So this test broke because `End` and `Step` defaulted to `0` in current .NET code. I added constructor overloads to change the default of `End` to `Start` and the default of `Step` to `1` to match what is now being done in the server. So `Step` is already set to `1` here.

And note that my comment saying "prove that step comes back as sent not as defaulted" is no longer valid (I was explicitly testing our server behavior of input/output asymmetry here).

Issue #44, opened 2023-04-10
Trying to start a workflow from asp.net and the client hangs


jakejscott:
### What are you really trying to do?

Trying to start a workflow from asp.net minimal api.

### Describe the bug

If I create the client at startup, and then try and use that client instance within an endpoint it hangs when I call `StartWorkflowAsync`. 

If I create a client within the MapGet endpoint lambda it works. 

### Minimal Reproduction

```csharp
using Temporalio.Client;
using Temporalio.Runtime;

var builder = WebApplication.CreateBuilder(args);

builder.Logging.AddSimpleConsole(options =>
{
    options.TimestampFormat = "yyyy-mm-ddTHH:mm:ss.ff ";
    options.UseUtcTimestamp = true;
    options.IncludeScopes = false;
    options.SingleLine = true;
});

// If I create the client here, when I call `client.StartWorkflowAsync` the call blocks and doesn't work.
var client = await TemporalClient.ConnectAsync(new()
{
    TargetHost = "localhost:7233",
    Namespace = "default",
});

var app = builder.Build();

app.MapGet("/", async () =>
{
    app.Logger.LogInformation("We are here");
    
    // NOTE: If I connect here it works
    // var client = await TemporalClient.ConnectAsync(new()
    // {
    //     TargetHost = "localhost:7233",
    //     Namespace = "default",
    // });

    var id = $"simple-workflow-{Guid.NewGuid().ToString()}";

    // This call blocks and doesn't work
    var handle = await client.StartWorkflowAsync(
        workflow: SimpleWorkflow.Ref.RunAsync,
        arg: "Jake",
        options: new WorkflowOptions(
            id: id,
            taskQueue: "my-task-queue"
        )
    );

    var result = await handle.GetResultAsync();
    return result;
});

app.Logger.LogInformation("Started");

app.Run();
```

### Environment/Versions

- x86 Windows
- using temporal cli

cretz:
I have replicated. There is something about how the container is working that does is preventing the Tokio thread from running properly.

In the meantime, what you have there is not best practice. And it's not good to connect per call either. Consider changing to:

```csharp
builder.Services.AddSingleton(_ =>
    Task.Run(() => TemporalClient.ConnectAsync(new()
    {
        TargetHost = "localhost:7233",
        Namespace = "default",
    })).Result);
```

And:

```csharp
app.MapGet("/", async (TemporalClient client) =>
```

cretz:
[...]

PR #97, opened 2023-07-01
Feature/di client provider

mrtristan:
## What was changed
added start of `ITemporalClientProvider` concept as mentioned in https://github.com/temporalio/sdk-dotnet/issues/44#issuecomment-1501953442 by @cretz

## Why?
I have the need for a lazily instantiated client. I need to pull certs / connection information at runtime (from DI'd components). This seems like an innocuous enough change that was already being considered.

### DIFF
--- a/src/Temporalio.Extensions.Hosting/ITemporalClientProvider.cs
+++ b/src/Temporalio.Extensions.Hosting/ITemporalClientProvider.cs
@@ -1,0 +1,16 @@
+using System.Threading.Tasks;
+using Temporalio.Client;
+
+namespace Temporalio.Extensions.Hosting
+{
+    /// <summary>
+    /// Interface for providing a lazily instantiated <see cref="ITemporalClient" />.
+    /// </summary>
+    public interface ITemporalClientProvider
+    {
+        /// <summary>
+        /// Get the client.
+        /// </summary>
+        Task<ITemporalClient> GetClient();
+    }
+}
--- a/src/Temporalio.Extensions.Hosting/README.md
+++ b/src/Temporalio.Extensions.Hosting/README.md
@@ -57,8 +57,9 @@ When this extension is depended upon, two overloads for `AddHostedTemporalWorker
 
 One overload of `AddHostedTemporalWorker`, used in the quick start sample above, accepts the client target host, the
 client namespace, and the worker task queue. This form will connect to a client for the worker. The other overload of
-`AddHostedTemporalWorker` only accepts the worker task queue. In the latter, an `ITemporalClient` can either be set on
-the services container and therefore reused across workers, or the resulting builder can have client options set.
+`AddHostedTemporalWorker` only accepts the worker task queue. In the latter, an `ITemporalClient` or
+`ITemporalClientProvider` can be set on the services container and therefore reused across workers, or the resulting
+builder can have client options set.
 
 When called, these register a `TemporalWorkerServiceOptions` options class with the container and return a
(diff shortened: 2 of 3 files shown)


src/Temporalio.Extensions.Hosting/ITemporalClientProvider.cs
```diff
@@ -0,0 +1,16 @@
...
+{
+    /// <summary>
+    /// Interface for providing a lazily instantiated <see cref="ITemporalClient" />.
+    /// </summary>
+    public interface ITemporalClientProvider
+    {
+        /// <summary>
+        /// Get the client.
+        /// </summary>
+        Task<ITemporalClient> GetClient();
```
devbased, line 14: @mrtristan @cretz 
Is it possible to create a HostedService that will start a client and set it, for example, in `ITemporalClientProvider.Client`? I'm not sure about the practical necessity, just curious, and I don't like that `GetClient` returns `Task<T>` 😊, even though initialization happens only once. It seems to me that it would be better to use `ValueTask<T>`..

Sketch with HostedService:
```csharp
public interface ITemporalClientProvider
{
  public ITemporalClient Client { get; }
}

public TemporalClientProviderInitializer(
  TemporalClientProvider provider,
  ITemporalClientFactory factory,
  ... other dependencies)
{
  _provider = provider;
  _factory = factory;
  ...
}
public async Task StartAsync(...)
{
  var client = await _factory.Create();
  _provider.Client = client;
}

services.AddSingleton<TemporalClientProvider>();
services.AddSingleton<ITemporalClientProvider>(provider => provider.GetRequiredService<TemporalClientProvider>());
services.AddHostedService<TemporalClientProviderInitializer>();
services.AddHostedService<TemporalWorkerService>(); // must be after

services.AddTemporalClientFactory(...); // user defined ITemporalClientFactory
```
Or maybe using IApplicationBuilder would be more explicit when the client is created, but it has the downside of using Minimal API:
```csharp
await app.InitializeTemporalClient();

public static Task InitializeTemporalClient(this IApplicationBuilder app)
{
  var factory = app.Services.GetRequiredService<ITemporalClientFactory>();
  var provider = app.Services.GetRequiredService <TemporalClientProvider>();
  provider.Client = await factory.Create();
}
```

mrtristan: @devbased 
chad pointed out to me what i was missing regarding the capabilities here to have a DI'd yet lazily instantiated client which was my fundamental need. if you can live with a few clients being created (between the workers and your own usages) it seems fine to just get fancy in injecting/configuring the `TemporalClientConnectOptions` (or at least that covered my need).

https://temporalio.slack.com/archives/C012SHMPDDZ/p1688156494059699

cretz: @devbased and @mrtristan - I am going to open up a PR hopefully later today on my thoughts here. It will probably include the ability to configure `TemporalClientConnectOptions` + `TemporalClientProvider`, and a `TemporalClient.ConnectLazy` which is non-async (and DI extensions for that too). I do think `GetClientAsync()` will be required because I don't think a client deserves to be a hosted service (there's no explicit close or lifetime on it). It's more akin to a DB client I think.

EDIT: Well, lazy may be a bit of a problem, we'll see.

cretz: Opened https://github.com/temporalio/sdk-dotnet/pull/98, take a look

PR #103, opened 2023-07-13
Worker Versioning

### LINKED ISSUE #99: Worker versioning support

Sushisource:
Closes https://github.com/temporalio/sdk-dotnet/issues/99


src/Temporalio/Client/BuildIdOp.cs
```diff
@@ -0,0 +1,54 @@
+namespace Temporalio.Client
+{
+    /// <summary>
+    /// A supertype for all operations that may be performed by <see cref="ITemporalClient.UpdateWorkerBuildIdCompatibilityAsync"/>.
+    /// </summary>
+    public abstract record BuildIdOp
```
cretz, line 6: So I have been capitalizing `ID` and I shouldn't have been. I am going to fix this before we get to beta: https://github.com/temporalio/sdk-dotnet/issues/104.

src/Temporalio/Client/TemporalClient.Workflow.cs
```diff
@@ -336,6 +336,81 @@ public override async Task TerminateWorkflowAsync(TerminateWorkflowInput input)
                 }
             }
 
+            /// <inheritdoc />
+            public override async Task UpdateWorkerBuildIdCompatibilityAsync(UpdateWorkerBuildIdCompatibilityInput input)
```
cretz: Similar to how `TemporalClient.Schedules.cs` does it, I think it'd be cleaner if we put these in `internal partial class Impl` of a `TemporalClient.TaskQueue.cs`

src/Temporalio/Client/ITemporalClient.TaskQueue.cs
```diff
@@ -0,0 +1,58 @@
...
+        /// information could not be retrieved for will be marked with a `NotFetched` entry in
+        /// <see cref="BuildIdReachability.TaskQueueReachability"/>. The caller may issue another call
+        /// to get the reachability for those task queues.</param>
+        /// <param name="reachabilityType">The kind of reachability this request is concerned with.</param>
+        /// <param name="rpcOptions">RPC options.</param>
+        /// <returns>The reachability information.</returns>
+        Task<WorkerTaskReachability> GetWorkerTaskReachabilityAsync(
+            IReadOnlyCollection<string> buildIds,
+            IReadOnlyCollection<string> taskQueues,
+            TaskReachabilityType? reachabilityType,
```
cretz:
```suggestion
            TaskReachabilityType? reachabilityType = null,
```

Nit pick...and feel free to leave unchanged if we want people to have to explicitly put `null` here to set unspecified.

src/Temporalio/Client/WorkerBuildIdVersionSets.cs
```diff
@@ -0,0 +1,57 @@
...
+    public record WorkerBuildIdVersionSets(IReadOnlyCollection<BuildIdVersionSet> VersionSets)
+    {
+        /// <summary>
+        /// Get the default Build ID for this Task Queue.
+        /// </summary>
+        /// <returns>That Build ID.</returns>
+        public string DefaultBuildId()
+        {
+            return this.VersionSets.Last().Default();
+        }
```
cretz:
```suggestion
        public string DefaultBuildId => DefaultSet.Default;
```

Might as well make it a property. Also can you confirm that `VersionSets` will never be empty?

src/Temporalio/Client/WorkerBuildIdVersionSets.cs
```diff
@@ -0,0 +1,57 @@
...
+
+        /// <summary>
+        /// Get the default compatible set for this Task Queue.
+        /// </summary>
+        /// <returns>That set.</returns>
+        public BuildIdVersionSet DefaultSet()
+        {
+            return this.VersionSets.Last();
+        }
```
cretz:
```suggestion
        public BuildIdVersionSet DefaultSet => VersionSets.Last();
```

src/Temporalio/Client/WorkerBuildIdVersionSets.cs
```diff
@@ -0,0 +1,57 @@
...
+        /// <summary>
+        /// Convert from proto.
+        /// </summary>
+        /// <param name="proto">Proto.</param>
+        /// <returns>Converted value.</returns>
+        internal static WorkerBuildIdVersionSets FromProto(
+            Api.WorkflowService.V1.GetWorkerBuildIdCompatibilityResponse proto)
+        {
+            var sets = proto.MajorVersionSets.Select(vs => new BuildIdVersionSet(vs.BuildIds)).ToList();
+            return new WorkerBuildIdVersionSets(sets);
```
cretz:
```suggestion
            return new(sets);
```

A shortcut if you want it (at least I think it works here)

src/Temporalio/Client/WorkerBuildIdVersionSets.cs
```diff
@@ -0,0 +1,57 @@
...
+    /// </summary>
+    /// <param name="BuildIds">The Build IDs.</param>
+    public record BuildIdVersionSet(IReadOnlyCollection<string> BuildIds)
+    {
+        /// <summary>Get the default Build ID for this set.</summary>
+        /// <returns>That Build ID.</returns>
+        public string Default()
+        {
+            return this.BuildIds.Last();
+        }
```
cretz:
```suggestion
        public string Default => BuildIds.Last();
```

Can you confirm that `BuildIds` will never be empty?

src/Temporalio/Client/WorkerTaskReachability.cs
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    /// <summary>
+    /// Contains information about the reachability of a specific Build ID.
+    /// </summary>
+    /// <param name="TaskQueueReachability">Maps Task Queue names to the reachability status of the Build ID on that
+    /// queue. If the value is an empty list, the Build ID is not reachable on that queue.</param>
+    /// <param name="UnretrievedTaskQueues">If any Task Queues could not be retrieved because the server limits the
+    /// number that can be queried at once, they will be listed here.</param>
+    public record BuildIdReachability(
```
cretz: I know it's a bit redundant wrt file count, but I think we want each namespace-level type to have its own file

src/Temporalio/Client/WorkerTaskReachability.cs
```diff
@@ -0,0 +1,133 @@
...
+        /// <summary>
+        /// Convert from proto.
+        /// </summary>
+        /// <param name="proto">Proto.</param>
+        /// <returns>Converted value.</returns>
+        internal static WorkerTaskReachability FromProto(Api.WorkflowService.V1.GetWorkerTaskReachabilityResponse proto)
+        {
+            var idMap = proto.BuildIdReachability.ToDictionary(
+                bid => bid.BuildId,
+                bid =>
```
cretz: A `internal static BuildIdReachability FromProto(...` in `BuildIdReachability` may be a bit clearer and more consistent

src/Temporalio/Client/WorkerTaskReachability.cs
```diff
@@ -0,0 +1,133 @@
...
+    /// <param name="UnretrievedTaskQueues">If any Task Queues could not be retrieved because the server limits the
+    /// number that can be queried at once, they will be listed here.</param>
+    public record BuildIdReachability(
+        IReadOnlyDictionary<string, ReadOnlyCollection<TaskReachabilityType>> TaskQueueReachability,
+        IReadOnlyCollection<string> UnretrievedTaskQueues);
+
+    /// <summary>
+    /// A supertype for different ways a task might reach Workflows.
+    /// </summary>
+    public abstract record TaskReachabilityType
```
cretz: Can we make this a C# enum? Note in .NET unlike many of our other SDKs, I simple reuse the proto enums because they are actually quite clean in many of these cases (unlike other langauges where they are ugly). But I don't mind if we translate like this either though.

src/Temporalio/Worker/TemporalWorkerOptions.cs
```diff
@@ -71,6 +71,13 @@ public TemporalWorkerOptions()
         /// </summary>
         public string? BuildID { get; set; }
 
+        /// <summary>
+        /// Gets or sets whether this worker opts into the worker versioning feature. This ensures it only receives
+        /// workflow tasks for workflows which it claims to be compatible with. The <see cref="BuildID"/> field
+        /// is used as this worker's version when enabled, and must be set.
+        /// </summary>
+        public bool? UseWorkerVersioning { get; set; }
```
cretz:
```suggestion
        public bool UseWorkerVersioning { get; set; }
```

Unless there's value in `null` separate from `false`

src/Temporalio/Worker/WorkflowInstance.cs
```diff
@@ -709,6 +709,10 @@ private async Task RunTopLevelAsync(Func<Task> func)
                     {
                         cmd.Headers.Add(headers);
                     }
+                    if (e.Input.Options?.VersioningIntent is { } vi)
```
cretz, line 712: Nice, didn't know about this `{ }` pattern, I should apply it elsewhere here

src/Temporalio/Workflows/VersioningIntent.cs
```diff
@@ -0,0 +1,45 @@
...
+{
+    /// <summary>
+    /// Indicates whether the user intends certain commands to be run on a compatible worker Build
+    /// Id version or not.
+    ///
+    /// Where this type is accepted optionally, an unset value indicates that the SDK should choose the
+    /// most sensible default behavior for the type of command, accounting for whether the command will
+    /// be run on the same task queue as the current worker.
+    /// </summary>
+    public abstract record VersioningIntent
```
cretz, line 11: I think this should be a C# enum

cretz (commented):
Looks good, just a few comments, some subjective. Regardless, we can't merge this yet, we have a NuGet issue holding a pending release up.

cretz:
Looks like CI is failing with:
> Invalid options: `max_concurrent_wft_polls` cannot exceed `max_outstanding_workflow_tasks`

I had this happen in Python too when updating core because of some defaults I set on replayer and some new Rust validation. I think you need to just update the replayer options like I did at https://github.com/temporalio/sdk-python/pull/337/files#diff-46c3cc62d1ae5939208d155811f3fa2b25beb57a7d8f170c8d5f71747976bb49. Can run `dotnet test` to confirm entire suite runs successfully.

src/Temporalio/Client/WorkerTaskReachability.cs
Sushisource: I really like this b/c these can be exhaustively pattern matched unlike the enums

src/Temporalio/Workflows/VersioningIntent.cs
Sushisource: Same, pretty nice to be able to exhaustively match

src/Temporalio/Client/ITemporalClient.TaskQueue.cs
Sushisource: Yeah I didn't realize until halfway through this that you had to do that to get them to not be required. Missed fixing this one :+1:

src/Temporalio/Client/WorkerBuildIdVersionSets.cs
Sushisource: Won't be empty - but, good point I forgot to make the overall client method able to return null, which is what I'll do if the TQ has no versions.

src/Temporalio/Client/WorkerBuildIdVersionSets.cs
Sushisource: Yep, won't instantiate this with empty sets

src/Temporalio/Bridge/OptionsExtensions.cs
```diff
@@ -370,6 +374,9 @@ internal static class OptionsExtensions
                 max_task_queue_activities_per_second = options.MaxTaskQueueActivitiesPerSecond ?? 0,
                 graceful_shutdown_period_millis =
                     (ulong)options.GracefulShutdownTimeout.TotalMilliseconds,
+                use_worker_versioning = (byte)(options.UseWorkerVersioning ? 1 : 0),
+                // TODO: Expose to user
```
Sushisource, line 378: @cretz FYI I left this for your discretion.

It might make sense to have the bridge layer actually allow nulls/optionals here, and that way you can rely on Core's defined defaults instead of having to re-assume them here.

src/Temporalio/Client/WorkerTaskReachability.cs
cretz: Such is life of a C# dev using enums, but I think we should still use enums since we're in C# world. We also use C# enums to represent Temporal enums elsewhere in the SDK (and in general it makes sense as the language feature is there). Also this doesn't seem as much for the user benefit as it does for ours internally, and we can add a default clause that errors.

The reason C# doesn't consider all named enums exhaustive is due to internal compiler things I guess which allow it to technically be casted from any int. But I think we can enable named enum exhaustion check by enabling CS8509 and disabling CS8524. But I may be wrong.

Or for this one just reuse the API enum type. I have elsewhere, is not bad actually. (but you can't do this for the bridge-based `VersioningIntent`, we shouldn't expose bridge proto of course)

I am a bit surprised this sum type exhaustiveness check works. I guess it's some analyzer that just knows this class can't be implemented externally due to the private constructor.

src/Temporalio/Workflows/VersioningIntent.cs
cretz: Same as other comment, I think we should use the C# enum feature even if your internal switch statement loses this compile time check.

src/Temporalio/Bridge/OptionsExtensions.cs
cretz: Yeah, I think we should expose with a `int?` that defaults to `null` (meaning core default). But for now, we can just leave off and let core continue to use defaults and make a new GH issue unless you want to tackle all that now.

Sushisource: I think it's good for now

cretz:
All looks good. Just not marking approved yet so it doesn't get accidentally merged before the pending release is tagged. Will merge post release.

cretz (approved):

PR #111, opened 2023-07-18
Rename ID to Id 💥 BREAKING CHANGE

### LINKED ISSUE #104: Change ID to Id

We erroneously uppercased `ID` instead of proper capitalization of `Id`. Fix this before beta.

cretz:
## What was changed

Everywhere `ID` is used in an identifier, we are now using `Id` to follow proper .NET guidelines. This is a breaking change.

## Checklist

1. Closes #104

### DIFF
--- a/src/Temporalio.Extensions.OpenTelemetry/TracingInterceptor.cs
+++ b/src/Temporalio.Extensions.OpenTelemetry/TracingInterceptor.cs
@@ -112,14 +112,14 @@ namespace Temporalio.Extensions.OpenTelemetry
         /// <summary>
         /// Create tag collection for the given workflow ID.
         /// </summary>
-        /// <param name="workflowID">Workflow ID.</param>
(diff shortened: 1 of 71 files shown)


src/Temporalio/Client/TlsOptions.cs
```diff
@@ -34,6 +34,11 @@ public class TlsOptions : ICloneable
         /// </remarks>
         public byte[]? ClientPrivateKey { get; set; }
 
+        /// <summary>
+        /// Gets or sets BLAH BLAH.
```
Quinn-With-Two-Ns: Blah?

Quinn-With-Two-Ns (approved):

src/Temporalio/Client/TlsOptions.cs
cretz: Thanks! I was testing the acronym fixer and missed this.

PR #112, opened 2023-07-18
Versioning enum updates, rename "Cancelled" identifiers to "Canceled" (💥 BREAKING CHANGE), and README updates

### LINKED ISSUE #106: [Feature Request] Clarify inheritance rules/reasons for workflow attributes

It's not as clear as it could be in README that 1) we intentionally do not inherit an parent's `[Workflow]` attribute or `[WorkflowRun]` method, 2) why we don't, 3) that we do allow inheriting workflow signals/queries from base classes, but if overridden must have attributes, and 4) why for that too (i.e. their value for shared code reasons).

Maybe a "Workflow Inheritance" section in README is warranted.

### LINKED ISSUE #107: [Feature Request] Clarify in README the role of the native shared library

Unlike all-managed libraries that many are used to, this SDK requires a shared library. This is `temporal_sdk_bridge.dll` on Windows, `libtemporal_sdk_bridge.so` on Linux, `libtemporal_sdk_bridge.dylib` on macOS, etc and different ones are for different arch's. By default `dotnet build` will copy the artifact to the proper place based on the machine it is building on. But many people may be cross-compiling, so they need to set something like `--runtime linux-x64` when invoking `dotnet build`.

We need to make clear this is backed by an unmanaged native shared library, which [RIDs](https://learn.microsoft.com/en-us/dotnet/core/rid-catalog) we support, and that they need to be explicitly provided when building for a different platform than where the build runs. Can also just provide some general information about how it works and why we do it (the idea of a high-quality .NET library backed by a Rust core is interesting regardless).

### LINKED ISSUE #109: Use single-L for all cancelled (sic) type names like .NET does

For example, .NET has `OperationCanceledException` so ours should be `CanceledFailureException`. But need to leave double-L for cancellation (e.g. `CancellationToken`).

cretz:
## What was changed

* Changed to use existing API enum for `TaskReachability`
* Changed `VersioningIntent` to be an enum
* Changed identifiers (but not text necessarily) from `Cancelled` to `Canceled` - 💥 BREAKING CHANGE
* Added a section in the README clarifying workflow inheritance
* Added a section in the README clarifying some details about the native shared library

## Checklist

1. Closes #109
1. Closes #106
1. Closes #107


Sushisource (approved):
Oh, so it looks like the exhaustive matches still work, which is all I cared about. When I did this before I was forced to add a blanket `_ => xxx` case to the switches which is why I didn't want to use enums.

PR #113, opened 2023-07-19
Support for experimental workflow task events

cretz:
## What was changed

Added experimental `WorkflowTaskStarting` and `WorkflowTaskCompleted` events to `TemporalWorkerOptions` and `WorkflowReplayerOptions`. These are marked as experimental today and until https://github.com/temporalio/sdk-core/issues/579 is created, this is only working on activations so is knowingly inaccurate.

This has been requested by others wanting to do metering based on task. If this use case becomes prevalent enough we will make a features repo entry for it and will build it elsewhere and remove the experimental label.


src/Temporalio/Worker/TemporalWorkerOptions.cs
```diff
@@ -36,6 +36,44 @@ public TemporalWorkerOptions()
         /// <param name="taskQueue">Task queue for the worker.</param>
         public TemporalWorkerOptions(string taskQueue) => TaskQueue = taskQueue;
 
+        /// <summary>
+        /// Event for when a workflow task is starting. This should only be used for very advanced
```
Sushisource, line 40: Since it's experimental and we can change it, maybe call it `ActivationStarting`?

Sushisource (approved):

src/Temporalio/Worker/TemporalWorkerOptions.cs
cretz: The goal is task tracking even if the implementation isn't there yet. But I think the experimental API should reflect the goal instead of the current reality. I would have waited on this PR altogether but wanted to let users see the goal early.

PR #118, opened 2023-07-25
Remove activity definition cache

### LINKED ISSUE #117: [Bug] Activity definition cache reuses instance
### Describe the bug

When an activity is added for the same method on different instances, we reuse the cache for the first instance which also has the invoker cached which is bound to that one instance. We need to only cache the non-invocable details, not the invoker as well.

cretz:
## What was changed

Removed the activity definition cache. This cache was only useful in the past, before we used lambda expressions, to be able to repeatedly get the name and types at the call site. But now with lambda expressions we just get the name and assume that a compiled lambda means arg/result types are accurate. Also, the cache wasn't changed when we started storing custom invokers so the instance from the first activity definition was inadvertently reused.

## Checklist

1. Closes #117

### DIFF
--- a/tests/Temporalio.Tests/Worker/WorkflowWorkerTests.cs
+++ b/tests/Temporalio.Tests/Worker/WorkflowWorkerTests.cs
@@ -3012,6 +3012,62 @@ public class WorkflowWorkerTests : WorkflowEnvironmentTestBase
             workerOptions);
     }
 
+    public class DuplicateActivities
+    {
+        private readonly string returnValue;
+
+        public DuplicateActivities(string returnValue) => this.returnValue = returnValue;
+
+        [Activity]
+        public string DoThing() => returnValue;
+    }
+
+    [Workflow]
+    public class DuplicateActivityWorkflow
+    {
+        [WorkflowRun]
+        public async Task<string[]> RunAsync(string taskQueue1, string taskQueue2)
+        {
+            return new[]
+            {
+                await Workflow.ExecuteActivityAsync(
+                    (DuplicateActivities act) => act.DoThing(),
+                    new()
+                    {
+                        TaskQueue = taskQueue1,
+                        ScheduleToCloseTimeout = TimeSpan.FromHours(1),
+                    }),
+                await Workflow.ExecuteActivityAsync(
+                    (DuplicateActivities act) => act.DoThing(),
+                    new()
+                    {
+                        TaskQueue = taskQueue2,
+                        ScheduleToCloseTimeout = TimeSpan.FromHours(1),
+                    }),
+            };
+        }
+    }
+
+    [Fact]
+    public async Task ExecuteWorkflowAsync_DuplicateActivity_DoesNotCacheInstance()
+    {
+        await ExecuteWorkerAsync<DuplicateActivityWorkflow>(
+            async worker1 =>
+            {
+                await ExecuteWorkerAsync<DuplicateActivityWorkflow>(
+                    async worker2 =>
+                    {
+                        var ret = await Env.Client.ExecuteWorkflowAsync(
+                            (DuplicateActivityWorkflow wf) =>
+                                wf.RunAsync(worker1.Options.TaskQueue!, worker2.Options.TaskQueue!),
+                            new(id: $"workflow-{Guid.NewGuid()}", taskQueue: worker1.Options.TaskQueue!));
+                        Assert.Equal(new[] { "instance1", "instance2" }, ret);
+                    },
+                    new TemporalWorkerOptions().AddAllActivities(new DuplicateActivities("instance2")));
+            },
+            new TemporalWorkerOptions().AddAllActivities(new DuplicateActivities("instance1")));
+    }
+
     private async Task ExecuteWorkerAsync<TWf>(
         Func<TemporalWorker, Task> action,
         TemporalWorkerOptions? options = null,
--- a/src/Temporalio/Activities/ActivityDefinition.cs
+++ b/src/Temporalio/Activities/ActivityDefinition.cs
@@ -1,5 +1,4 @@
 using System;
-using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
@@ -13,8 +12,6 @@ namespace Temporalio.Activities
     /// </summary>
     public class ActivityDefinition
     {
-        private static readonly ConcurrentDictionary<MethodInfo, ActivityDefinition> CachedDefinitions = new();
-
         private readonly Func<object?[], object?> invoker;
 
         private ActivityDefinition(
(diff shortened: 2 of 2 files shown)


tests/Temporalio.Tests/Worker/WorkflowWorkerTests.cs
```diff
@@ -3012,6 +3012,62 @@ public async Task ExecuteWorkflowAsync_TaskEvents_AreRecordedProperly()
...
+                            (DuplicateActivityWorkflow wf) =>
+                                wf.RunAsync(worker1.Options.TaskQueue!, worker2.Options.TaskQueue!),
+                            new(id: $"workflow-{Guid.NewGuid()}", taskQueue: worker1.Options.TaskQueue!));
+                        Assert.Equal(new[] { "instance1", "instance2" }, ret);
+                    },
+                    new TemporalWorkerOptions().AddAllActivities(new DuplicateActivities("instance2")));
+            },
+            new TemporalWorkerOptions().AddAllActivities(new DuplicateActivities("instance1")));
+    }
+
```
dandavison, line 3070: Explaining this test to myself:

1. We define a workflow that returns a result containing the results of two activity calls.

2. The two activities are calls to the same activity method but on distinct instances (and using distinct task queues)

3. The utility `ExecuteWorkerAsync` allows us to start a worker and, while it's running, execute an arbitrary function that has access to the worker instance. We use it twice:

4. The first usage starts the first worker and uses the function to make the second nested call to `ExecuteWorkerAsync`

5. The second usage starts a second worker and uses the function to execute our workflow and check that the result shows that distinct activity methods were invoked.

tests/Temporalio.Tests/Worker/WorkflowWorkerTests.cs
```diff
@@ -3012,6 +3012,62 @@ public async Task ExecuteWorkflowAsync_TaskEvents_AreRecordedProperly()
...
+                await Workflow.ExecuteActivityAsync(
+                    (DuplicateActivities act) => act.DoThing(),
+                    new()
+                    {
+                        TaskQueue = taskQueue2,
+                        ScheduleToCloseTimeout = TimeSpan.FromHours(1),
+                    }),
+            };
+        }
+    }
```
dandavison, line 3049: Why does this test use two workers with different task queues?

tests/Temporalio.Tests/Worker/WorkflowWorkerTests.cs
```diff
@@ -3012,6 +3012,62 @@ public async Task ExecuteWorkflowAsync_TaskEvents_AreRecordedProperly()
...
+                    {
+                        var ret = await Env.Client.ExecuteWorkflowAsync(
+                            (DuplicateActivityWorkflow wf) =>
+                                wf.RunAsync(worker1.Options.TaskQueue!, worker2.Options.TaskQueue!),
+                            new(id: $"workflow-{Guid.NewGuid()}", taskQueue: worker1.Options.TaskQueue!));
+                        Assert.Equal(new[] { "instance1", "instance2" }, ret);
+                    },
+                    new TemporalWorkerOptions().AddAllActivities(new DuplicateActivities("instance2")));
+            },
+            new TemporalWorkerOptions().AddAllActivities(new DuplicateActivities("instance1")));
```
dandavison, line 3068: The way this test is written in a point-free style makes it a bit hard to read because the test assertion makes reference to `"instance1"` and `"instance2"` before the reader has encountered them.

dandavison (approved):
A couple of non-blocking questions/comments but LGTM! (I believe you confirmed in Slack that this test failed with the cache in place.)

tests/Temporalio.Tests/Worker/WorkflowWorkerTests.cs
cretz: It only needs the activities on different workers not really the workflow (but I just reuse the execute worker helper). We use two workers to confirm that the same activity definition but with different instances for different workers returns different values.

So basically this workflow calls the same activity on different task queues and the test confirms that the activity called was the different instances on different task queues.

tests/Temporalio.Tests/Worker/WorkflowWorkerTests.cs
cretz: That's the consequence of lambdas. Parameters to methods accepting lambdas are invoked and often used before the lambda is invoked. We make the parameters after the lambda because they are optional and many don't use them. This is test code so it is an acceptable tradeoff IMO.

tests/Temporalio.Tests/Worker/WorkflowWorkerTests.cs
dandavison: So we're using the lambda passed to `ExecuteWorkerAsync` to essentially play the role of customer application (non-Worker) code, since it starts a workflow. Therefore it's a bit artificial that the function has access to an actual instance of the worker (this confused me initially). But it looks like the pass-a-function-that-receives-a-worker-instance is a deeper part of the non-test-code SDK design.

cretz: Bingo. Before I removed the cache, this test would fail because the result of the workflow would be `["instance1","instance1"]` because no matter which activity instance you gave to a worker, it used the first cached one based on method which was bad. Now it returns `["instance1","instance2"]` as expected.

tests/Temporalio.Tests/Worker/WorkflowWorkerTests.cs
dandavison: It feels to me that the assertion being made is about what happens when we have distinct activity method instances and so it's not obvious that testing it requires multiple workers and task queues.

tests/Temporalio.Tests/Worker/WorkflowWorkerTests.cs
cretz: > So we're using the lambda passed to ExecuteWorkerAsync to essentially play the role of customer application (non-Worker) code, since it starts a workflow.

Yes

> Therefore it's a bit artificial that the function has access to an actual instance of the worker (this confused me initially).

Mostly. It's just so I can get it to create that random task queue which I can use

> But it looks like the pass-a-function-that-receives-a-worker-instance is a deeper part of the non-test-code SDK design.

Basically you have a worker that can run for some amount of time. In non-test code we support a worker running forever, until cancellation token is triggered, and/or until the completion of an async function. In test code, I made `ExecuteWorkerAsync` which is just a shortcut for the latter that also adds a workflow and makes a random task queue and does a couple of other things and gives the worker as the parameter just in case the test function wants to do something with it (most just use the task queue).

tests/Temporalio.Tests/Worker/WorkflowWorkerTests.cs
cretz: The test is specifically "what happens when a workflow calls the same activity method but on different task queues where the different activity instances may have different state". This requires multiple task queues (task queues are 1:1 with workers, so it requires multiple workers).

Issue #126, opened 2023-08-18
[Feature Request] Advanced metrics support


cretz:
### Describe the solution you'd like
With https://github.com/temporalio/sdk-core/pull/544 landed, we need to:

* [x] Support any missing prom options (done in #129)
* [x] Support any missing otel options (done in #129)
* ~Support user-accessible metric buffer-based forwarding~ (EDIT: Not needed with the below being present)
* [ ] Support .NET provided metric implementation (interesting...)
* [x] Support access to metrics from runtime (done in #129)
* [x] Support access to metrics from activity with default activity tags (done in #129)
* [x] Support access to metrics from workflow with default workflow tags and ignore metric updates during replay by default (done in #129)

cretz:
Most of this is now done. The thing remaining, and it's non-trivial, is "Support .NET provided metric implementation".

PR #129, opened 2023-08-23
Improved .NET metrics support

cretz:
## What was changed

* Exposed metric meter from core
* Added interfaces/impls for metric meter, counters, histograms, and gauges all with tag support
* Added `TemporalRuntime.MetricMeter` for users to add/record metrics
* Added `ActivityExecutionContext.MetricMeter` for users to add/record metrics from activities
* Added `Workflow.MetricMeter` for users to add/record metrics from workflows (replay safe)
* Added options to `MetricsOptions`, `PrometheusOptions`, and `OpenTelemetryOptions` to support all current metric options from core
* 💥 BREAKING CHANGE - Removed `Temporalio.Runtime.TelemetryOptions.Tracing` property and associated class
  * This is only for advanced internal support for core tracing which wasn't really used anyways

This is only part of the things needed for #126. This notably doesn't include a user-defined implementation for metrics handling which will come later.

Also, this PR will wait on https://github.com/temporalio/sdk-core/issues/598 which will then change a unit test here.


src/Temporalio/Bridge/Interop/Interop.cs
```diff
@@ -3,6 +3,27 @@
...
 {
+    internal enum MetricAttributeValueType
+    {
+        String = 1,
+        Int,
+        Float,
+        Bool,
+    }
+
+    internal enum MetricIntegerKind
```
Sushisource, line 14: I'd just call this `MetricKind`. If/when we accept non-integer values, the kinds will become generic over the value types.

src/Temporalio/Bridge/MetricAttributes.cs
```diff
@@ -0,0 +1,130 @@
...
+        /// Gets the pointer to the attributes.
+        /// </summary>
+        internal unsafe Interop.MetricAttributes* Ptr { get; private init; }
+
+        /// <summary>
+        /// Create a new instance with the given attributes appended.
+        /// </summary>
+        /// <param name="attributes">Attributes to append.</param>
+        /// <returns>New attributes with the given ones appended.</returns>
+        public MetricAttributes Append(IEnumerable<KeyValuePair<string, object>> attributes)
```
Sushisource, line 55: You'll probably want to add a method which uses the `merge` function exposed on attributes. This is the most efficient way (when using lang-side exporting) as it re-uses the IDs and doesn't need to allocate any new attributes - and it's fairly common to build up some attributes as a series of combined ones.

src/Temporalio/Common/MetricMeter.cs
```diff
@@ -0,0 +1,57 @@
...
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="MetricMeter" /> class.
+        /// </summary>
+        /// <param name="meter">Core meter.</param>
+        /// <param name="attributes">Core attributes.</param>
+        internal MetricMeter(Bridge.MetricMeter meter, Bridge.MetricAttributes attributes)
+        {
+            Meter = meter;
+            this.attributes = attributes;
```
Sushisource, line 19: This is sort of more like `defaultAttributes`?

src/Temporalio/Worker/WorkflowInstance.cs
```diff
@@ -237,6 +246,9 @@ public WorkflowInstance(WorkflowInstanceDetails details)
         /// <inheritdoc />
         public IReadOnlyDictionary<string, IRawValue> Memo => memo.Value;
 
+        /// <inheritdoc />
+        public IMetricMeter MetricMeter => metricMeter.Value;
```
Sushisource, line 250: Do we really want to expose a way to record metrics directly from the workflow? I can certainly see how it's useful but I definitely don't love adding things that do side effects in the middle of workflow code.

Sushisource (commented):
Making sense to me, just a few things.

src/Temporalio/Bridge/Interop/Interop.cs
cretz: This is only used for `MetricIntegerOptions` on a call to `metric_integer_new` which returns a `MetricInteger`. It'd make no sense to have to make `metric_integer_new`/`MetricIntegerOptions` support a non-metric-integer kind. If there's a `metric_double_new`/`MetricDoubleOptions`, it'd have `MetricDoubleKind` because it'd make no sense for such a method to support metric-integer kinds. The separation of metrics by their accepted type is due to the inability to expose generics in C.

src/Temporalio/Bridge/MetricAttributes.cs
cretz: This is internal code not user code. There is no exposed user code for an attribute class, so there is no value in a "merge" in internal code. It would never be called.

src/Temporalio/Common/MetricMeter.cs
cretz: Not really. You can call `WithTags` which makes a new meter with new tags. This is more like "base attributes", but it's pretty clear that "attributes" means the attributes on all metrics created from this meter just like "attributes" on the metric is for all values recorded on the metric.

src/Temporalio/Worker/WorkflowInstance.cs
cretz: Yes, but it is replay safe just like loggers and tracing. Our other SDKs have this and is a primary parity point we want to achieve with this project IMO.

bergundy (approved):

src/Temporalio/Bridge/MetricAttributes.cs
Sushisource: `WithTags` could work in terms of merging is I suppose what I really mean, whereas just recording and passing the tags there wouldn't. `WithTags` implying you're going to keep around that new meter for a while.

src/Temporalio/Bridge/Interop/Interop.cs
Sushisource: Huh? Why wouldn't `metric_integer_new` and `metric_double_new` be able to just accept `MetricKind`? All the kinds would be supported.

src/Temporalio/Worker/WorkflowInstance.cs
Sushisource: Makes sense

src/Temporalio/Bridge/Interop/Interop.cs
cretz: If that is the case I can refactor then. My concern was that every kind may not be supported by every data type. For right now there is only one, so we can just wait until another comes and then I can share if all are supported, sure.

src/Temporalio/Bridge/MetricAttributes.cs
cretz: Given the .NET API currently, what would merging look like from a user POV? We don't expose `MetricAttributes` to users, we only expose `WithTags` that only ever needs to append.

Sushisource: It wouldn't be any different - I mean `WithTags` can work _by_ calling merge.

cretz: But there is no real difference between "`clone` + create `MetricsAttributes` + `merge`" and just "`clone` + `add_new_attrs`" . It's all just clone + vec extend with how I use this. I don't allow metrics attributes to be mutable at this time because I don't give users the concept of a standalone set of pre-created attributes because it's usually not worth it (and the .NET API at https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.metrics works this way as well though they go one further and make you provide the full set every record call).

mjameswh (approved):
That all makes sense to me.

PR #136, opened 2023-09-11
Expose LastFailure and LastResult from WorkflowInfo

### LINKED ISSUE #133: [Feature Request] Need LastFailure and LastResult on WorkflowInfo

These are present on other SDKs. `LastResult` is mostly only for cron/schedule, but `LastFailure` can also have value for workflow retry.

cretz:
## What was changed

Expose `LastFailure` and `LastResult` from `WorkflowInfo`

## Checklist

1. Closes #133

### DIFF
--- a/src/Temporalio/Worker/WorkflowInstance.cs
+++ b/src/Temporalio/Worker/WorkflowInstance.cs
@@ -146,6 +146,9 @@ namespace Temporalio.Worker
                     RunId: start.ParentWorkflowInfo.RunId,
                     WorkflowId: start.ParentWorkflowInfo.WorkflowId);
             }
+            var lastFailure = start.ContinuedFailure == null ?
+                null : failureConverter.ToException(start.ContinuedFailure, PayloadConverter);
+            var lastResult = start.LastCompletionResult?.Payloads_.Select(v => new RawValue(v)).ToArray();
             static string? NonEmptyOrNull(string s) => string.IsNullOrEmpty(s) ? null : s;
             Info = new(
                 Attempt: start.Attempt,
@@ -153,6 +156,8 @@ namespace Temporalio.Worker
                 CronSchedule: NonEmptyOrNull(start.CronSchedule),
                 ExecutionTimeout: start.WorkflowExecutionTimeout?.ToTimeSpan(),
                 Headers: start.Headers,
+                LastFailure: lastFailure,
+                LastResult: lastResult,
                 Namespace: details.Namespace,
                 Parent: parent,
                 RetryPolicy: start.RetryPolicy == null ? null : Common.RetryPolicy.FromProto(start.RetryPolicy),
--- a/src/Temporalio/Workflows/WorkflowInfo.cs
+++ b/src/Temporalio/Workflows/WorkflowInfo.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using Temporalio.Common;
+using Temporalio.Converters;
 
 namespace Temporalio.Workflows
 {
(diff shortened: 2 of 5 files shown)


src/Temporalio/Workflows/WorkflowInfo.cs
```diff
@@ -12,6 +13,8 @@ namespace Temporalio.Workflows
     /// <param name="CronSchedule">Cron schedule if applicable.</param>
     /// <param name="ExecutionTimeout">Execution timeout for the workflow.</param>
     /// <param name="Headers">Headers from when the workflow was started.</param>
+    /// <param name="LastFailure">Failure if this workflow run is a continuation of a failure.</param>
+    /// <param name="LastResult">Successful result if this workflow is a continuation of a success.</param>
```
Sushisource, line 17: "workflow run was continued-as-new from a previous run which ended in a..."

Is a bit more clear I think

Sushisource (approved):

src/Temporalio/Worker/WorkflowInstance.cs
```diff
@@ -146,13 +146,18 @@ public WorkflowInstance(WorkflowInstanceDetails details)
                     RunId: start.ParentWorkflowInfo.RunId,
                     WorkflowId: start.ParentWorkflowInfo.WorkflowId);
             }
+            var lastFailure = start.ContinuedFailure == null ?
+                null : failureConverter.ToException(start.ContinuedFailure, PayloadConverter);
+            var lastResult = start.LastCompletionResult?.Payloads_.Select(v => new RawValue(v)).ToArray();
```
bergundy, line 151: Are codecs automatically applied to this? I think you used some reflection magic somewhere to ensure we apply codecs to anything over the lang-core bridge.

src/Temporalio/Workflows/WorkflowInfo.cs
```diff
@@ -32,6 +35,8 @@ public record WorkflowInfo(
         string? CronSchedule,
         TimeSpan? ExecutionTimeout,
         IReadOnlyDictionary<string, Api.Common.V1.Payload>? Headers,
+        Exception? LastFailure,
+        IReadOnlyCollection<IRawValue>? LastResult,
```
bergundy, line 39: nit: there's likely never be more than one result, none of our SDKs support returning more than one result.

bergundy: Can you also please test last result?

src/Temporalio/Workflows/WorkflowInfo.cs
cretz: This also available for successive cron and schedule workflow runs (technically it's exposed in the start API, we just don't expose in SDKs)

src/Temporalio/Worker/WorkflowInstance.cs
cretz: Yes, codecs are applied outside of this and there's a test to confirm no payloads slip by (even if they were just added to proto)

src/Temporalio/Workflows/WorkflowInfo.cs
cretz: A collection of one is ok. Another benefit is that this will be empty but not null when there's no result but there was a last run.

> Can you also please test last result?

Yes, I figured it may be hard to do in a timely manner with schedules, but maybe continue as new works. I will look to see how TS tested this value.

cretz: Added test

PR #137, opened 2023-10-03
Advanced .NET metrics support

### LINKED ISSUE #126: [Feature Request] Advanced metrics support

cretz:
## What was changed

* Added Rust and C FFI support for callback-based metric meters
* Added `Temporalio.Runtime.ICustomMetricMeter` with all of the metric types and added option on `Temporalio.Runtime.MetricOptions` to use it
* Reworked `Temporalio.Common.MetricMeter` by removing unnecessary `IMetricMeter` interface (and metric interfaces), disallowing custom implementations at this time, and accepting generic numeric types for metrics like the .NET meter
* Added new `Temporalio.Extensions.DiagnosticSource` extension which contains an `ICustomMetricMeter` implementation backed by `System.Diagnostics.Metrics.Meter` so core metrics now flow through as regular .NET metrics

## Checklist

1. Closes #126


src/Temporalio.Extensions.DiagnosticSource/CustomMetricMeter.cs
```diff
@@ -0,0 +1,188 @@
...
+            internal CustomMetricGauge(string name) => this.name = name;
+
+            ~CustomMetricGauge() => pendingLock.Dispose();
+
+            public void Set(T value, object tags)
+            {
+                // We need to support an extreme max here just to prevent unbounded memory growth.
+                // We do not want to require a logger for this, so at some extreme value we will
+                // just write to stderr and drop.
+                if (pending.Count > 50000)
```
Sushisource, line 101: If you have a readily available LRU thing maybe use that, but not a big deal.

src/Temporalio/Bridge/OptionsExtensions.cs
```diff
@@ -166,17 +167,30 @@ internal static class OptionsExtensions
...
+                // This object pins itself in memory and is only freed on the Rust side
+                customMeter = new CustomMetricMeter(options.CustomMetricMeter).Ptr;
+            }
             else
             {
                 throw new ArgumentException(
                     "Must have either Prometheus or OpenTelemetry metrics options");
             }
+            // WARNING: It is important that nothing after this point throws, because we have
+            // allocated a pointer for the custom meter which can only be freed on the Rust side
```
Sushisource, line 188: nit: Doesn't always happen
```suggestion
            // WARNING: It is important that nothing after this point throws, because we may have
            // allocated a pointer for the custom meter which can only be freed on the Rust side
```

src/Temporalio/Bridge/src/metric.rs
```diff
@@ -87,12 +90,19 @@ pub extern "C" fn metric_attributes_free(attrs: *mut MetricAttributes) {
...
     attrs: *const MetricAttribute,
     size: libc::size_t,
-) {
+) -> MetricAttributes {
     let attrs = unsafe { std::slice::from_raw_parts(attrs, size) };
-    orig.add_new_attrs(attrs.iter().map(metric_attribute_to_key_value));
+    let core = meter.core.inner.extend_attributes(
+        orig.clone(),
+        metrics::NewAttributes {
+            attributes: attrs.iter().map(metric_attribute_to_key_value).collect(),
```
Sushisource, line 102: Should probably be `into_iter` I think, to avoid an extra clone

src/Temporalio/Bridge/src/metric.rs
```diff
@@ -186,3 +196,295 @@ impl From<&MetricIntegerOptions> for metrics::MetricParameters {
...
+
+#[repr(C)]
+pub struct CustomMetricAttribute {
+    pub key: ByteArrayRef,
+    pub value: CustomMetricAttributeValue,
+    pub value_type: MetricAttributeValueType,
+}
+
+#[repr(C)]
+pub union CustomMetricAttributeValue {
```
Sushisource, line 245: You can, I believe, still represent this as an enum: https://doc.rust-lang.org/reference/type-layout.html#combining-primitive-representations-of-enums-with-fields-and-reprc

Not terribly important though, just wanted to point it out.

src/Temporalio/Bridge/src/metric.rs
```diff
@@ -186,3 +196,295 @@ impl From<&MetricIntegerOptions> for metrics::MetricParameters {
...
+            meter_impl: Arc::new(CustomMetricMeterImpl(meter)),
+        }
+    }
+
+    fn build_attributes(
+        &self,
+        append_from: Option<metrics::MetricAttributes>,
+        attribs: metrics::NewAttributes,
+    ) -> metrics::MetricAttributes {
+        unsafe {
```
Sushisource, line 306: Unsafe blocks can return values like other blocks - since we already use this pattern of blocks inside of not-unsafe-fns, might as well constrain the scope of the block to just the unsafe part. The meat of the function is still all safe.

src/Temporalio/Bridge/src/metric.rs
```diff
@@ -186,3 +196,295 @@ impl From<&MetricIntegerOptions> for metrics::MetricParameters {
...
+            let meter = &*(self.meter_impl.0);
+            let append_from = match append_from {
+                Some(metrics::MetricAttributes::Dynamic(v)) => {
+                    v.clone()
+                        .as_any()
+                        .downcast::<CustomMetricAttributes>()
+                        .unwrap()
```
Sushisource: Prefer expect to unwrap

src/Temporalio/Bridge/src/runtime.rs
```diff
@@ -137,6 +143,20 @@ pub extern "C" fn byte_array_free(runtime: *mut Runtime, bytes: *const ByteArray
...
 impl Runtime {
     fn new(options: &RuntimeOptions) -> anyhow::Result<Runtime> {
+        // Create custom meter here so it will be dropped on any error and
+        // therefore will call "free"
+        let custom_meter = unsafe {
+            options
+                .telemetry
+                .as_ref()
+                .map(|v| v.metrics.as_ref())
+                .flatten()
```
Sushisource: map then flatten can usually be written as `and_then` assuming these end up as options

src/Temporalio/Bridge/src/runtime.rs
```diff
@@ -204,45 +224,49 @@ impl TryFrom<&TelemetryOptions> for CoreTelemetryOptions {
...
-                .unit_suffix(prom_options.unit_suffix);
-            Ok(start_prometheus_metric_exporter(build.build()?)?.meter)
-        } else {
-            Err(anyhow::anyhow!(
-                "Either OpenTelemetry or Prometheus config must be provided"
-            ))
+fn create_meter(options: &MetricsOptions, custom_meter: Option<CustomMetricMeterRef>) -> anyhow::Result<Arc<dyn CoreMeter>> {
+    // OTel, Prom, or custom
+    if let Some(otel_options) = unsafe { options.opentelemetry.as_ref() } {
+        if !options.prometheus.is_null() || custom_meter.is_some() {
```
Sushisource, line 229: Personally I'd probably convert the C repr to a real enum before passing it in here so this validation logic can live by itself nice and separate from the real init stuff, but, up to you, a matter of taste.

src/Temporalio/Common/MetricCounter.cs
```diff
@@ -0,0 +1,37 @@
...
+        /// same tags, use <see cref="WithTags" /> for better performance.</param>
+        public abstract void Add(
+            T value, IEnumerable<KeyValuePair<string, object>>? extraTags = null);
+
+        /// <summary>
+        /// Create a new counter with the given tags.
+        /// </summary>
+        /// <param name="tags">Tags to append to existing tags.</param>
+        /// <returns>New counter.</returns>
+        public abstract MetricCounter<T> WithTags(IEnumerable<KeyValuePair<string, object>> tags);
```
Sushisource, line 35: Why not have WithTags be required by `Metric<T>`?

Sushisource (approved):
Looking good to me, nothing blocking. The whole generic instruments that you can maybe use the wrong type for isn't ideal, but, at the same time I don't know that making concrete versions is any better and misuse seems pretty unlikely, so I'm good with it.

src/Temporalio.Extensions.DiagnosticSource/CustomMetricMeter.cs
cretz: I don't unfortunately. Tracking and purging older entries would be more effort than this rare case is worth.

src/Temporalio/Bridge/src/metric.rs
cretz: Don't want to do a move from user-provided/owned slice and I accept a reference in `metric_attribute_to_key_value` and copy everything inside of there as needed. (not sure I can, `attrs` is a slice reference)

src/Temporalio/Bridge/src/metric.rs
cretz: I think that's only if every value in the enumerate would be the same size/primitive

src/Temporalio/Bridge/src/metric.rs
cretz: Yeah, I do that in most other places. I think I originally didn't here and other places in this PR is because most top-level statements have unsafe semantics and/or use pointers so I was lazy. But yes, I definitely look to limit unsafe usually.

src/Temporalio/Bridge/src/runtime.rs
cretz: Somewhere I have to have the mutual-exclusivity validation, whether in here or at enum creation time before here (and the added complexity of requiring that custom meter is _always_ created so it can be dropped even if it fails validation makes this a bit confusing)

src/Temporalio/Common/MetricCounter.cs
cretz: Because the return type is unique to each metric (covariant return types get ugly IMO)

src/Temporalio.Extensions.DiagnosticSource/CustomMetricMeter.cs
```diff
@@ -0,0 +1,188 @@
...
+                {
+                    pending[tagsObj] = measurement;
+                }
+                finally
+                {
+                    pendingLock.ExitReadLock();
+                }
+            }
+
+            public List<Measurement<T>> Snapshot()
```
mjameswh, line 122: Would be pertinent to add some comments here to explain how this Snapshot method is used. Had to search a bit to figure it out.

src/Temporalio.Extensions.DiagnosticSource/CustomMetricMeter.cs
```diff
@@ -0,0 +1,188 @@
...
+            ~CustomMetricGauge() => pendingLock.Dispose();
+
+            public void Set(T value, object tags)
+            {
+                // We need to support an extreme max here just to prevent unbounded memory growth.
+                // We do not want to require a logger for this, so at some extreme value we will
+                // just write to stderr and drop.
+                if (pending.Count > 50000)
+                {
+                    Console.Error.WriteLine($"Dropping gauge metric {name} since cardinality has grown over 50k");
```
mjameswh, line 103: I understand this is a very rare case, but it seems dangerous as it currently stand. If ever that `pending.Count > 50000` condition is reached (eg. user got creative with their tags, or just didn't understand that adding "workflow id" as a tag isn't a good idea), then console will get flooded with that error message, possibly bringing down the log target... Can't we just empty `pending`, and move on?

src/Temporalio.Extensions.DiagnosticSource/CustomMetricMeter.cs
```diff
@@ -0,0 +1,188 @@
...
+                    pendingLock.ExitReadLock();
+                }
+            }
+
+            public List<Measurement<T>> Snapshot()
+            {
+                pendingLock.EnterWriteLock();
+                try
+                {
+                    return pending.Values.ToList();
```
mjameswh, line 127: Not sure what underlying's expecting here, but shouldn't `pending` be emptied once snapshot has been called?

src/Temporalio.Extensions.DiagnosticSource/CustomMetricMeter.cs
cretz: This is private to this file and only used in this file. Having said that, I can add something to make it clear this is to resolve the measurement list callback for observable gauge

src/Temporalio.Extensions.DiagnosticSource/CustomMetricMeter.cs
cretz: Silently emptying without telling a user is arguably more dangerous. Not only are you discarding 50k pieces of valuable information but you aren't even telling them you did it. We can't reasonably weave a logger through here and we decided, same on the core side that will also spam on dropped metric when buffer is full in the buffer impl, that we can come back and debounce logs or open up more options if it's ever needed.

Having said that, if enough agree we should purge the dictionary once it gets so big, I can do that, sure.

src/Temporalio.Extensions.DiagnosticSource/CustomMetricMeter.cs
cretz: No, this isn't a "drain" but a "snapshot". Yes, that means that the dictionary entries never go away, so you can only have 50k tag combos for a single gauge (but the value can change) and gauge + tag combo is basically around for the life of the process. We have the same implementation on the Rust side. Basically this is asking for the gauge's current values and when it asks again, it will want the current values again.

cretz:
> The whole generic instruments that you can maybe use the wrong type for isn't ideal, but, at the same time I don't know that making concrete versions is any better and misuse seems pretty unlikely, so I'm good with it.

@Sushisource - It matches what .NET devs are used to: https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.metrics. And I am hedging just in case https://github.com/temporalio/sdk-core/issues/604 gets implemented.

src/Temporalio/Bridge/src/metric.rs
cretz: Fixed

src/Temporalio/Bridge/src/runtime.rs
cretz: Fixed

src/Temporalio.Extensions.DiagnosticSource/CustomMetricMeter.cs
cretz: Added comment

cretz:
I'm calling this good so I can move on to other things that also get the benefit of this sdk-core upgrade, but since it's both unreleased and beta, feel free to continue to give feedback and know we can adjust.

PR #142, opened 2023-10-18
Workflow update

### LINKED ISSUE #100: [dotnet] Workflow update support

cretz:
## What was changed

* Add `[WorkflowUpdate]` and `[WorkflowUpdateValidator]` attributes
* Support for workflow update in workflows including dynamic workflow update
* Support for workflow update in client including `WorkflowUpdateHandle` and `ExecuteWorkflowUpdateAsync` extension that is just `StartWorkflowUpdateAsync` + `GetResultAsync`

## Checklist

1. Closes #100


README.md
```diff
@@ -467,6 +468,17 @@ Attributes that can be applied:
   * `Dynamic = true` can be set for the query which makes the query a dynamic query meaning it will be called when
     no other queries match. The call must accept a `string` for the query name and `Temporalio.Converters.IRawValue[]`
     for the arguments. Only one dynamic query may be present on a workflow.
+* `[WorkflowUpdate]` attribute may be present on any public method that handles updates.
+  * Update methods must return a `Task` (can be a `Task<Result>`).
+  * The attribute can have a string argument for the update name. Otherwise the name is defaulted to the unqualified
+    method name with `Async` trimmed off the end if it is present.
+  * This attribute is not inherited and therefore must be explicitly set on any override.
+  * `Dynamic = true` can be set for the update which makes the update a dynamic update meaning it will be called when
```
dandavison:
```suggestion
  * `Dynamic = true` can be set for the update which makes the update a dynamic update, meaning it will be called when
```

src/Temporalio/Client/WorkflowHandleExtensions.cs
```diff
@@ -0,0 +1,149 @@
...
+        public static async Task ExecuteUpdateAsync<TWorkflow>(
+            this WorkflowHandle workflowHandle,
+            Expression<Func<TWorkflow, Task>> updateCall,
+            WorkflowUpdateOptions? options = null)
+        {
+            // TODO(cretz): Should we accept a timeout for GetResult? Or just reuse RPC timeout and
+            // cancellation token?
+            // TODO(cretz): Should we force-set wait for stage as completed? Still need to have
+            // discussion on what we should do on timeout of StartUpdateAsync while waiting for
+            // completion.
```
cretz: This is likely going to change based on outstanding questions to the team...

src/Temporalio/Client/WorkflowUpdateOptions.cs
```diff
@@ -0,0 +1,50 @@
...
+        /// Gets or sets the unique identifier for the update. This is optional and is defaulted to
+        /// a GUID if not set. This must be unique within the scope of a workflow execution (i.e.
+        /// namespace + workflow ID + run ID).
+        /// </summary>
+        public string? UpdateID { get; set; }
+
+        /// <summary>
+        /// Gets or sets the stage to wait for before returning from starting an update.
+        /// </summary>
+        public UpdateWorkflowExecutionLifecycleStage WaitForStage { get; set; } = UpdateWorkflowExecutionLifecycleStage.Completed;
```
cretz: Once we have a discussion with team about start/execute behavior, we may change this default to unspecified or accepted

src/Temporalio/Worker/WorkflowInstance.cs
```diff
@@ -841,6 +866,144 @@ private void Apply(WorkflowActivationJob job)
...
+                                    {
+                                        ProtocolInstanceId = update.ProtocolInstanceId,
+                                        Rejected = failureConverter.ToFailure(exc, PayloadConverter),
+                                    },
+                                });
+                            }
+                            else if (task.Exception is { } taskExc)
+                            {
+                                // Fails the task
+                                currentActivationException =
```
cretz, line 1003: Need to discuss how cancellation that propagates to a command _inside_ the update handler is returned. Right now it would be a task failure which seems wrong.

README.md
```diff
@@ -467,6 +468,17 @@ Attributes that can be applied:
   * `Dynamic = true` can be set for the query which makes the query a dynamic query meaning it will be called when
     no other queries match. The call must accept a `string` for the query name and `Temporalio.Converters.IRawValue[]`
     for the arguments. Only one dynamic query may be present on a workflow.
+* `[WorkflowUpdate]` attribute may be present on any public method that handles updates.
+  * Update methods must return a `Task` (can be a `Task<Result>`).
```
bergundy: nit: Is this clear enough that Result could be any serializable type and not some predefined thing? Maybe use `TResult`?

README.md
```diff
@@ -501,9 +513,9 @@ var handle = await client.StartWorkflowAsync(
 Console.WriteLine(
     "Current greeting: {0}",
     await handle.QueryWorkflowAsync(wf => wf.CurrentGreeting()));
-// Change the params via signal
-var signalArg = new GreetingParams(Salutation: "Aloha", Name: "John");
-await handle.SignalWorkflowAsync(wf => wf.UpdateGreetingParamsAsync(signalArg));
+// Change the params via update
+var updateArg = new GreetingParams(Salutation: "Aloha", Name: "John");
+await handle.ExecuteUpdateAsync(wf => wf.UpdateGreetingParamsAsync(updateArg));
```
bergundy, line 518: Worth showing that this can return a value.

src/Temporalio/Client/TemporalClient.Workflow.cs
```diff
@@ -220,6 +221,128 @@ public override async Task<TResult> QueryWorkflowAsync<TResult>(QueryWorkflowInp
...
+                        foreach (var kvp in req.Request.Input.Header.Fields)
+                        {
+                            req.Request.Input.Header.Fields[kvp.Key] =
+                                await codec.EncodeSingleAsync(kvp.Value).ConfigureAwait(false);
+                        }
+                    }
+                }
+                // Invoke
+                var resp = await Client.Connection.WorkflowService.UpdateWorkflowExecutionAsync(
+                    req, DefaultRetryOptions(input.Options?.Rpc)).ConfigureAwait(false);
```
bergundy, line 281: We discussed automatically retrying here on empty service response.
Unfortunately the server does not behave that way today so you can't really test this but eventually we should automatically start polling IMHO. Can punt on that for now.

I'm wondering if we additionally want to start polling on deadline exceeded as long as the cancellation token from rpc options is not expired as you did below (and optionally accept a timeout same as poll does) - again this would require async update to be available on the server so we could defer this until that stabilizes:

```
catch (RpcException e) when (
                            e.Code == RpcException.StatusCode.DeadlineExceeded && !tokenSource.IsCancellationRequested)
```

src/Temporalio/Client/TemporalClient.Workflow.cs
```diff
@@ -220,6 +221,128 @@ public override async Task<TResult> QueryWorkflowAsync<TResult>(QueryWorkflowInp
...
+                        },
+                        UpdateId = input.Id,
+                    },
+                    Identity = Client.Connection.Options.Identity,
+                    WaitPolicy = new() { LifecycleStage = UpdateWorkflowExecutionLifecycleStage.Completed },
+                };
+                // Create cancellation token based on timeout. This is a different timeout than RPC
+                // timeout because RPC timeout is _per call_ timeout, this is overall.
+                using (var tokenSource = CancellationTokenSource.CreateLinkedTokenSource(
+                    input.RpcOptions?.CancellationToken ?? default))
```
bergundy, line 312: Why does this input have `RpcOptions` and the one above has `Options.Rpc`? Is it worth trying to make it consistent?

src/Temporalio/Client/WorkflowHandleExtensions.cs
```diff
@@ -0,0 +1,149 @@
...
+        /// <see cref="WorkflowHandle.StartUpdateAsync{TWorkflow}(Expression{Func{TWorkflow, Task}}, WorkflowUpdateOptions?)" />
+        /// +
+        /// <see cref="WorkflowUpdateHandle.GetResultAsync(TimeSpan?, RpcOptions?)" />.
+        /// </summary>
+        /// <typeparam name="TWorkflow">Workflow class type.</typeparam>
+        /// <param name="workflowHandle">Workflow handle to use.</param>
+        /// <param name="updateCall">Invocation of workflow update method.</param>
+        /// <param name="options">Extra options.</param>
+        /// <returns>Completed update task.</returns>
+        public static async Task ExecuteUpdateAsync<TWorkflow>(
```
bergundy: Why is this an extension and not a part of workflow handle? Is it just to avoid the clutter? (for my education)

src/Temporalio/Client/WorkflowHandleExtensions.cs
```diff
@@ -0,0 +1,149 @@
...
+            this WorkflowHandle workflowHandle,
+            Expression<Func<TWorkflow, Task>> updateCall,
+            WorkflowUpdateOptions? options = null)
+        {
+            // TODO(cretz): Should we accept a timeout for GetResult? Or just reuse RPC timeout and
+            // cancellation token?
```
bergundy: If you accept a timeout on start, you could reuse the same timeout / token for both start and poll.
I'm starting to think that's the right way to go given that start and wait for accepted / completed could take longer than the RPC timeout.

src/Temporalio/Client/WorkflowHandleExtensions.cs
bergundy: Doesn't really make much sense to allow providing the wait stage here if progress not visible to the user. Implicitly set this to Completed.

src/Temporalio/Client/WorkflowUpdateOptions.cs
bergundy: IIUC we agree that for start it should be Accepted and IMHO it should be Completed for execute.

src/Temporalio/Worker/WorkflowInstance.cs
```diff
@@ -841,6 +866,144 @@ private void Apply(WorkflowActivationJob job)
 
         private void ApplyCancelWorkflow() => cancellationTokenSource.Cancel();
 
+        private void ApplyDoUpdate(DoUpdate update)
```
bergundy, line 869: I wasn't involved with the core PR but this should have been named UpdateWorkflow for consistency 🤷

bergundy (commented):
I didn't finish the entire review, I'll try to continue this tomorrow.

src/Temporalio/Client/WorkflowHandleExtensions.cs
dandavison: In Python and Typescript this call is just called `update()`, but here we're proposing to call it `ExecuteUpdate()`. Would you say that ideally it should be called `executeUpdate()` / `execute_update()` in Typescript and Python? **Edit**: Chad confirmed elsewhere that he does propose that.

I understand that you're choosing this name to make it similar to other stateful entities such as Workflow and Activity executions. The other consideration though is that "update" is a verb, whereas "workflow" and "activity" are not. I kind of buy your preference, but I would like us to discuss across-language consistency.

README.md
cretz: I think it's fine. This is basic README that will be supplanted by official docs anyways.

README.md
cretz: I don't think that's necessary in this README, I think that can happen when this is properly documented. Note I am not showing you can set an ID either.

src/Temporalio/Client/TemporalClient.Workflow.cs
cretz: > but eventually we should automatically start polling IMHO

Disagree. We decided as a team that start will default to `Accepted`. Therefore this call should error (or return unset handle?) if your stage is not reached in a single API call. Granted this error can contain the handle in it for the _user_ to start polling (called via get result). The start call should never poll I don't think, that's what the get-result call does. If you choose to set `WaitForStage` as `Completed` you are saying "error if it doesn't complete within the single call". Otherwise you can use an alternative.

When polling is stable server side we should discuss what we want behavior here to be.

src/Temporalio/Client/TemporalClient.Workflow.cs
cretz: Because some calls have call-specific options which _include_ RPC options within them and some calls just accept RPC options directly because they don't need their own options object (and/or they can take parameters explicitly). The get-result call of workflows and updates which both poll do not need a special options objects so they accepts `RpcOptions` directly.

src/Temporalio/Client/WorkflowHandleExtensions.cs
cretz: A bit that, but more to match what we do for workflows which is an extension on the client. Granted that client is an interface so extensions make more sense there. I have no good reason here, so yeah, I am going to move these to the handle directly.

src/Temporalio/Client/WorkflowHandleExtensions.cs
cretz: We do accept an RPC timeout on start, but that is not the same as the timeout for start + get result. We as a team have decided not to support polling on execute currently so this won't apply until then so we won't add the timeout parameter until then.

src/Temporalio/Client/WorkflowHandleExtensions.cs
cretz: Yes, forcibly overriding next commit.

src/Temporalio/Client/WorkflowUpdateOptions.cs
cretz: Right, this is based on the team discussion that happened after I made this comment. Next commit will update this.

src/Temporalio/Worker/WorkflowInstance.cs
cretz: Maybe, but doesn't apply to this PR of course.

src/Temporalio/Client/WorkflowHandleExtensions.cs
cretz: Discussed internally and made appropriate changes

src/Temporalio/Client/WorkflowUpdateOptions.cs
cretz: Updated.

src/Temporalio/Client/WorkflowHandleExtensions.cs
cretz: Updated

src/Temporalio/Client/WorkflowHandleExtensions.cs
cretz: Done

cretz:
Updates made based on team discussions

src/Temporalio/Client/TemporalClient.Workflow.cs
bergundy: > We decided as a team that start will default to Accepted.

Yes, considering that `Admitted` isn't yet available, but I can see arguments both ways.

> Therefore this call should error (or return unset handle?) if your stage is not reached in a single API call.

I would agree if the API didn't have a wait condition, it's not just a start call, it's start and wait.

I think we shouldn't expose `StartUpdate` with `WaitForStage=Completed`, you might as well use `ExecuteUpdate` and give us a timeout and we'll poll if necessary and the timeout permits.

For start with `Accepted`, I think we should poll if the user is willing to wait, why would the user care how many RPCs we make internally, what they care about is that it happens within a given timeout.
When using `Admitted`, there's obviously no need to poll.

> When polling is stable server side we should discuss what we want behavior here to be.

We can defer this conversation for when polling is stable though.

src/Temporalio/Client/WorkflowHandleExtensions.cs
bergundy: Makes sense for now, agree that considering polling isn't stable we can table this discussion.

src/Temporalio/Client/TemporalClient.Workflow.cs
dandavison: > I think we shouldn't expose StartUpdate with WaitForStage=Completed, you might as well use ExecuteUpdate and give us a timeout and we'll poll if necessary and the timeout permits.

You don't think there are any valid use cases for wanting to get a handle back in that situation rather than a result? I know they have the option of specifying an updateID up front and using it to obtain a handle, but I'm not entirely convinced that it's unreasonable to allow the handle to be obtained straightforwardly.

I would say that we should go for uniformity/consistency and allow it, since it makes sense (unlike for example passing `waitFor=Accepted` into `ExecuteUpdate`).

> why would the user care how many RPCs we make internally

It's not implausible that high-volume users would want control over egress costs.

cretz: Ok, there is a case for no enum today until admitted is implemented, and no completed ever (that's the difference between start and execute). Discussing with team.

src/Temporalio/Worker/WorkflowInstance.cs
cretz: Cancellation is an update failure, fixed

README.md
cretz: Updated

src/Temporalio/Client/TemporalClient.Workflow.cs
cretz: Removed publicly visible enum (except for interceptor which needs it still). Will defer polling until it's stable.

README.md
cretz: Fixed

README.md
bergundy: It's worth it IMHO, otherwise it's unclear how this is different from a signal. Of course it'll be properly documented later but for a Temporal noob this would be a bit confusing.

bergundy (approved):

README.md
Sushisource: Agreed. No reason _not_ to show it, and at least this reason _to_ show it.

src/Temporalio/Client/WorkflowHandle.cs
```diff
@@ -302,6 +302,189 @@ public record WorkflowHandle(
...
+                Id: Id,
+                RunId: RunId,
+                FirstExecutionRunId: FirstExecutionRunId,
+                Update: update,
+                Args: args,
+                Options: options,
+                Headers: null));
+
+        /// <summary>
+        /// Shortcut for
```
Sushisource: Worth an extra sentence just to make it obvious
```suggestion
        /// Start a workflow update and wait for it to complete
        /// Shortcut for
```

src/Temporalio/Client/WorkflowUpdateHandle.cs
```diff
@@ -0,0 +1,92 @@
...
+        /// <summary>
+        /// Wait for an update result disregarding any return value.
+        /// </summary>
+        /// <param name="timeout">Optional timeout. Defaults to 1 minute, but can be set to
+        /// <see cref="System.Threading.Timeout.InfiniteTimeSpan" /> but probably shouldn't. A
+        /// timeout will throw a <see cref="OperationCanceledException" />.</param>
+        /// <param name="rpcOptions">Extra RPC options.</param>
+        /// <returns>Completed update task.</returns>
+        /// <remarks>WARNING: Workflow update is experimental and APIs may change.</remarks>
+        public Task GetResultAsync(TimeSpan? timeout = null, RpcOptions? rpcOptions = null) =>
```
Sushisource, line 35: I'm a bit inclined to say this should actually have a different name if it disregards the result. Like `WaitForCompletedAsync`.

But, ignoring that - why even bother with this separately. Can't someone just use the result-returning version and just not use the result?

src/Temporalio/Worker/Interceptors/WorkflowInboundInterceptor.cs
```diff
@@ -65,5 +65,20 @@ private protected WorkflowInboundInterceptor()
         /// <param name="input">Input details of the call.</param>
         /// <returns>Query result.</returns>
         public virtual object? HandleQuery(HandleQueryInput input) => Next.HandleQuery(input);
+
+        /// <summary>
+        /// Intercept update validation. This is invoked every time validation is needed regardless
+        /// of whether a validator was defined.
+        /// </summary>
+        /// <param name="input">Input details of the call.</param>
+        public virtual void ValidateUpdate(HandleUpdateInput input) => Next.ValidateUpdate(input);
```
Sushisource, line 74: In my implementation I only invoke if the validator is defined. I can see doing it always, if you want some blanket way to reject based on... whatever... without defining a validator for every update. I'll change that in my impl.

src/Temporalio/Worker/WorkflowInstance.cs
```diff
@@ -841,6 +866,145 @@ private void Apply(WorkflowActivationJob job)
...
+                            dynamicArgPrepend: update.Name),
+                        Headers: update.Headers);
+                    // If validation is requested, perform it inline. We call the interceptor
+                    // regardless of whether a validator is defined. An interceptor can validate
+                    // even if there is no validator method.
+                    if (update.RunValidator)
+                    {
+                        // Capture command count so we can ensure it is not altered during validation
+                        var origCmdCount = completion?.Successful?.Commands?.Count ?? 0;
+                        inbound.Value.ValidateUpdate(updateInput);
```
Sushisource: Seems this would be useful to put behind some kind of "read only mode" protector like exists in Python / Java?

README.md
cretz: I think it's reading too much into tiny sample (aren't they gonna get confused w/ update vs query? what about showing that queries can accept parameters? what about showing validators?), but I will add it.

Sushisource (approved):
Looking good! Just some minor stuff.

src/Temporalio/Client/WorkflowHandle.cs
cretz: :+1: I was mimicking my terminology for `ExecuteWorkflow` (e.g. [here](https://dotnet.temporal.io/api/Temporalio.Client.ITemporalClientExtensions.html#Temporalio_Client_ITemporalClientExtensions_ExecuteWorkflowAsync_Temporalio_Client_ITemporalClient_System_String_System_Collections_Generic_IReadOnlyCollection_System_Object__Temporalio_Client_WorkflowOptions_)) but I can add a sentence.

src/Temporalio/Client/WorkflowUpdateHandle.cs
cretz: > I'm a bit inclined to say this should actually have a different name if it disregards the result. Like WaitForCompletedAsync.

This matches the workflow get result (and we tend to kinda call it that for things that can optionally return results like workflows and activities). Users may get confused that the presence of a return value changes the function name since other than return value they are the same.

> But, ignoring that - why even bother with this separately. Can't someone just use the result-returning version and just not use the result?

In .NET `Task` and `Task<Result>` are distinctly defined and there is no "unit" or "void" type for these purposes in .NET. So every .NET call (e.g. [Task.Run](https://learn.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.run)) has to differentiate.

src/Temporalio/Worker/Interceptors/WorkflowInboundInterceptor.cs
cretz: :+1: Yeah I went back and forth here too but I figured "why not"

src/Temporalio/Worker/WorkflowInstance.cs
cretz: Yeah, this was only the second time I needed such a thing and the error message is different enough and impl simpl;e enough that I didn't make reuseable (if .NET had simple yield-based contextmanager type thing like Python, definitely).

src/Temporalio/Worker/Interceptors/WorkflowInboundInterceptor.cs
cretz: Ok, after team discussion, here is what we want the SDK logic to be due to possible mutation of input and performance:

* if core determines validation should occur
  * If validator present
    * decode args for validator (or reject on fail)
    * call validator (interceptor) w/ those args
  * decode args for update (or reject on fail)
  * accept
* else
  * decode args for update (or task-fail on fail)
  * accept
* call updator (interceptor) w/ those args

README.md
cretz: Hrmm, I'm struggling to fit the update-with-return value use case into the above sample. Any ideas on what I can change the `GreetingWorkflow`'s `UpdateGreetingParamsAsync` to for demonstrating a return value?

(@Sushisource - you'll find the same challenge when you update the Python README)

src/Temporalio/Client/WorkflowHandle.cs
cretz: Extra sentence added

src/Temporalio/Worker/Interceptors/WorkflowInboundInterceptor.cs
cretz: The logic has been updated to support this approach

Quinn-With-Two-Ns: Sorry to beat a dead horse, but I thought if we have to decode the args we would go through the interceptor

Quinn-With-Two-Ns: because it is a non trivial change in go and java to even know if a validator is present or not without going through the interceptor

src/Temporalio/Client/WorkflowUpdateHandle.cs
Sushisource: Aaah, ok, makes sense.

src/Temporalio/Worker/WorkflowInstance.cs
```diff
@@ -841,6 +866,177 @@ private void Apply(WorkflowActivationJob job)
...
+                object?[]? argsForUpdate = null;
+                // We define this up here because there are multiple places it's called below
+                object?[] DecodeUpdateArgs() => DecodeArgs(
+                    method: updateDefn.Method ?? updateDefn.Delegate!.Method,
+                    payloads: update.Input,
+                    itemName: $"Update {update.Name}",
+                    dynamic: updateDefn.Dynamic,
+                    dynamicArgPrepend: update.Name);
+
+                // Do validation. Whether or not this runs a validator, this should accept/reject.
```
dandavison, line 907:
```suggestion
                // Do validation. Whether or not this runs a validator, this must accept/reject.
```

src/Temporalio/Worker/WorkflowInstance.cs
```diff
@@ -841,6 +866,177 @@ private void Apply(WorkflowActivationJob job)
...
+                            {
+                                currentActivationException = new InvalidOperationException(
+                                    $"Update validator for {update.Name} created workflow commands");
+                                return Task.CompletedTask;
+                            }
+                        }
+
+                        // We want to try to decode args here _inside_ the validator rejection
+                        // try/catch so we can prevent acceptance on invalid args
+                        argsForUpdate = DecodeUpdateArgs();
```
dandavison, line 941: If you move this down a couple of lines to outside the `if (runValidator)`, then it will be inside the validator rejection try/catch, but it will always run. Then you can eliminate the other call below in the handler section.

src/Temporalio/Worker/WorkflowInstance.cs
```diff
@@ -841,6 +866,177 @@ private void Apply(WorkflowActivationJob job)
...
+                }
+
+                // Issue actual update. We are using ContinueWith instead of await here so that user
+                // code can run immediately. But the user code _or_ the task could fail so we need
+                // to reject in both cases.
+                try
+                {
+                    // If the args were not already decoded because we didn't run the validation
+                    // step, decode them here
+                    argsForUpdate ??= DecodeUpdateArgs();
```
dandavison, line 975: As mentioned above, you can eliminate this line by always decoding inside the validation `try` above. (Just a clean up -- I know the way you have it it won't be called more than necessary.)

src/Temporalio/Worker/WorkflowInstance.cs
```diff
@@ -841,6 +866,177 @@ private void Apply(WorkflowActivationJob job)
...
+                    AddCommand(new()
+                    {
+                        UpdateResponse = new()
+                        {
+                            ProtocolInstanceId = update.ProtocolInstanceId,
+                            Rejected = failureConverter.ToFailure(e, PayloadConverter),
+                        },
+                    });
+                    return Task.CompletedTask;
+                }
```
dandavison, line 1036: To check my understanding, exceptions other than Temporal `FailureException` here bubble up and fail the workflow task, correct?

src/Temporalio/Worker/WorkflowInstance.cs
```diff
@@ -841,6 +866,177 @@ private void Apply(WorkflowActivationJob job)
...
+                    method: updateDefn.Method ?? updateDefn.Delegate!.Method,
+                    payloads: update.Input,
+                    itemName: $"Update {update.Name}",
+                    dynamic: updateDefn.Dynamic,
+                    dynamicArgPrepend: update.Name);
+
+                // Do validation. Whether or not this runs a validator, this should accept/reject.
+                try
+                {
+                    if (update.RunValidator)
```
dandavison, line 910: Go and Java do not run validation when replaying. Do we need to put the same check in place here?

src/Temporalio/Worker/Interceptors/WorkflowInboundInterceptor.cs
cretz: > Sorry to beat a dead horse
No problem, we need to get this right/consistent. Happy to discuss as much as is necessary.

So double-decoding args has already been settled to prevent validator mutability. We will discuss as a team whether to prevent interceptor call if validator not present.

src/Temporalio/Worker/WorkflowInstance.cs
cretz: I think you're overthinking the terms used in simple internal English comments as if they're an RFC heh

src/Temporalio/Worker/WorkflowInstance.cs
cretz: Core is doing that check for you and setting this boolean

src/Temporalio/Worker/WorkflowInstance.cs
cretz: But I don't want to catch argument decoding failure validator-style w/ rejection if run validator is false. If run validator is false, I don't want to decode until later, where the failure is a task failure not a rejection.

src/Temporalio/Worker/WorkflowInstance.cs
cretz: I don't want to decode in the try above, I want this non-failure-exception to throw out of this method and therefore fail the task.

src/Temporalio/Worker/WorkflowInstance.cs
dandavison: Ah-ha. I had not understood that, thanks.

src/Temporalio/Worker/WorkflowInstance.cs
cretz: Correct. So we only reject/fail update if there is an exception when run-validator is true or there is a failure exception. Otherwise we fail the task.

src/Temporalio/Worker/Interceptors/WorkflowInboundInterceptor.cs
cretz: Ok, after discussion, the logic in the bullets above is what we want. Don't run validate interceptor (and therefore don't decode args for it) if a validator is not present

PR #144, opened 2023-10-19
Add missing gRPC calls

### LINKED ISSUE #143: [Bug] DeleteWorkflowExecution service call not exposed properly
### Describe the bug

On the Rust side, we don't expose the `DeleteWorkflowExecution` call. Not only do we need to, but we need a unit test to make sure no calls are ever missed going forward. This should basically be us standing up a fake gRPC server, calling every call on workflow and operator service reflectively, and confirming every call reaches the server.

cretz:
## What was changed

Added missing gRPC calls from the core string list and added test to ensure they are all there

## Checklist

1. Closes #143

### DIFF
--- a/tests/Temporalio.Tests/Client/TemporalClientTests.cs
+++ b/tests/Temporalio.Tests/Client/TemporalClientTests.cs
@@ -1,7 +1,11 @@
 namespace Temporalio.Tests.Client;
 
+using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Reflection;
 using Temporalio.Client;
+using Temporalio.Runtime;
 using Xunit;
 using Xunit.Abstractions;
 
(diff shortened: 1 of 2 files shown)


tests/Temporalio.Tests/Client/TemporalClientTests.cs
```diff
@@ -48,4 +52,124 @@ public async Task CreateLazy_Connection_NotConnectedUntilCallMade()
...
+        actualCalls.Clear();
+
+        // Make calls and populate expected calls
+        var expectedCalls = new List<string>();
+        foreach (var method in typeof(T).GetMethods(
+            BindingFlags.Instance | BindingFlags.Public | BindingFlags.DeclaredOnly))
+        {
+            // Add expected call sans Async suffix
+            var call = method.Name.Substring(0, method.Name.Length - 5);
+            Console.WriteLine("!!! CALL: " + call);
```
bergundy: Did you mean to leave this console statement here?

tests/Temporalio.Tests/Client/TemporalClientTests.cs
```diff
@@ -48,4 +52,124 @@ public async Task CreateLazy_Connection_NotConnectedUntilCallMade()
...
+        {
+            Runtime = new(new()
+            {
+                Telemetry = new() { Metrics = new() { CustomMetricMeter = captureMeter } },
+            }),
+        });
+
+        // Check workflow service and operator service
+        await AssertAllRpcsAsync(captureMeter.Calls, env.Client.Connection.WorkflowService);
+        await AssertAllRpcsAsync(captureMeter.Calls, env.Client.Connection.OperatorService, skip: "AddOrUpdateRemoteCluster");
```
bergundy, line 74: Curious why this call is skipped and if this is temporary.

bergundy (approved):

tests/Temporalio.Tests/Client/TemporalClientTests.cs
cretz: I did not, thanks!

tests/Temporalio.Tests/Client/TemporalClientTests.cs
cretz: It hangs/messes up the server when you call it with an empty request. Not temporary, but hopefully it's the only one like it.

tests/Temporalio.Tests/Client/TemporalClientTests.cs
cretz: Fixed

PR #150, opened 2023-10-30
Update release version to beta 2

cretz:
## What was changed

Prepare for beta 2 release


Directory.Build.props
```diff
@@ -18,7 +18,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <Version>0.1.0-beta1</Version>
+    <Version>0.1.0-beta2</Version>
```
Quinn-With-Two-Ns, line 21: Is this preferred over `0.2.0-beta`?

Quinn-With-Two-Ns (approved):

Directory.Build.props
mjameswh: Definitely not `0.2.0-beta`, since there's never been a 0.1.0 release (ie. without a prerelease suffix). Now, given that we are striving for GA, maybe that should be 1.0.0-beta instead.

mjameswh: By the way, here are the pertinent paragraphs from the [semver](https://semver.org/) spec. Not that we adhere fanatically to that document, but it still appears to be an interesting frame of reference. So compliant values would be `0.1.0-beta.2` or `1.0.0-beta`.

>  A pre-release version MAY be denoted by appending a hyphen and a series of dot separated identifiers immediately following the patch version. Identifiers MUST comprise only ASCII alphanumerics and hyphens [0-9A-Za-z-]. Identifiers MUST NOT be empty. Numeric identifiers MUST NOT include leading zeroes. Pre-release versions have a lower precedence than the associated normal version. A pre-release version indicates that the version is unstable and might not satisfy the intended compatibility requirements as denoted by its associated normal version. Examples: 1.0.0-alpha, 1.0.0-alpha.1, 1.0.0-0.3.7, 1.0.0-x.7.z.92, 1.0.0-x-y-z.--.
>
> Precedence MUST be calculated by separating the version into major, minor, patch and pre-release identifiers in that order (Build metadata does not figure into precedence).
> 
> When major, minor, and patch are equal, a pre-release version has lower precedence than a normal version.
> 
> Precedence for two pre-release versions with the same major, minor, and patch version MUST be determined by comparing each dot separated identifier from left to right until a difference is found as follows.
>
> A larger set of pre-release fields has a higher precedence than a smaller set, if all of the preceding identifiers are equal.
> 
> Example: 1.0.0-alpha < 1.0.0-alpha.1 < 1.0.0-alpha.beta < 1.0.0-beta < 1.0.0-beta.2 < 1.0.0-beta.11 < 1.0.0-rc.1 < 1.0.0.

mjameswh (approved):

Directory.Build.props
cretz: 👍 We adhere to https://learn.microsoft.com/en-us/nuget/concepts/package-versioning which prefers `-beta2` over `-beta.2` (i.e. semver v1 instead of semver v2 though it still works w/ v2). During alpha/beta I am not seeing much value in bumping actual revision numbers since I have to provide the suffix anyways. This is before-first-release project-wide beta as opposed to a beta concept we may have one day where we beta release a specific version. _Technically_ this is more like `1.0.0-beta2` but I am zerover'ing it to make it clear to users we haven't made a GA release.

PR #152, opened 2023-11-08
Update README.md to remove note about DI issue

jakejscott:
## What was changed
Removed note that I don't think it relevant anymore

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


cretz (approved):

PR #153, opened 2023-11-08
Log forwarding

### LINKED ISSUE #140: [Feature Request] Add support for Log forwarding to lang-side

Add support for Log forwarding to lang-side. This would be functionally similar to what already exists in TypeScript SDK, though we may want to consider a solution for .NET where logs are passed directly from core to lang side, without buffering (akin to what has been done for metrics forwarding).

A corresponding option already exists (as [seen here](https://github.com/temporalio/sdk-dotnet/blob/ab4ded44cb15c2aba6ef7472e712ba12f3fef6ad/src/Temporalio/Bridge/src/runtime.rs#L212-L215)) and triggers registration of the Log forwarding export layer in Bridge, but log entries never get polled by lang, so this practically behave as a noop log exporter.

cretz:
## What was changed

* Added `Temporalio.Runtime.LogForwardingConfig` that accepts a `ILogger` to forward logs to
* Added `Temporalio.Runtime.LoggingOptions.Forwarding` to accept the above
* Added native layer to send logs to callback

## Checklist

1. Closes #140

### DIFF
--- a/src/Temporalio/Bridge/OptionsExtensions.cs
+++ b/src/Temporalio/Bridge/OptionsExtensions.cs
@@ -131,7 +131,8 @@ namespace Temporalio.Bridge
             return new Interop.LoggingOptions()
             {
                 filter = scope.ByteArray(options.Filter.FilterString),
-                forward = (byte)0,
+                // Forward callback is set in the Runtime constructor
+                // forward_to = ???
             };
         }
 
--- a/src/Temporalio/Bridge/OptionsExtensions.cs
+++ b/src/Temporalio/Bridge/OptionsExtensions.cs
@@ -132,7 +132,7 @@ namespace Temporalio.Bridge
             {
                 filter = scope.ByteArray(options.Filter.FilterString),
                 // Forward callback is set in the Runtime constructor
-                // forward_to = ???
+                // forward_to = <set-elsewhere>
             };
         }
 
--- a/src/Temporalio/Bridge/src/runtime.rs
+++ b/src/Temporalio/Bridge/src/runtime.rs
@@ -4,18 +4,25 @@ use crate::ByteArray;
 use crate::ByteArrayRef;
 use crate::MetadataRef;
 
+use serde_json::json;
+use std::fmt;
 use std::net::SocketAddr;
 use std::str::FromStr;
-use std::sync::Arc;
+use std::sync::atomic::AtomicBool;
+use std::sync::atomic::Ordering;
+use std::sync::{Arc, Mutex};
 use std::time::Duration;
+use std::time::UNIX_EPOCH;
 use temporal_sdk_core::telemetry::{build_otlp_metric_exporter, start_prometheus_metric_exporter};
 use temporal_sdk_core::CoreRuntime;
 use temporal_sdk_core_api::telemetry::metrics::CoreMeter;
 use temporal_sdk_core_api::telemetry::MetricTemporality;
+use temporal_sdk_core_api::telemetry::{CoreLog, CoreLogConsumer};
 use temporal_sdk_core_api::telemetry::{
     Logger, OtelCollectorOptionsBuilder, PrometheusExporterOptionsBuilder,
     TelemetryOptions as CoreTelemetryOptions, TelemetryOptionsBuilder,
 };
+use tracing::Level;
 use url::Url;
 
 #[repr(C)]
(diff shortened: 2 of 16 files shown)


src/Temporalio/Bridge/OptionsExtensions.cs
```diff
@@ -131,7 +131,8 @@ internal static class OptionsExtensions
             return new Interop.LoggingOptions()
             {
                 filter = scope.ByteArray(options.Filter.FilterString),
-                forward = (byte)0,
+                // Forward callback is set in the Runtime constructor
+                // forward_to = ???
```
Sushisource: Leftover line?

src/Temporalio/Bridge/src/runtime.rs
```diff
@@ -32,7 +39,36 @@ pub struct TelemetryOptions {
...
+    core: CoreLog,
+    fields_json: Arc<Mutex<Option<String>>>,
+}
+
+#[repr(C)]
+pub struct ForwardedLogDetails {
+    target: ByteArray,
+    message: ByteArray,
+    timestamp_millis: u64,
+    fields_json: ByteArray,
```
Sushisource, line 62: I'm confused why we're converting the fields to JSON here rather than passing something structured back into .NET? Can we not convert into some .NET object here and send that over in `ForwardedLog` rather than doing this serialization?

Sushisource (commented):
One main question.

Also I noticed we're not sending the `span_contexts` field from core (and I think I missed that in the Python PR). That's useful to send and have show up somehow as it provides a good bit of context (can kind of be considered part of the target)

src/Temporalio/Bridge/OptionsExtensions.cs
cretz: Intentional, as a way to say "I know I am not setting this field here". Maybe `???` should be `<set elsewhere>`

EDIT: Updated

src/Temporalio/Bridge/src/runtime.rs
cretz: To pass something structured over the C FFI boundary means I build an entire object model in C to represent that. This usually means a recursive union with a new enum and figure out lifetime issues (vs copying). It's just more complicated, I don't already have `serde_json::Value` at my disposal in C like you have in Rust. It's a similar question to why I send proto bytes to/from core instead of creating structures to represent those.

But if it's worth it, I can do it. For now, I'm just converting to `Dictionary<string, JsonElement>` on the .NET side anyways. The other option I considered is to just pass a "fields_suffix" string for now that is built Rust side, but I figured I might as well provide the somewhat-structured field.

cretz:
> Also I noticed we're not sending the span_contexts field from core (and I think I missed that in the Python PR). That's useful to send and have show up somehow as it provides a good bit of context (can kind of be considered part of the target)

I left these off as I couldn't see a good way of them making it into log and I'm hesitant to bring in more data just for the log record if it's not on the default log message currently. But if they should be on the default log message it can make more sense, any suggestions of log message format welcome. Otherwise, I was just gonna wait until someone asked.

cretz:
I am seeing some segfaults on Ubuntu ARM with the new logging test which is scary. I am moving this into draft mode while I do the push-debug-code-watch-CI dance to see if I can figure out what is happening.

src/Temporalio/Bridge/src/runtime.rs
Sushisource: Yeah, I get that - I just thought it would probably be easy enough to make the right structure here since there's only a handful of types. I think it's probably worth it in the sense that it's not code we're likely to have any problem with, just a bit of extra boilerplate but it'll be quite a bit more efficient on a fairly hot path.

cretz: I don't think it's going to be that much more efficient since I am turning it back into a string anyways on the .NET side. So I guess I can just let the Rust side turn it into a string for me and then we'll have the best performance. The only reason to use structured data is if the user wants structured data. If we're concerned about performance, we should make the user decide to turn it to structured (which they can given a JSON string).

So now I've come full circle back to my current impl except not expose it as a `Dictionary<string, JsonElement>` but rather `Dictionary<string, string>` where values are JSON strings and if a user wants more structure, they can parse values.

cretz: Ok, I updated this to give keys and values where the values are raw JSON. This is the best performing I can do, and is better than converting each nested Rust object into a nested .NET object just to turn it to a string in the log message on the .NET side. Having Rust just serialize JSON is plenty fast for this use case (and not worth a new C map structure).

cretz:
Segfault lifetime issue fixed

src/Temporalio/Bridge/src/runtime.rs
Sushisource: :+1: makes sense to me. What I missed originally is that `ILogger` doesn't have any explicit structured k/v pairs in the interface anyway, so yeah there's no "place" to offer the structured objects.

Sushisource (approved):

src/Temporalio/Bridge/src/runtime.rs
cretz: > there's no "place" to offer the structured objects.

Well, there is in that a logger gets an arbitrary "state" which is an enumerable of k/v pairs. So advanced users can use that to get the fields out (we do in our tests). But the default is to append the field values stringified anyways. So I'd prefer to make that the performant path and force users to deserialize JSON if they need more advanced structure if they happen to be doing something advanced with the values.

PR #154, opened 2023-11-10
Demonstrate OpenTelemetry + dependency injection

glencbz:
Hi Temporal folks! Here's a doc change for something I figured out that I wish was called out in the docs. Feel free to reword this as you wish.
## What was changed

Add an example of configuring OpenTelemetry with dependency injection to src/Temporalio.Extensions.OpenTelemetry/README.md.

## Why?

src/Temporalio.Extensions.OpenTelemetry/README.md doesn't contain an example of how to configure OpenTelemetry when using dependency injection. It's pretty easy to figure out how to configure the client by also reading the src/Temporalio.Extensions.Hosting/README.md, _but_ if one uses a hosted worker, it typically has to be configured separately. This is a common pitfall (according to
https://temporalio.slack.com/archives/C012SHMPDDZ/p1699016782885319?thread_ts=1698974629.231319&cid=C012SHMPDDZ at least), so document this.

## Checklist

1. Closes 

NA

2. How was this tested:

NA

3. Any docs updates needed?

      
NA

### DIFF
--- a/src/Temporalio.Extensions.OpenTelemetry/README.md
+++ b/src/Temporalio.Extensions.OpenTelemetry/README.md
@@ -53,6 +53,40 @@ var client = await TemporalClient.ConnectAsync(new("localhost:7233")
 // Now client and worker calls are traced...
 ```
 
+If you are using dependency injection with `Temporalio.Extensions.Hosting`, you have several options depending on
+how you choose to set up your client and worker.
+
+Option 1. A client and worker where the worker reuses the client: 
+
+```csharp
+// In Program.cs
+// NOTE: Remember to setup the tracer provider like in the earlier example.
+var builder = WebApplication.CreateBuilder(args);
+builder.Services
+    .AddTemporalClient(opts =>
+        {
+            opts.TargetHost = "localhost:7233";
+            opts.Namespace = "default";
+            opts.Interceptors = new[] { new TracingInterceptor() };
+        })
+    .AddHostedTemporalWorker("my-task-queue")
+```
+
+Option 2. A worker with no client: 
+
+```csharp
+// In Program.cs
+// NOTE: Remember to setup the tracer provider like in the earlier example.
+var builder = WebApplication.CreateBuilder(args);
+builder.Services
+    .AddHostedTemporalWorker("localhost:7233", "default", "my-task-queue")
+    .ConfigureOptions(opts =>
+      {
+          opts.Interceptors = new[] { new TracingInterceptor() };
+          opts.AddWorkflow<MyWorkflow>();
+      });
+```
+
 ## How it Works
 
 The tracing interceptor uses OpenTelemetry context propagation to serialize the currently active diagnostic activity as


src/Temporalio.Extensions.OpenTelemetry/README.md
```diff
@@ -53,6 +53,45 @@ var client = await TemporalClient.ConnectAsync(new("localhost:7233")
...
+var builder = WebApplication.CreateBuilder(args);
+builder.Services
+    // Create a client to localhost on "default" namespace with the tracing
+    // interceptor
+    .AddTemporalClient(opts =>
+        {
+            opts.TargetHost = "localhost:7233";
+            opts.Namespace = "default";
+            opts.Interceptors = new[] { new TracingInterceptor() };
+        })
```
cretz, line 71: Many users don't even configure a client in their worker processes when setting up with generic host support. I wonder if we should make that clear.

src/Temporalio.Extensions.OpenTelemetry/README.md
```diff
@@ -53,6 +53,45 @@ var client = await TemporalClient.ConnectAsync(new("localhost:7233")
...
+    // Create a worker to localhost on "default" namespace with the tracing
+    // interceptor
+    .AddHostedTemporalWorker("localhost:7233", "default", "my-task-queue")
```
cretz, line 82: Per the hosting README, you could use the task-queue-only `AddHostedTemporalWorker` overload and it _will_ use the `ITemporalClient` from DI, which would make more sense in this example (why not reuse the client?). From the README:

> One overload of `AddHostedTemporalWorker`, used in the quick start sample above, accepts the client target host, the client namespace, and the worker task queue. This form will connect to a client for the worker. The other overload of `AddHostedTemporalWorker` only accepts the worker task queue. In the latter, an `ITemporalClient` can either be set on the services container and therefore reused across workers, or the resulting builder can have client options set.

I wonder if it's worth explaining that.

src/Temporalio.Extensions.OpenTelemetry/README.md
glencbz: Hm, I have two interpretations of what you're suggesting to clarify:

* What this client setup code is doing
* How a configured client affects the worker process

Though in either case, I think it isn't this README's job to explain generic host support. If it's clear to users that they should look at https://github.com/temporalio/sdk-dotnet/blob/main/src/Temporalio.Extensions.Hosting/README.md for that, I think we've done our job here. If you think _that_ isn't clear, I'm happy to add the pointer :)

src/Temporalio.Extensions.OpenTelemetry/README.md
glencbz: Oh! TIL. I thought it was weird that I had to configure the client twice, I must have missed it skimming that README (I'm bad at reading prose 😅 ). Like before, I think it isn't this README's job to explain hosting, but it does make sense to give good examples that replicate real hosting scenarios.

AIUI, the task-queue only overload should be a one-liner, so I think I can fit it into the same code block (perhaps in a comment) + an explanation of when users will need to use which.

src/Temporalio.Extensions.OpenTelemetry/README.md
cretz: > Though in either case, I think it isn't this README's job to explain generic host support

Right, don't need to explain generic host support, it's just a bit confusing to see `AddTemporalClient` and `AddHostedTemporalWorker` both here. Either remove the former or make the latter use the overload that works with the former.

src/Temporalio.Extensions.OpenTelemetry/README.md
cretz: >  it does make sense to give good examples that replicate real hosting scenarios.

But this is a bad example because you're creating a client twice unnecessarily. Either remove `AddTemporalClient` or change `AddHostedTemporalWorker` to use the overload that uses that client.

> AIUI, the task-queue only overload should be a one-liner, so I think I can fit it into the same code block (perhaps in a comment) + an explanation of when users will need to use which.

Sure. You can show both ways. You can show a client + worker that uses client, and a no-client + worker that creates a client internally. Feel free to get rid of the tracer provider part of the sample as that's likely implied from the previous samples.

glencbz: >> it does make sense to give good examples that replicate real hosting scenarios.

> But this is a bad example because you're creating a client twice unnecessarily. Either remove AddTemporalClient or change AddHostedTemporalWorker to use the overload that uses that client.

Ah yes, that was what I meant by that comment (that my example is not great).

> Feel free to get rid of the tracer provider part of the sample as that's likely implied from the previous samples.

Yeah I was on the fence about this. I'll probably leave a comment to the effect of "set up the tracer like in the previous example".

glencbz:
Updated and removed some cruft from the project I copy-pasted from.

src/Temporalio.Extensions.OpenTelemetry/README.md
```diff
@@ -53,6 +53,38 @@ var client = await TemporalClient.ConnectAsync(new("localhost:7233")
...
+// reuse the client in the worker.
+builder.Services
+    .AddTemporalClient(opts =>
+        {
+            opts.TargetHost = "localhost:7233";
+            opts.Namespace = "default";
+            opts.Interceptors = new[] { new TracingInterceptor() };
+        })
+    .AddHostedTemporalWorker("my-task-queue")
+// Option 2: If you have a worker but no client, you can configure the worker directly
```
cretz: Sorry to be pedantic, but can you move this to a separate snippet? Can explain the options above the snippet if you'd like instead of comments, but as is, this snippet if taken as a whole is a bit broken.

cretz (commented):
Looks great! Just a minor pedantic thing about breaking up the snippet, then will merge. Thanks!

src/Temporalio.Extensions.OpenTelemetry/README.md
glencbz: Done!

cretz (approved):

PR #164, opened 2023-12-04
Handle new .NET 8 analyzer suggestions

cretz:
## What was changed

.NET 8 has new analyzer suggestions that are failing builds

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -62,8 +62,9 @@ jobs:
 
       - name: Regen confirm unchanged
         if: ${{ matrix.checkTarget }}
+        # Needs to be 16.0.0 until https://github.com/dotnet/ClangSharp/pull/506 released
         run: |
-          dotnet tool install --global ClangSharpPInvokeGenerator
+          dotnet tool install --global --version 16.0.0 ClangSharpPInvokeGenerator
           ClangSharpPInvokeGenerator @src/Temporalio/Bridge/GenerateInterop.rsp
           dotnet run --project src/Temporalio.Api.Generator
           npx doctoc README.md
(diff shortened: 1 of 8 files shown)


cretz:
I am moving this to draft while I debug some CI issues

cretz:
It looks like .NET started setting `DeterministicSourcePaths` to `true` (or similar, setting base source paths to `/_`). I fixed my reliance on source paths in tests and we're good now. Marking ready for review.

.github/workflows/ci.yml
```diff
@@ -62,8 +62,9 @@ jobs:
 
       - name: Regen confirm unchanged
         if: ${{ matrix.checkTarget }}
+        # Needs to be 16.0.0 until https://github.com/dotnet/ClangSharp/pull/506 released
         run: |
-          dotnet tool install --global ClangSharpPInvokeGenerator
+          dotnet tool install --global --version 16.0.0 ClangSharpPInvokeGenerator
```
cretz, line 67: They broke ClangSharpPInvokeGenerator last week

Directory.Build.props
```diff
@@ -3,7 +3,6 @@
     <AnalysisLevel>latest</AnalysisLevel>
     <AnalysisMode>AllEnabledByDefault</AnalysisMode>
     <Authors>Temporal</Authors>
-    <ContinuousIntegrationBuild Condition="'$(GITHUB_ACTIONS)' == 'true'">true</ContinuousIntegrationBuild>
```
cretz, line 6: .NET 8 release (or MSBuild release) seems to now be properly setting deterministic source paths when this is set, which means it strips runtime paths which hurts us because we test with them.

Quinn-With-Two-Ns (approved):

Issue #169, opened 2023-12-29
[Bug] Forcing .net 7 when .net 6 is in LTS


rsr-maersk:
Hi

Thanks for the package. Qhy have you not build in .net 6 LTS?

System.Diagnostics.DiagnosticSource .net 7 is renowned for giving net 6 issues.

https://github.com/temporalio/sdk-dotnet/blob/b64dd6def30060ca29aabd3bfec70848ac36d052/src/Temporalio.Extensions.OpenTelemetry/Temporalio.Extensions.OpenTelemetry.csproj#L16

https://github.com/temporalio/sdk-dotnet/blob/b64dd6def30060ca29aabd3bfec70848ac36d052/src/Temporalio.Extensions.Hosting/Temporalio.Extensions.Hosting.csproj#L15

Can we build net6 packages? I can make the csproj spit it all out :-)

![image](https://github.com/temporalio/sdk-dotnet/assets/76153851/627e1827-e116-42f3-9775-ddb88d4ea7e3)

cretz:
> System.Diagnostics.DiagnosticSource .net 7 is renowned for giving net 6 issues.

Can you clarify the issues? The .NET version is unrelated to that dependency version. We support many versions for this extension, see https://www.nuget.org/packages/Temporalio.Extensions.Hosting#supportedframeworks-body-tab. And that dependency supports .NET 6 and many others, see https://www.nuget.org/packages/System.Diagnostics.DiagnosticSource#supportedframeworks-body-tab. In fact, our test case project is `net6.0` and tests this project with these dependencies.

rsr-maersk:
Thanks for answer @cretz .
It does give a lot of issues with AF Funcs.
When on net 6, it is best to keep everything on .net6. Trust me, you think this ddl and that dll are safe, then it ends up using something else, and no longer safe.

My recommendation. keep 6 on 6, 7 on 7 and 8 on 8 ;-)

cretz:
>  It does give a lot of issues with AF Funcs.

What are AF Funcs? Can you show an example of such an issue? Our unit tests are all on .NET 6 using these 7.x libraries. A unit test on .NET 6 that would break here would help us know if we need to downgrade.

cretz:
(closing pending response, but feel free to update if there is more info or a standalone replication)

PR #170, opened 2023-12-29
Use .Net 6.0 LTS for Microsoft.Extensions.Hosting

rsr-maersk:
#169 reduce Microsoft.Extensions.Hosting to .net 6 to allow current LTS support


cretz:
I added a comment on #169. The version of `Microsoft.Extensions.Hosting` is not directly related to the .NET version though they are similar. Notice it supports many versions: https://www.nuget.org/packages/Microsoft.Extensions.Hosting#supportedframeworks-body-tab. Our test case project is `net6.0` and tests this project with these dependencies.

PR #172, opened 2024-01-04
Fix issue with DI-based static async activities

cretz:
## What was changed

There was an issue where we weren't properly unwrapping an async static result in the DI-based activity invoker because statics took a different code path. The consolidates the code paths to fix the issue.


bergundy (approved):
Rubber stamping

PR #175, opened 2024-01-10
Properly ignore unknown fields when parsing history JSON

### LINKED ISSUE #173: [Bug] Not ignoring unknown fields when parsing history for replay
### Describe the bug

Python and Go and others ignore unknown fields when parsing JSON for history proto, but .NET doesn't. Even though docs don't show it, there is an ignore-unknown-fields setting on JsonParser settings.

cretz:
## What was changed

We weren't properly ignoring unknown fields when parsing history JSON for replay, we are now. There is an argument to be made for/against whether we _should_, but other SDKs do.

## Checklist

1. Closes #173


dandavison (approved):
LGTM!

PR #180, opened 2024-01-22
Prefix some errors with rule identifiers

cretz:
## What was changed

Add a couple of https://github.com/temporalio/rules references as error prefixes

### DIFF
--- a/src/Temporalio/Worker/WorkflowWorker.cs
+++ b/src/Temporalio/Worker/WorkflowWorker.cs
@@ -138,7 +138,7 @@ namespace Temporalio.Worker
                     // deadlocked and just rethrow the same exception as before
                     if (deadlockedWorkflows.ContainsKey(act.RunId))
                     {
-                        throw new InvalidOperationException($"Workflow with ID {act.RunId} deadlocked after {deadlockTimeout}");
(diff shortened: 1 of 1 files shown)


src/Temporalio/Worker/WorkflowWorker.cs
```diff
@@ -167,7 +167,7 @@ private async Task HandleActivationAsync(IPayloadCodec? codec, WorkflowActivatio
                             // next activation which will be a remove job.
                             deadlockedWorkflows[act.RunId] = workflowTask;
                             throw new InvalidOperationException(
-                                $"Workflow with ID {act.RunId} deadlocked after {deadlockTimeout}");
+                                $"[TMPRL1101] Workflow with ID {act.RunId} deadlocked after {deadlockTimeout}");
```
Sushisource, line 170: Maybe re-use this format string?

Sushisource (approved):

src/Temporalio/Worker/WorkflowWorker.cs
cretz: Meh, just a couple of places for the string literal, I don't think it's worth extracting

PR #195, opened 2024-02-29
Eager workflow start

### LINKED ISSUE #183: [Feature Request] Eager Workflow Start
Support for Eager Workflow Start (EWS) was added in the server in 1.20, and is currently supported in Java, Python, Core, and Go SDKs. It
reduces the latency to start a workflow by allowing the client to directly schedule the first workflow task to a local worker. See https://github.com/temporalio/features/issues/242 for details.

The goal is to enable support for EWS in  the .NET SDK using the Core implementation.

This mechanism is mostly transparent to the .NET SDK client, but we need to add an optional flag when we start the workflow.

cretz:
## What was changed

Add eager workflow start option. There's no good way to test (sure we could make sure it comes back from server, but that doesn't verify behavior that server actually did eagerly route it, and it'd require a secret variable).

## Checklist

1. Closes #183

### DIFF
--- a/src/Temporalio/Client/TemporalClient.Workflow.cs
+++ b/src/Temporalio/Client/TemporalClient.Workflow.cs
@@ -462,6 +462,7 @@ namespace Temporalio.Client
                     RequestId = Guid.NewGuid().ToString(),
                     WorkflowIdReusePolicy = input.Options.IdReusePolicy,
                     RetryPolicy = input.Options.RetryPolicy?.ToProto(),
+                    RequestEagerExecution = input.Options.RequestEagerStart,
                 };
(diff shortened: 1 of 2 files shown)


antlai-temporal (approved):
This is great thanks, just 9 lines!
 So there is no way at this point to implement a grpc interceptor in .Net, is that true?

cretz:
> This is great thanks, just 9 lines!

The entire .NET SDK implementation has great/clean options separation.

> So there is no way at this point to implement a grpc interceptor in .Net, is that true?

Correct, because gRPC is done in Rust. But people do want it (and there are thoughts on how we can do it).

PR #200, opened 2024-03-06
Worker poller and ratio options

### LINKED ISSUE #196: [Feature Request] Expose max poller options for workers

We expose this in Python and TypeScript, we should expose in .NET. Also sticky to nonsticky ratio.

cretz:
## What was changed

Expose `MaxConcurrentWorkflowTaskPolls`, `NonStickyToStickyPollRatio`, and `MaxConcurrentActivityTaskPolls`. There's not an easy/good way to test, though if really wanted, I can make a test that checks metrics to confirm poller count goes up (would need to check how it reaches "max").

## Checklist

1. Closes #196


mjameswh (approved):
LGTM

Sushisource (approved):

PR #202, opened 2024-03-07
Fix issue with unconfigured metric meter

### LINKED ISSUE #135: [Feature Request] Confirm absense of metric config still results in no-op metric meter

When developing this feature in Python, it appeared that the metric meter was not created when metrics config is not present. So we need to use a no-op one

cretz:
## What was changed

It was a bug today that if you used a runtime metric meter but had never configured one, you'd get a scary segfault. This fixes that to set a no-op meter.

## Checklist

1. Closes #135


mjameswh (approved):
Makes sense.

Sushisource (approved):

PR #205, opened 2024-03-18
Ability for certain task failure types to fail workflow

### LINKED ISSUE #166: [Feature Request] Ability for certain task failure types to fail workflow

From https://github.com/temporalio/features/issues/322, need:

* `IReadOnlyCollection<Type> WorkflowFailureExceptionTypes` on `TemporalWorkerOptions`
* `Type[] FailureExceptionTypes` on `WorkflowAttribute` and `IReadOnlyCollection<Type> FailureExceptionTypes` on `WorkflowDefinition`
* Disallow `InvalidWorkflowOperationException` until https://github.com/temporalio/sdk-core/issues/656 is done

cretz:
## What was changed

* Added `TemporalWorkerOptions.WorkflowFailureExceptionTypes` to allow configuring failure exception types at a worker level
* Added `WorkflowAttribute.FailureExceptionTypes` to allow configuring failure exception types at a workflow level
* Update core to get https://github.com/temporalio/sdk-core/pull/702
* Abstracted `InvalidWorkflowOperationException` into separate exceptions for `InvalidWorkflowSchedulerException` and `WorkflowNondeterminismException`
  * This exception used to be used for both, but it's abstracted now so the latter can be referenced as a failure type
* Updated bridge code to include worker-level and workflow-type-level non-determinism-as-fail based on the presence of `WorkflowNondeterminismException` (or super type) in worker/workflow options
* Documented in README, API doc, and marked experimental
* Several tests for several scenarios

## Checklist

1. Closes #166


src/Temporalio/Exceptions/WorkflowNondeterminismException.cs
```diff
@@ -0,0 +1,22 @@
...
+{
+    /// <summary>
+    /// Occurs when the workflow does something non-deterministic.
+    /// </summary>
+    /// <remarks>
+    /// This is usually only thrown in a replayer and therefore can't be caught in a workflow, but
+    /// this exception can be used as a failure exception type to have non-deterministic exceptions
+    /// fail workflows.
+    /// </remarks>
+    public class WorkflowNondeterminismException : InvalidWorkflowOperationException
```
cretz, line 11: So Java capitalizes the `D` here but core does not, open to suggestions

src/Temporalio/Bridge/include/temporal-sdk-bridge.h
```diff
@@ -358,6 +358,11 @@ typedef struct WorkerOptions {
   uint32_t max_concurrent_workflow_task_polls;
   float nonsticky_to_sticky_poll_ratio;
   uint32_t max_concurrent_activity_task_polls;
+  bool nondeterminism_as_workflow_fail;
+  /**
+   * This is expected to be a newline-delimited list
+   */
+  struct ByteArrayRef nondeterminism_as_workflow_fail_for_types;
```
Sushisource: Why have these specially for nondeterminism rather than passing through the types and converting that to the Rust types? 

If/when we add more types there will need to be two fields for every future type rather than just adding a branch to the converter

src/Temporalio/Bridge/src/worker.rs
```diff
@@ -503,6 +509,26 @@ impl TryFrom<&WorkerOptions> for temporal_sdk_core::WorkerConfig {
...
+            .workflow_failure_errors(if opt.nondeterminism_as_workflow_fail {
+                HashSet::from([WorkflowErrorType::Nondeterminism])
+            } else {
+                HashSet::new()
+            })
+            .workflow_types_to_failure_errors(
+                opt.nondeterminism_as_workflow_fail_for_types
+                    .to_option_str()
+                    .map(|s| {
+                        s.split('\n')
```
Sushisource: You could make the to map helper for this generic and use it here.

I'm still not a fan of parsing on newlines like this, but it's not hugely important to change.

src/Temporalio/Worker/TemporalWorkerOptions.cs
```diff
@@ -214,6 +214,21 @@ public TemporalWorkerOptions()
...
+        /// cause the workflow/update to fail instead of suspending the workflow via task failure.
+        /// These are applied in addition to <see cref="WorkflowAttribute.FailureExceptionTypes" />
+        /// on a specific workflow. If <c>typeof(Exception)</c> is set, it effectively will fail a
+        /// workflow/update in all user exception cases.
+        /// </summary>
+        /// <remarks>
+        /// WARNING: This property is experimental and may change in the future. If unset
+        /// (i.e. left null), currently the default is to only fail the workflow/update on
+        /// <see cref="Exceptions.FailureException" /> + cancellation and suspend via task failure
+        /// all others. But this default may change in the future.
```
Sushisource, line 228:
```suggestion
        /// <see cref="Exceptions.FailureException" />. Other exception types will cause either cancellation or suspension via task failure. This default may change in the future.
```

src/Temporalio/Workflows/WorkflowAttribute.cs
```diff
@@ -45,5 +45,21 @@ public WorkflowAttribute(string name)
...
+        /// These are applied in addition to
+        /// <see cref="Worker.TemporalWorkerOptions.WorkflowFailureExceptionTypes" /> for the
+        /// overall worker. If <c>typeof(Exception)</c> is set, it effectively will fail a
+        /// workflow/update in all user exception cases.
+        /// </summary>
+        /// <remarks>
+        /// WARNING: This property is experimental and may change in the future. If unset
+        /// (i.e. left null), currently the default is to only fail the workflow/update on
+        /// <see cref="Exceptions.FailureException" /> + cancellation and suspend via task failure
+        /// all others. But this default may change in the future.
```
Sushisource, line 61:
```suggestion
        /// <see cref="Exceptions.FailureException" />. Other exception types will cause either cancellation or suspension via task failure. This default may change in the future.
```

src/Temporalio/Workflows/WorkflowDefinition.cs
```diff
@@ -145,6 +153,10 @@ public class WorkflowDefinition
             {
                 errs.Add("Cannot have custom name for dynamic workflow");
             }
+            if (name != null && name.Contains('\n'))
+            {
+                errs.Add("Workflow name cannot have a newline");
+            }
```
Sushisource: This is a good example why the newline thing scares me.

I mean, it's dumb to have a newline in a workflow name *anyway*, but, still.

src/Temporalio/Bridge/src/worker.rs
cretz: I already have a map helper for metadata (and it's newline delimited too), but I admit being lazy, because getting collections of strings right in a non-annoying way over FFI boundary requires basically building your own structures (or saying something like `\0` delimited) and I was hoping to avoid it.

src/Temporalio/Bridge/include/temporal-sdk-bridge.h
cretz: Because this is a private FFI boundary and I only expose the most I need. I am trying to avoid overcomplicating the FFI code until there is a second type. I would not be surprised if that enum stays as only one item forever, heh, but it's much easier for Rust to build those abstractions than C.

src/Temporalio/Worker/TemporalWorkerOptions.cs
cretz: This isn't necessarily true with how the code is today. I trap cancellation errors and return workflow fail even when cancellation was not requested because of how common they appear and that it's unfair to have already-started timer cancellation be a wf fail but not-yet-started timer cancellation be a task fail.

src/Temporalio/Workflows/WorkflowDefinition.cs
cretz: It already occurs with metadata today, but I am preventing the newline explicitly. I don't think this use case is worth me building my own C string vec representation. But I can if I must.

src/Temporalio/Bridge/include/temporal-sdk-bridge.h
Sushisource: Sure, but, you've got to do the conversion either way. It's either the same amount of complicated, or if we ever add anything else, less complicated. You can just pass through an an enum or something, super simple, and less work to extend later.

cretz: It's more complicated to do C arrays (not "super simple") and I have been lucky to avoid them so far because I can make expectations about how I can split single-value fields. But it sounds like my luck is running out, ug.

src/Temporalio/Workflows/WorkflowDefinition.cs
Sushisource: Well does that really mean we need to build our own? I've got to imagine this comes up fairly often...

And yeah, I know it's happening with metadata but the fact that we'll sometimes have to deal with edge cases like this as we add new things says to me it probably is worth a little bit of effort to just deal with that problem for good.

cretz: In C what you have to do is accept a pointer to the first element and then a length. And each element needs to know its size. And with ownership rules over FFI, it means each element needs to remain owned on the lang side. So it's a pointer to an owned array of pointers to owned strings. But this is totally doable, I've just avoided it because I've never needed it and it's ugly/dangerous. Didn't seem worth it just to keep newlines in workflow types, but I will try to add it now.

src/Temporalio/Workflows/WorkflowAttribute.cs
cretz: See other comment on similar change

src/Temporalio/Workflows/WorkflowDefinition.cs
cretz: I just added string arrays in [this commit](https://github.com/cretz/temporal-sdk-dotnet/commit/07524da4cfb568d61213f58420b324c5b1d10a31) and removed the newline restriction

src/Temporalio/Bridge/src/worker.rs
cretz: I just added string arrays in [this commit](https://github.com/cretz/temporal-sdk-dotnet/commit/07524da4cfb568d61213f58420b324c5b1d10a31) and removed the newline restriction

src/Temporalio/Bridge/include/temporal-sdk-bridge.h
cretz: I just added string arrays in [this commit](https://github.com/cretz/temporal-sdk-dotnet/commit/07524da4cfb568d61213f58420b324c5b1d10a31) and removed the newline restriction

Sushisource (approved):
Cool! Yeah, that looked pretty easy and worth it.

PR #208, opened 2024-03-25
Remove experimental flag from StartDelay

tlalfano:
Remove experimental flag from StartDelay

## What was changed
Remove experimental flag from StartDelay

## Why?
To indicate the current state of StartDelay workflow option

### DIFF
--- a/src/Temporalio/Client/WorkflowOptions.cs
+++ b/src/Temporalio/Client/WorkflowOptions.cs
@@ -1,4 +1,4 @@
-using System;
+fusing System;
 using System.Collections.Generic;
 using System.Linq.Expressions;
 using System.Threading.Tasks;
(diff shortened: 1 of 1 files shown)


src/Temporalio/Client/WorkflowOptions.cs
```diff
@@ -1,4 +1,4 @@
-using System;
+fusing System;
```
cretz: Accidental typo here

tlalfano (commented):
Good catch!

src/Temporalio/Client/WorkflowOptions.cs
tlalfano: good catch!

cretz (approved):
Thanks! Merged `main` into your branch, and I'll merge into `main` if/when CI passes.

cretz:
Thanks!

PR #216, opened 2024-04-08
Use build ID in the unique DI identifier for workers

### LINKED ISSUE #212: [Bug] Cannot add multiple versions of a workflow using DI

Trying to add two versions of the same workflow using dependency injection.

### Describe the bug

It's possible to add multiple workflows using this method, the part that makes it break is having two workflows with the same name in the `WorkflowAttribute`, even though they are being registered with different build IDs.

 I talked with @cretz who said:
> Yes, this is a bug. This is because the DI setup predates versioning, so we only supported one worker per task queue in a service collection: [https://github.com/temporalio/sdk-dotnet/blob/c6b7aca05cdd2b61e4b99098f3f390469232[…]3a4a/src/Temporalio.Extensions.Hosting/TemporalWorkerService.cs](https://github.com/temporalio/sdk-dotnet/blob/c6b7aca05cdd2b61e4b99098f3f3904692323a4a/src/Temporalio.Extensions.Hosting/TemporalWorkerService.cs#L90). We need to change what we uniquely identify options set as on the container to include build ID

### Minimal Reproduction
```csharp
[Workflow("my-workflow")]
public class TestWorkflow
{
    [WorkflowRun]
    public Task RunAsync() => Workflow.DelayAsync(TimeSpan.FromSeconds(3));
}

[Workflow("my-workflow")]
public class TestWorkflowV2
{
    [WorkflowRun]
    public Task RunAsync() => Workflow.DelayAsync(TimeSpan.FromSeconds(1));
}

static async Task Main(string[] args)
{
    await using var env = await WorkflowEnvironment.StartTimeSkippingAsync();

    IHost host = Host.CreateDefaultBuilder()
        .ConfigureServices(services =>
        {
            services.AddSingleton(env.Client);
            services
                .AddHostedTemporalWorker("task-queue")
                    .ConfigureOptions(options =>
                    {
                        options.UseWorkerVersioning  = true;
                        options.BuildId                            = "1.0";
                    })
                    .AddWorkflow<TestWorkflow>();

            services
                .AddHostedTemporalWorker("task-queue")
                    .ConfigureOptions(options =>
                    {
                        options.UseWorkerVersioning  = true;
                        options.BuildId                           = "2.0";
                    })
                    .AddWorkflow<TestWorkflowV2>();
        })
        .Build();

    host.Run();
}
```
### Environment/Versions

- OS and processor: [Windows]
- Temporal Version: SDK 1.0.0

cretz:
## What was changed

* Added build ID option when adding a worker to the service collection
* Made sure to reference options by task queue + build ID instead of just task queue
* Throw an exception if users try to create the same worker identifier twice or if they change the options from what the identifier expects

## Checklist

1. Closes #212

### DIFF
--- a/src/Temporalio.Extensions.Hosting/TemporalHostingServiceCollectionExtensions.cs
+++ b/src/Temporalio.Extensions.Hosting/TemporalHostingServiceCollectionExtensions.cs
@@ -28,8 +28,8 @@ namespace Microsoft.Extensions.DependencyInjection
         /// </param>
         /// <param name="taskQueue">Task queue for the worker.</param>
         /// <param name="buildId">
-        /// Build ID for the worker. Set to non-null to opt in to versioning. If versioning wanted,
-        /// this must be set here and not later via configure options. This is because the
+        /// Build ID for the worker. Set to non-null to opt in to versioning. If versioning is
+        /// wanted, this must be set here and not later via configure options. This is because the
         /// combination of task queue and build ID make up the unique identifier for a worker in the
         /// service collection.
         /// </param>
@@ -55,8 +55,8 @@ namespace Microsoft.Extensions.DependencyInjection
         /// <param name="services">Service collection to create hosted worker on.</param>
         /// <param name="taskQueue">Task queue for the worker.</param>
         /// <param name="buildId">
-        /// Build ID for the worker. Set to non-null to opt in to versioning. If versioning wanted,
-        /// this must be set here and not later via configure options. This is because the
+        /// Build ID for the worker. Set to non-null to opt in to versioning. If versioning is
+        /// wanted, this must be set here and not later via configure options. This is because the
         /// combination of task queue and build ID make up the unique identifier for a worker in the
         /// service collection.
         /// </param>
--- a/src/Temporalio.Extensions.Hosting/TemporalHostingServiceCollectionExtensions.cs
+++ b/src/Temporalio.Extensions.Hosting/TemporalHostingServiceCollectionExtensions.cs
@@ -17,9 +17,9 @@ namespace Microsoft.Extensions.DependencyInjection
         /// Add a hosted Temporal worker service as a <see cref="IHostedService" /> that contains
         /// its own client that connects with the given target and namespace. To use an injected
         /// <see cref="ITemporalClient" />, use
-        /// <see cref="AddHostedTemporalWorker(IServiceCollection, string)" />. The worker service
-        /// will be registered as a singleton. The result is an options builder that can be used to
-        /// configure the service.
+        /// <see cref="AddHostedTemporalWorker(IServiceCollection, string, string?)" />. The worker
+        /// service will be registered as a singleton. The result is an options builder that can be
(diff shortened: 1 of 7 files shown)


src/Temporalio.Extensions.Hosting/TemporalHostingServiceCollectionExtensions.cs
```diff
@@ -17,47 +17,69 @@ public static class TemporalHostingServiceCollectionExtensions
...
-        /// not expect an injected instance and instead connect to a client on worker start. The
+        /// <see cref="AddHostedTemporalWorker(IServiceCollection, string, string, string, string?)" />
+        /// to not expect an injected instance and instead connect to a client on worker start. The
         /// worker service will be registered as a singleton. The result is an options builder that
         /// can be used to configure the service.
         /// </summary>
         /// <param name="services">Service collection to create hosted worker on.</param>
         /// <param name="taskQueue">Task queue for the worker.</param>
+        /// <param name="buildId">
+        /// Build ID for the worker. Set to non-null to opt in to versioning. If versioning wanted,
```
Sushisource:
```suggestion
        /// Build ID for the worker. Set to non-null to opt in to versioning. If versioning is wanted,
```

src/Temporalio.Extensions.Hosting/TemporalWorkerServiceOptionsBuilderExtensions.cs
```diff
@@ -167,22 +168,45 @@ public static class TemporalWorkerServiceOptionsBuilderExtensions
...
+        /// <param name="disallowDuplicates">If true, will fail if options already registered for
+        /// this builder's task queue and build ID.</param>
         /// <returns>Options builder.</returns>
         public static OptionsBuilder<TemporalWorkerServiceOptions> ConfigureOptions(
-            this ITemporalWorkerServiceOptionsBuilder builder) =>
-            builder.Services.AddOptions<TemporalWorkerServiceOptions>(builder.TaskQueue);
+            this ITemporalWorkerServiceOptionsBuilder builder,
+            bool disallowDuplicates = false)
+        {
+            // To ensure the user isn't accidentally double, we can disallow duplicates
```
Sushisource: Isn't "doubled"? Or, double what?

Sushisource (approved):

src/Temporalio.Extensions.Hosting/TemporalHostingServiceCollectionExtensions.cs
cretz: Fixed

src/Temporalio.Extensions.Hosting/TemporalWorkerServiceOptionsBuilderExtensions.cs
cretz: Fixed

cretz:
💥 Note, in #451 we have changed some behavior here where we now no longer assume this non-GA-will-be-removed form versioning when build ID is present.

PR #223, opened 2024-04-16
Support duration and float metrics

### LINKED ISSUE #209: [Feature Request] Support metric option for using seconds (and other recent core metric changes)

Need to at least expose `use_seconds_for_duration` option on Prometheus metrics from https://github.com/temporalio/sdk-core/pull/706, may also have to update other things due to changes made by that PR.

cretz:
## What was changed

* Allow `MetricMeter.CreateHistogram` to provide floating point types or `TimeSpan` in addition to already-supported integer types
* Allow `MetricMeter.CreateGauge` to provide floating point types in addition to already-supported integer types
* Added `OpenTelemetryOptions.UseSecondsForDuration` option to have OTel metrics use float seconds instead of ms
* Added `PrometheusOptions.UseSecondsForDuration` option to have Prom metrics use float seconds instead of ms
* `ICustomMetricMeter.CreateHistogram` now supports `double` and `TimeSpan` (for those creating custom meter impls)
  * Minor 💥 compat break, but only if the caller starts creating histograms for floats/durations or sets non-default duration format option
* `ICustomMetricMeter.CreateGauge` now supports `double` (for those creating custom meter impls)
  * Minor 💥 compat break, but only if the caller starts creating gauges for floats
* Added `MetricsOptions.CustomMetricMeterOptions` which accepts an option type whose only field is `HistogramDurationFormat` to specify whether meter impls will get duration metrics as integer milliseconds, float seconds, or time spans
* Tests

## Checklist

1. Closes #209

### DIFF
--- a/src/Temporalio.Extensions.DiagnosticSource/CustomMetricMeter.cs
+++ b/src/Temporalio.Extensions.DiagnosticSource/CustomMetricMeter.cs
@@ -13,6 +13,12 @@ namespace Temporalio.Extensions.DiagnosticSource
     /// Implementation of <see cref="ICustomMetricMeter" /> for a <see cref="Meter" /> that can be
     /// set on <see cref="MetricsOptions.CustomMetricMeter" /> to record metrics to the meter.
     /// </summary>
+    /// <remarks>
+    /// By default all histograms are set as a <c>long</c> of milliseconds unless
+    /// <see cref="MetricsOptions.CustomMetricMeterOptions"/> is set to <c>FloatSeconds</c>.
+    /// Similarly, if the unit for a histogram is "duration", it is changed to "ms" unless that same
+    /// setting is set, at which point the unit is changed to "s".
+    /// </remarks>
     public class CustomMetricMeter : ICustomMetricMeter
     {
         /// <summary>
@@ -35,8 +41,22 @@ namespace Temporalio.Extensions.DiagnosticSource
         /// <inheritdoc />
         public ICustomMetricHistogram<T> CreateHistogram<T>(
             string name, string? unit, string? description)
-            where T : struct =>
-            new CustomMetricHistogram<T>(Meter.CreateHistogram<T>(name, unit, description));
+            where T : struct
+        {
+            // Have to convert TimeSpan to something .NET meter can work with. For this to even
+            // happen, a user would have had to set custom options to report as time span.
+            if (typeof(T) == typeof(TimeSpan))
+            {
+                // If unit is "duration", change to "ms since we're converting here
+                if (unit == "duration")
+                {
+                    unit = "ms";
+                }
+                return (new CustomMetricHistogramTimeSpan(
+                    Meter.CreateHistogram<long>(name, unit, description)) as ICustomMetricHistogram<T>)!;
+            }
+            return new CustomMetricHistogram<T>(Meter.CreateHistogram<T>(name, unit, description));
+        }
 
         /// <inheritdoc />
         public ICustomMetricGauge<T> CreateGauge<T>(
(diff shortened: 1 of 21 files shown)


src/Temporalio.Extensions.DiagnosticSource/CustomMetricMeter.cs
```diff
@@ -35,8 +41,22 @@ public class CustomMetricMeter : ICustomMetricMeter
         /// <inheritdoc />
         public ICustomMetricHistogram<T> CreateHistogram<T>(
             string name, string? unit, string? description)
-            where T : struct =>
-            new CustomMetricHistogram<T>(Meter.CreateHistogram<T>(name, unit, description));
+            where T : struct
+        {
+            // Have to convert TimeSpan to something .NET meter can work with. For this to even
+            // happen, a user would have had to set custom options to report as time span.
+            if (typeof(T) == typeof(TimeSpan))
```
Sushisource, line 48: This whole non-erased generic type check thing feels *deeply illegal* to me 😆

src/Temporalio/Bridge/Metric.cs
```diff
@@ -51,18 +51,44 @@ internal class MetricInteger : SafeHandle
         /// </summary>
         /// <param name="value">Value to record.</param>
         /// <param name="attributes">Attributes to set.</param>
-        public void Record(ulong value, MetricAttributes attributes)
+        public void RecordInteger(ulong value, MetricAttributes attributes)
```
Sushisource, line 54: Rather than allowing all record kinds on one `Metric` type it seems like it'd be pretty easy to hand out specialized versions with only one record method each corresponding to the kind passed in when created

src/Temporalio/Bridge/src/metric.rs
```diff
@@ -455,11 +510,50 @@ impl metrics::Histogram for CustomMetricInteger {
...
+    }
+}
+
+impl metrics::HistogramDuration for CustomMetric {
+    fn record(&self, value: Duration, attributes: &metrics::MetricAttributes) {
+        unsafe {
+            let meter = &*(self.meter_impl.0);
+            (meter.metric_record_duration)(
+                self.metric,
+                value.as_millis().try_into().unwrap_or(u64::MAX),
```
Sushisource, line 532: I can't decide in the rare case that this happens if it's better to record this max or simply not record anything at all

Sushisource (approved):
Looking good. Dang C# has some nice bits but it is verboooooossseeee.

src/Temporalio.Extensions.DiagnosticSource/CustomMetricMeter.cs
cretz: Reified generics are _so nice_ in .NET compared to Java's type erasure. This is why you can call `AddWorkflow<MyWorkflowClass>()` and similar. But there are some downsides.

src/Temporalio/Bridge/Metric.cs
cretz: This is not public user-facing code. We do it in a .NET idiomatic way when it comes to the actual user-facing code. But it is nice for the C FFI if we do combine these into fewer calls and expect the caller be smart.

src/Temporalio/Bridge/src/metric.rs
cretz: I was stuck a bit here too and just settled on this. Hopefully the issue never arises. But if we think it's enough of an issue, we can look info logging instead.

This gets even worse when we convert back to .NET types because. E.g. for .NET a `ulong` (i.e. `u64`) is "not CLS-compliant", so it's best to use `long` (i.e. `i64`) instead, so I limit to max there too (default unchecked cast wraps into negative). This is much more likely of a problem, but I can't currently think of a case where Core side will go over max-i64.

cretz:
@Sushisource - merged based on approval (waited a day after my most recent comment responses), but we can revisit any of the pieces if you think they are important.

PR #226, opened 2024-04-20
Make ctor for `DefaultPayloadConverter` public to expose `JsonSerializerOptions`

### LINKED ISSUE #225: [Feature Request] Make protected DefaultPayloadConverter constructors public

These constructors were only meant for inheritors when we required converter classes instead of instances in beta forms of the SDK, but now since we let people instantiate the converter, we might as well expose these.

ElanHasson:
Resolves #225.

## What was changed
Made the ctor for `DefaultPayloadConverter` public to expose `JsonSerializerOptions`.

## Why?
This is to allow customizing the json serializer without creating a new Payload Converter and incurring additional configuration overhead.

## Checklist

1. Closes #225 

2. How was this tested:
The compiler + existing tests.

3. Any docs updates needed?
Do not think so. Maybe worth a mention of this capability via https://github.com/temporalio/samples-dotnet/issues/38

### DIFF
--- a/src/Temporalio/Converters/DefaultPayloadConverter.cs
+++ b/src/Temporalio/Converters/DefaultPayloadConverter.cs
@@ -56,7 +56,7 @@ namespace Temporalio.Converters
         /// instances, so only subclasses would call this.
         /// </remarks>
         /// <seealso cref="DefaultPayloadConverter()" />
-        protected DefaultPayloadConverter(JsonSerializerOptions jsonSerializerOptions)
+        public DefaultPayloadConverter(JsonSerializerOptions jsonSerializerOptions)
             : this(
                 new BinaryNullConverter(),
                 new BinaryPlainConverter(),
--- a/src/Temporalio/Converters/DefaultPayloadConverter.cs
+++ b/src/Temporalio/Converters/DefaultPayloadConverter.cs
@@ -51,10 +51,6 @@ namespace Temporalio.Converters
         /// Initializes a new instance of the <see cref="DefaultPayloadConverter"/> class.
         /// </summary>
         /// <param name="jsonSerializerOptions">Custom serializer options.</param>
-        /// <remarks>
-        /// This is protected because payload converters are referenced as class types, not
-        /// instances, so only subclasses would call this.
-        /// </remarks>
         /// <seealso cref="DefaultPayloadConverter()" />
         public DefaultPayloadConverter(JsonSerializerOptions jsonSerializerOptions)
             : this(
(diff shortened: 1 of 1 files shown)


src/Temporalio/Converters/DefaultPayloadConverter.cs
```diff
@@ -56,7 +56,7 @@ public DefaultPayloadConverter()
         /// instances, so only subclasses would call this.
         /// </remarks>
         /// <seealso cref="DefaultPayloadConverter()" />
-        protected DefaultPayloadConverter(JsonSerializerOptions jsonSerializerOptions)
+        public DefaultPayloadConverter(JsonSerializerOptions jsonSerializerOptions)
```
cretz, line 55: Can you make the one below this public too and remove the remarks paragraph from xmldoc on both that says why it is protected?

cretz (approved):
Looks great. I'm going to merge `main` into your branch and then will merge this into `main` upon CI complete. Thanks!

src/Temporalio/Converters/DefaultPayloadConverter.cs
```diff
@@ -87,7 +79,6 @@ protected DefaultPayloadConverter(params IEncodingConverter[] encodingConverters
 
         // KeyedCollection not worth it and OrderedDictionary not generic. So we'll expose the
         // collection and maintain an internal dictionary.
-
```
cretz: Looks like this line removal broke lint check, may need to add it back. Also note since I merged `main` into your branch, you'll want to `pull` before making this change.

src/Temporalio/Converters/DefaultPayloadConverter.cs
```diff
@@ -87,7 +79,6 @@ protected DefaultPayloadConverter(params IEncodingConverter[] encodingConverters
 
         // KeyedCollection not worth it and OrderedDictionary not generic. So we'll expose the
         // collection and maintain an internal dictionary.
```
ElanHasson, line 81:
```suggestion
        // collection and maintain an internal dictionary.

```

ElanHasson:
```suggestion
        // collection and maintain an internal dictionary.

```

src/Temporalio/Converters/DefaultPayloadConverter.cs
ElanHasson: @cretz mobile edits ftw :) 

Should be good for you to retrigger CI

PR #233, opened 2024-04-26
Support macOS ARM alongside macOS Intel and upgrade Protobuf library

### LINKED ISSUE #210: [Feature Request] Use macOS M1/ARM runners for all SDK CI
See: https://github.com/temporalio/features/issues/396

cretz:
## What was changed

* Add macos-intel and macos-arm both as separate CI targets
* Switch to building binaries and doing smoke test on macos-arm properly instead of cross compile
  * This was tested in "draft mode" of this PR, then changed back to its normal state of not running on every PR
* Update protobuf library and update `protoc` expectation and regenerate protobufs with the newer `protoc`

Branch protection will be updated upon approval

## Checklist

1. Closes #210


.github/workflows/ci.yml
```diff
@@ -38,23 +42,19 @@ jobs:
...
-      - name: Install protoc (x64)
-        # Does not work on arm
-        if: ${{ matrix.os != 'ubuntu-arm' }}
-        uses: arduino/setup-protoc@v1
+      - name: Install protoc
+        uses: arduino/setup-protoc@v3
         with:
-          # TODO(cretz): Upgrade when https://github.com/arduino/setup-protoc/issues/33 fixed
-          version: '3.x'
+          # TODO(cretz): Can upgrade proto when https://github.com/arduino/setup-protoc/issues/99 fixed
```
mjameswh, line 54: Not very important, but I'm a bit puzzled by this TODO. Looks like the issue is about the _default_ version of protoc that `setup-protoc` will install. But if you specify the version number explicitly, there should be no problem, no?

Quinn-With-Two-Ns (approved):

.github/workflows/ci.yml
cretz: > Looks like this is about the default version of protoc that setup-protoc will install.

We want to pin our protoc version (different versions can generate different code). This is just saying we can't upgrade past 23.x because of that issue (though may be indirectly because of it, but I tried "27.x" here and it failed)

mjameswh (approved):
LGTM

PR #236, opened 2024-05-03
Worker client replacement

### LINKED ISSUE #235: [Feature Request] Client replacement in worker

Need to be able to replace the worker client with a new client that may be connected with different options (e.g. a new TLS cert). See https://github.com/temporalio/features/issues/11.

cretz:
## What was changed

* Update core to get https://github.com/temporalio/sdk-core/pull/721
* Exposed `Client` property on `Worker` w/ getter/setter
  * Setter now replaces the worker client
* Added test proving the worker switches servers (even if this is not how people should use client replacement)

## Checklist

1. Closes #235

### DIFF
--- a/src/Temporalio/Worker/TemporalWorker.cs
+++ b/src/Temporalio/Worker/TemporalWorker.cs
@@ -15,9 +15,11 @@ namespace Temporalio.Worker
     /// </summary>
     public class TemporalWorker : IDisposable
     {
+        private readonly object clientLock = new();
         private readonly ActivityWorker? activityWorker;
         private readonly WorkflowWorker? workflowWorker;
         private readonly bool workflowTracingEventListenerEnabled;
+        private IWorkerClient client;
         private int started;
 
         /// <summary>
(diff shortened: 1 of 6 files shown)


src/Temporalio/Worker/TemporalWorker.cs
```diff
@@ -102,9 +104,39 @@ public TemporalWorker(IWorkerClient client, TemporalWorkerOptions options)
         public TemporalWorkerOptions Options { get; private init; }
 
         /// <summary>
-        /// Gets the client this worker was created with.
+        /// Gets or sets the client for this worker.
         /// </summary>
-        internal IWorkerClient Client { get; private init; }
+        /// <remarks>
+        /// When this property is set, it actually replaces the underlying client that is being used
+        /// by the worker. This means the next calls by the worker to Temporal (e.g. responding
```
Sushisource, line 111: nitpicky

```suggestion
        /// by the worker. This means subsequent calls by the worker to Temporal (e.g. responding
```

Sushisource (approved):

src/Temporalio/Worker/TemporalWorker.cs
cretz: :+1: This one in particular may not be worth another commit.

PR #239, opened 2024-05-08
Fix keyed service error

### LINKED ISSUE #238: [Bug] System.InvalidOperationException: This service descriptor is keyed. Your service provider may not support keyed services
### Describe the bug

See https://github.com/dotnet/runtime/issues/95789. 

As of the `v1.1.0` release,  the `System.InvalidOperationException: This service descriptor is keyed. Your service provider may not support keyed services` is thrown when calling `AddHostedTemporalWorker` after registering a keyed service descriptor to container.

I believe this to be the offending code, introduced in #216.

https://github.com/temporalio/sdk-dotnet/blob/ee378c17595bccf8a390ff82e4e1f514d4a6344f/src/Temporalio.Extensions.Hosting/TemporalWorkerServiceOptionsBuilderExtensions.cs#L184-L186

### Environment/Versions

- OS and processor: Windows/Linux
- Temporal Version: n/a -- can't even startup
- Are you using Docker or Kubernetes or building Temporal from source? n/a

ElanHasson:
# THIS NEEDS A DISCUSSION

I wanted to get something to share up and start a discussion as targeting.net 8 is probably discussion worthy 😅. I do not expect this to get merged as-is.

## What was changed

Fixes #238.
Also fixes a bunch of style issues, adds conditional compilation for targeting .net 8, oh and also retarget to .net 8.

See https://github.com/AzureAD/microsoft-identity-web/pull/2676 for prior art.
## Why?

Unable to use the SDK in .NET 8 when keyed services are registered first 
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -35,7 +35,12 @@
     <AssemblyVersion>1.1.0</AssemblyVersion>
   </PropertyGroup>
 
-  <ItemGroup>
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' != 'NET8_0_OR_GREATER'">
+    <PackageReference Include="Microsoft.VisualStudio.Threading.Analyzers" Version="17.4.33" PrivateAssets="all" />
+    <PackageReference Include="StyleCop.Analyzers" Version="1.2.0-beta.435" PrivateAssets="all" />
+  </ItemGroup>
+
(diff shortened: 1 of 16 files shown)


ElanHasson:
Fixed via https://github.com/temporalio/sdk-dotnet/pull/241

PR #242, opened 2024-05-08
Run wait-condition callbacks in workflow context

### LINKED ISSUE #240: [Bug] WaitConditionAsync is run outside of Workflow context
### Describe the bug

The callback for `WaitConditionAsync` is run outside of the event loop, so the current task scheduler is not the workflow instance. This means any calls on the `Workflow` object within the callback will fail.

cretz:
## What was changed

Changed logic to make the wait condition callbacks a task in so it is run within the workflow context so the user can access things on the `Workflow` static class

## Checklist

1. Closes #240

### DIFF
--- a/src/Temporalio/Worker/WorkflowInstance.cs
+++ b/src/Temporalio/Worker/WorkflowInstance.cs
@@ -648,38 +648,58 @@ namespace Temporalio.Worker
         private void RunOnce(bool checkConditions)
         {
             // Run as long as we have scheduled tasks
-            // TODO(cretz): Fix to run as long as any tasks not yielded on Temporal
             while (scheduledTasks.Count > 0)
             {
-                while (scheduledTasks.Count > 0)
-                {
-                    // Pop last
-                    var task = scheduledTasks.Last!.Value;
-                    scheduledTasks.RemoveLast();
-                    scheduledTaskNodes.Remove(task);
+                // Run all tasks until empty
+                RunAllTasks();
 
-                    // This should never return false
-                    if (!TryExecuteTask(task))
-                    {
-                        logger.LogWarning("Task unexpectedly was unable to execute");
-                    }
-                    if (currentActivationException != null)
-                    {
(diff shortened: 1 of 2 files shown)


src/Temporalio/Worker/WorkflowInstance.cs
cretz, line 1: So git diff makes these changes look a bit scarier than they are. We just moved the run-all-task loop into a separate method, then enqueue the condition checker task instead of invoking it inline (issuing a run-all-task again for that new task).

src/Temporalio/Worker/WorkflowInstance.cs
```diff
@@ -648,38 +648,58 @@ protected override bool TryDequeue(Task task)
...
+                // tasks again. This is a scheduled task instead of just run inline because it needs
+                // to be in the workflow context (i.e. current task scheduler) so the `Workflow`
+                // methods work properly.
+                if (checkConditions && conditions.Count > 0)
                 {
-                    // Pop last
-                    var task = scheduledTasks.Last!.Value;
-                    scheduledTasks.RemoveLast();
-                    scheduledTaskNodes.Remove(task);
+                    _ = QueueNewTaskAsync(CheckConditionsAsync);
```
Quinn-With-Two-Ns, line 662: For my own understanding in dotnet, Do we detect if the user tries to call any illegal APIs in the wait condition?

Quinn-With-Two-Ns (approved):

src/Temporalio/Worker/WorkflowInstance.cs
cretz: We don't and we probably should. We have been mostly counting on the fact that it's not async to prevent commands, but there are commands like upsert search attribute that are not async. This is actually a gap for queries and update validators too. I have opened #243.

Tyler-Harker (approved):

cretz:
This caused a bug. Now conditions in coroutines depending on each other may not move forward because conditions are not re-run upon task resumption.

PR #251, opened 2024-05-21
Required wait update stage and polling improvements

### LINKED ISSUE #198: [Feature Request] SDK clients should set a UpdateID on any update request even if the user did not specify one
See: https://github.com/temporalio/features/issues/431

### LINKED ISSUE #199: [Feature Request] SDK should not return an update handle if the update has not reached the desired state
See https://github.com/temporalio/features/issues/432

### LINKED ISSUE #231: [Feature Request] Make StartUpdateAsync users aware that it's synchronous w/ worker.
See https://github.com/temporalio/features/issues/469

https://github.com/temporalio/sdk-dotnet/blob/2bd9ed3894adb290f78f4a4409797dc5d4ec82a0/src/Temporalio/Client/WorkflowHandle.cs#L313
wait stage shouldn't be optional here, and docs should clarify availability.

cretz:
## What was changed

* Created new `WorkflowUpdateStartOptions` as an extension of `WorkflowUpdateOptions` that has a required `WaitForStage`
* Created new `WorkflowUpdateStage` enum representing its proto equivalent
* Changed start update calls to _require_ `WorkflowUpdateStartOptions` (:boom: breaking change)
* Disallow `None` or `Admitted` wait stage on start, user must specify one
* Updated start to continually start until stage reached or stage accepted, then poll if needed to reach complete
* Added test to confirm workflow and update completed in same task

## Checklist

1. Closes #198
1. Closes #199
1. Closes #231

### DIFF
--- a/src/Temporalio/Client/TemporalClient.Workflow.cs
+++ b/src/Temporalio/Client/TemporalClient.Workflow.cs
@@ -229,6 +229,15 @@ namespace Temporalio.Client
             public async override Task<WorkflowUpdateHandle<TResult>> StartWorkflowUpdateAsync<TResult>(
                 StartWorkflowUpdateInput input)
             {
+                if (input.Options.WaitForStage == WorkflowUpdateStage.None)
+                {
+                    throw new ArgumentException("WaitForStage is required to start workflow update");
+                }
+                else if (input.Options.WaitForStage == WorkflowUpdateStage.Admitted)
+                {
+                    throw new ArgumentException(
+                        "Admitted is not an allowed wait stage to start workflow update");
+                }
                 // Build request
                 var req = new UpdateWorkflowExecutionRequest()
                 {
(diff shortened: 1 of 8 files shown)


src/Temporalio/Client/TemporalClient.Workflow.cs
```diff
@@ -229,6 +229,15 @@ public override async Task<TResult> QueryWorkflowAsync<TResult>(QueryWorkflowInp
             public async override Task<WorkflowUpdateHandle<TResult>> StartWorkflowUpdateAsync<TResult>(
                 StartWorkflowUpdateInput input)
             {
+                if (input.Options.WaitForStage == WorkflowUpdateStage.None)
+                {
+                    throw new ArgumentException("WaitForStage is required to start workflow update");
+                }
+                else if (input.Options.WaitForStage == WorkflowUpdateStage.Admitted)
```
Quinn-With-Two-Ns, line 236: Could just allow the server to reject this call? That way you don't need to update the SDK if the server adds support .

cretz: Same thing was mentioned at https://github.com/temporalio/sdk-python/pull/521#discussion_r1607125013. I think we need to document what works and what doesn't here anyways, so that'll have to be updated anyways, and hopefully this becomes a default parameter in some durable-update future.

src/Temporalio/Client/WorkflowUpdateOptions.cs
```diff
@@ -1,30 +1,37 @@
...
-        /// Gets or sets RPC options for starting the workflow.
+        /// Initializes a new instance of the <see cref="WorkflowUpdateOptions"/> class.
         /// </summary>
-        public RpcOptions? Rpc { get; set; }
+        /// <param name="id">Update ID.</param>
+        public WorkflowUpdateOptions(string id) => Id = id;
 
         /// <summary>
-        /// Gets or sets the stage to wait for on start. Internal only.
+        /// Gets or sets the unique workflow identifier. If not set, this is defaulted to a GUID.
```
Sushisource:
```suggestion
        /// Gets or sets the unique update identifier. If not set, this is defaulted to a GUID.
```

src/Temporalio/Client/WorkflowUpdateStage.cs
```diff
@@ -0,0 +1,34 @@
...
+
+        /// <summary>
+        /// Admitted stage. This stage is reached when the server receives the update to process.
+        /// This is currently an invalid value on start.
+        /// </summary>
+        Admitted = UpdateWorkflowExecutionLifecycleStage.Admitted,
+
+        /// <summary>
+        /// Accepted stage. This stage is reached when a workflow has received the update and either
+        /// accepted or rejected it.
```
Sushisource:
```suggestion
        /// accepted (it has passed validation) or rejected it.
```

Sushisource (approved):

src/Temporalio/Client/WorkflowUpdateStage.cs
cretz: Fixed

src/Temporalio/Client/WorkflowUpdateOptions.cs
cretz: Fixed

PR #252, opened 2024-05-22
Improved docs for RetryPolicy.NonRetryableErrorTypes

petrkoutnycz:
## What was changed
Extended XML docs for RetryPolicy.NonRetryableErrorTypes property.

## Why?
Wanted to make it obvious that values from NonRetryableErrorTypes property are matched against ApplicationFailureException.ErrorType property.

### DIFF
--- a/src/Temporalio/Common/RetryPolicy.cs
+++ b/src/Temporalio/Common/RetryPolicy.cs
@@ -1,7 +1,6 @@
 using System;
 using System.Collections.Generic;
 using Google.Protobuf.WellKnownTypes;
-using Temporalio.Exceptions;
 
 namespace Temporalio.Common
 {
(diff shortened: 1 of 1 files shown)


src/Temporalio/Common/RetryPolicy.cs
```diff
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using Google.Protobuf.WellKnownTypes;
+using Temporalio.Exceptions;
```
cretz: Can you qualify the name in the docs instead of add this `using` that is only referenced in docs?

cretz (commented):
Looks good, one minor suggestion

src/Temporalio/Common/RetryPolicy.cs
petrkoutnycz: Sure.

cretz (approved):

cretz:
Thanks! Will merge if/when CI passes.

PR #255, opened 2024-05-29
Extended reporting of failed workflow with OTel

petrkoutnycz:
## What was changed
Marked Activity for workflow completion as erroneous.

## Why?
So a developer might see the activity to be failed in observability tool, recording event with exception is not enough. At least not in our case with New Relic.


src/Temporalio.Extensions.OpenTelemetry/ActivityExtensions.cs
cretz, line 1: Instead of a separate public extension, can you make a private static helper in `TracingInterceptor` that you pass the `Activity` and `Exception` to?

cretz (approved):
Thanks!

Issue #257, opened 2024-05-31
[Feature Request] Support worker client update when using generic host


cretz:
### Describe the solution you'd like

When using generic host support for workers people still want to update the client on the worker. We could use `IOptionsMonitor`, but the `IOptionsMonitor` we use is named with an internal unique name we'd have to expose and it only has connection options, not the client itself and we want to force users to create the client itself. A better pattern would be a `ClientUpdater` which the user can instantiate (or use from DI) and set on the worker service options. They can then use that to push a client down to all workers (so internally workers would subscribe to client update events on the updater). Make sure to test and document manual instantiation _and_ DI-based instantiation of this updater.

cretz:
This is now supported in 1.1.2 with #258.

PR #258, opened 2024-05-31
feat: add workflow client updater for updating workflow client

robcao:
## What was changed

Added an optional `WorkerClientUpdater` parameter to `TemporalWorkerOptions`. If specified, the worker will subscribe to worker client changes from the updater, and update its underlying worker client.

## Why?

This allows users to update the underlying `IWorkerClient` on running workers without having to restart the workers entirely when using the DI extension.

## Checklist

1. Closes 
https://github.com/temporalio/sdk-dotnet/issues/257

2. How was this tested:

Added test cases both for manual instantiation and DI based instantiation.

3. Any docs updates needed?

I updated the README of the hosting extension. I didn't update the README of the regular library as worker refresh is not documented there, and I don't really think there is less of a reason to use the worker client updater if not using DI.


tests/Temporalio.Tests/Worker/WorkflowWorkerTests.cs
```diff
@@ -4449,7 +4448,58 @@ await foreach (var evt in handle2.FetchHistoryEventsAsync())
             worker.Client = otherEnv.Client;
 
             // Now confirm the other workflow has started
-            await AssertHasEventEventuallyAsync(handle1, e => e.WorkflowTaskCompletedEventAttributes != null);
+            await handle2.AssertHasEventEventuallyAsync(e => e.WorkflowTaskCompletedEventAttributes != null);
```
robcao: I think this was a typo from the original PR? I changed the test case to check `handle2` instead of `handle1`.

tests/Temporalio.Tests/WorkerAssertionExtensions.cs
```diff
@@ -0,0 +1,64 @@
+﻿using Temporalio.Api.History.V1;
+using Temporalio.Client;
+using Xunit;
+
+namespace Temporalio.Tests
+{
+    public static class WorkerAssertionExtensions
```
robcao: I wanted to re-use some of these methods that were originally in `WorkflowWorkerTests.cs` in `TemporalWorkerServiceTests.cs`, so I pulled them out into extension methods here.

src/Temporalio/Worker/IWorkerClientUpdater.cs
cretz, line 1: I don't think any of this is needed in the `Temporalio` project. This is a generic host problem only. Those making workers directly can just use the `Client` setter. I also don't think an interface is needed unless you really think it's important for testing (even then not sure the event handler needs to be part of the interface). Also, even though it's a mouthful, I'd call this `TemporalWorkerClientUpdater` w/ the `Temporal` prefix to match the worker class name for discoverability.

tests/Temporalio.Tests/WorkerAssertionExtensions.cs
cretz: Can you instead add them to the `AssertMore` class that already exists (and remove the `Assert` prefix of the methods and not make them extension methods)?

cretz (commented):
Thanks for the PR! My original thought was that this was best in the hosting extension only since those creating workers directly don't need this abstraction.

tests/Temporalio.Tests/Worker/WorkflowWorkerTests.cs
cretz: Yes, thanks! (hopefully my typo here didn't cause the test to pass when it shouldn't have)

tests/Temporalio.Tests/WorkerAssertionExtensions.cs
robcao: Done

src/Temporalio/Worker/IWorkerClientUpdater.cs
robcao: Makes sense, done.

I don't think a lack of an interface hampers testing, since the implementation has no dependencies

tests/Temporalio.Tests/Worker/WorkflowWorkerTests.cs
robcao: nope, tests passed fine!

src/Temporalio.Extensions.Hosting/TemporalWorkerClientUpdateSubscriber.cs
```diff
@@ -0,0 +1,68 @@
+﻿using System;
+using Temporalio.Worker;
+
+namespace Temporalio.Extensions.Hosting
+{
+    /// <summary>
+    /// Subscribes to the <see cref="TemporalWorkerClientUpdater"/>, and propagates the worker client changes to the Temporal worker.
+    /// </summary>
+    public class TemporalWorkerClientUpdateSubscriber : IDisposable
```
cretz: Arguably does not need to be public and arguably not needed (I think a try/finally in the worker is enough, though would recommend the disposable approach if this were public API). Can probably just remove this and let the worker service subscribe to events itself without this layer of indirection.

src/Temporalio.Extensions.Hosting/TemporalWorkerClientUpdater.cs
```diff
@@ -0,0 +1,48 @@
...
+    /// Notification hub that can be used to push Temporal worker client updates to subscribing Temporal workers.
+    /// </summary>
+    public class TemporalWorkerClientUpdater
+    {
+        private readonly object clientLock = new();
+
+        /// <summary>
+        /// The <see cref="EventHandler"/> used to dispatch notifications to subscribers that the Temporal worker client was updated.
+        /// </summary>
+        public event EventHandler<TemporalWorkerClientUpdatedEventArgs> TemporalWorkerClientUpdated
```
cretz: I think this can just have an `internal` `Subscribe` and `Unsubscribe` and not need this public property (and therefore not need `EventArgs` even though that's best practice for public API; can disable analyzers for that if needed). Or you can just make the `private` field be `internal` and disable those static analyzer checks (but still, don't need `EventArgs` wrapper for internal-only handler IMO).

src/Temporalio.Extensions.Hosting/TemporalWorkerClientUpdater.cs
```diff
@@ -0,0 +1,48 @@
...
+            }
+        }
+
+        private event EventHandler<TemporalWorkerClientUpdatedEventArgs>? TemporalWorkerClientUpdatedEvent;
+
+        /// <summary>
+        /// Dispatches a notification to all subscribers that a new worker client should be used.
+        /// </summary>
+        /// <param name="client">The new <see cref="IWorkerClient"/> that should be pushed out to all subscribing workers.</param>
+        public void UpdateTemporalWorkerClient(IWorkerClient client)
```
cretz:
```suggestion
        public void UpdateClient(IWorkerClient client)
```

No need to qualify when already in the class

src/Temporalio.Extensions.Hosting/TemporalWorkerServiceOptions.cs
```diff
@@ -39,6 +45,12 @@ public override object Clone()
             {
                 options.ClientOptions = (TemporalClientConnectOptions)ClientOptions.Clone();
             }
+
+            if (WorkerClientUpdater != null)
+            {
+                options.WorkerClientUpdater = WorkerClientUpdater;
+            }
+
```
cretz:
```suggestion
```

Clone will handle this fine

src/Temporalio.Extensions.Hosting/TemporalWorkerService.cs
```diff
@@ -19,6 +19,7 @@ public class TemporalWorkerService : BackgroundService
         private readonly TemporalClientConnectOptions? newClientOptions;
         private readonly ITemporalClient? existingClient;
         private readonly TemporalWorkerOptions workerOptions;
+        private readonly TemporalWorkerClientUpdater? workerClientUpdater;
```
cretz, line 22:
```suggestion
        private readonly ITemporalClient? existingClient;
        private readonly TemporalWorkerServiceOptions workerOptions;
```

Now that we are basically storing the three separate pieces anyways, let's just store the full service options as one field

tests/Temporalio.Tests/Extensions/Hosting/TemporalWorkerServiceTests.cs
```diff
@@ -1,3 +1,5 @@
+using Temporalio.Tests.Worker;
```
cretz: Don't usually have `using` above the namespace here and would prefer to just copy that simple workflow into here instead of have test files depend on each other.

cretz (commented):
Thanks! Some more comments about visibility and general simplification I think we can do

src/Temporalio.Extensions.Hosting/TemporalWorkerService.cs
robcao: There's not really a good way I've found to handle the constructors that take in `TemporalWorkerOptions` directly (from some testing, cloning is not contravariant).

I don't know of a better way to convert an instance of `TemporalWorkerOptions` into `TemporalWorkerServiceOptions` other than writing a direct translation method, which imo is not desirable for maintenance. Do you have a suggestion?

I think it would be easier if `TemporalWorkerServiceOptions` was composed of `TemporalWorkerOptions` instead of directly inheriting from it, but that is a breaking change.

cretz: Ah, I see now that we have multiple constructors for this `TemporalWorkerService` that accept partial forms. Ok, can ignore my comment and leave these as separate fields.

src/Temporalio.Extensions.Hosting/TemporalWorkerClientUpdater.cs
```diff
@@ -0,0 +1,48 @@
...
+        /// <summary>
+        /// Adds a new subscriber that will be notified when a new worker client should be used.
+        /// </summary>
+        /// <param name="eventHandler">The event handler to add to the event listeners.</param>
+        internal void Subscribe(EventHandler<IWorkerClient> eventHandler)
+        {
+            lock (clientLock)
+            {
+                OnClientUpdatedEvent += eventHandler;
+            }
```
cretz, line 33: Looks like locking technically not required per https://stackoverflow.com/questions/22825117/is-it-thread-safe-to-register-for-a-c-sharp-event, but this is fine

cretz (approved):
Looks great! Will merge a bit later today.

PR #259, opened 2024-06-03
Fix wait condition evaluation issue

cretz:
## What was changed

This fixes/reverts an issue caused by #242. Specifically that issue scheduled condition checks as a task and just ran all tasks to completion. The problem is that we don't re-schedule the condition checks after the tasks that the conditions caused to wake up are completed. So we have effectively reverted that issue to return to evaluating conditions inline, but we added a feature to allow `Workflow` access from inside the conditionals.

💥 **WARNING** This is technically a _history_ backwards incompatible change for anyone that has used 1.1.1 and has relied on wait conditions to be woken up after other coroutines set values. After discussion it was determined that short life and limited use of of the 1.1.1 release allows us to fix this 3-week bug. But this can technically cause non-determinism in those upgrading from 1.1.1 in that it can make a condition properly be re-evaluated a second time in a task where it wasn't before. Users before 1.1.1 are not affected.

### DIFF
--- a/src/Temporalio/Worker/WorkflowInstance.cs
+++ b/src/Temporalio/Worker/WorkflowInstance.cs
@@ -653,19 +653,22 @@ namespace Temporalio.Worker
                 // Run all tasks until empty
                 RunAllTasks();
 
-                // If there are any conditions, schedule a new condition-check task and the run all
-                // tasks again. This is a scheduled task instead of just run inline because it needs
-                // to be in the workflow context (i.e. current task scheduler) so the `Workflow`
-                // methods work properly.
+                // If there are any conditions, run the condition-check as a task. This is a task
+                // instead of just run inline because it needs to be in the workflow context (i.e.
+                // current task scheduler) so the `Workflow` methods work properly. However, we make
+                // sure that we only run this task because we want the loop to continue with other
+                // scheduled tasks that the conditions caused to wake up. An original, naive form of
+                // this ran all tasks including conditions, but that didn't allow conditions that
+                // depended on each other to properly re-schedule earlier conditions.
                 if (checkConditions && conditions.Count > 0)
                 {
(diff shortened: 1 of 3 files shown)


cretz:
This has become a bit more challenging on forever-waiting conditions, moving to draft while I work it out...

cretz:
Updated to basically revert #242, please review

src/Temporalio/Worker/WorkflowInstance.cs
```diff
@@ -651,55 +651,48 @@ private void RunOnce(bool checkConditions)
...
 
-        private Task CheckConditionsAsync()
-        {
-            try
-            {
-                foreach (var source in conditions.Where(t => t.Item1()).Select(t => t.Item2))
+                // Check conditions. It would be nice if we could run this in the task scheduler
+                // because then users could have access to the `Workflow` context in the condition
+                // callback. However, this cannot be done because even just running one task in the
+                // scheduler causes .NET to add more tasks to the scheduler. And you don't want to
```
Quinn-With-Two-Ns, line 675: What other tasks is dotnet adding ?

cretz: Just looked at stack trace, it seems like this is actually in my stack trace logic at https://github.com/temporalio/sdk-dotnet/blob/1f6676bb9563ddb8eb4606a8df650fb540420fc2/src/Temporalio/Worker/WorkflowInstance.cs#L730 (and just below it for the return-type-based one) which is occurring during stack trace test (stack trace disabled by default in .NET). So I could technically disable stack trace continuations during condition checking, but I think the behavior here in the PR of not pushing the condition checks into the task queue is the best anyways.

Quinn-With-Two-Ns (approved):

PR #260, opened 2024-06-04
fix: update generated rpc requests to not omit ConfigureAwait(false)

robcao:
## What was changed

Updates the code generated rpc calls to just return the `Task<Resp>` instead of awaiting.

## Why?

The current code is being awaited without a `ConfigureAwait(false)`, which can cause deadlocks when being invoked inside synchronization contexts such as `WindowsFormsSynchronizationContext`.

## Checklist

1. Closes 
No issue created, discussed with Chad on Slack, could create an issue if desired.

2. How was this tested:

I ran the existing tests locally and made sure they passed.

I could add a windows-only test for this if wanted, but this is very edge case and there are a few msbuild property changes needed to get WinForms into .net6 that would have to be windows-only.

Or I guess I could write a custom SynchronizationContext that does essentially the same thing, and use it for a test?

The deadlock can be simulated by running this console application on a Windows machine:

```
using System;
using System.Threading;
using System.Threading.Tasks;
using System.Windows.Forms;

namespace WinFormsRepro
{
	public class Program
	{
		public static async Task Main(string[] args)
		{
			SynchronizationContext.SetSynchronizationContext(new WindowsFormsSynchronizationContext());

			Console.Out.WriteLine("About to call the workflow service!");

			await SimulateWorkflowServiceRpcAsync().ConfigureAwait(false);

			Console.Out.WriteLine("Finished calling the workflow service.");
		}


		public static async Task SimulateWorkflowServiceRpcAsync() => await SimulateTemporalRpcCallAsync();

		public static async Task SimulateTemporalRpcCallAsync()
		{
			var completion = new TaskCompletionSource<string>();

			Task task = Task.Run(() =>
			{
				Thread.Sleep(100);
				completion.SetResult("complete");
			});

			await task.ConfigureAwait(false);

		}
	}

}

```

with the following .csproj

```
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net6.0-windows</TargetFramework>
    <UseWindowsForms>true</UseWindowsForms>
  </PropertyGroup>

</Project>

```

I tested that changing `SimulateWorkflowServiceRpcAsync` to the following does not deadlock.

```
public static Task SimulateWorkflowServiceRpcAsync() => SimulateTemporalRpcCallAsync();
```

3. Any docs updates needed?

No

### DIFF
--- a/src/Temporalio.Api.Generator/Program.cs
+++ b/src/Temporalio.Api.Generator/Program.cs
@@ -214,10 +214,10 @@ static string GenerateServiceRPCSource(
         code += "\n        /// <param name=\"options\">Optional RPC options.</param>";
         code += "\n        /// <returns>RPC response</returns>";
         code +=
-            $"\n        public async Task<{resp}> {call}Async({req} req, RpcOptions? options = null)\n";
+            $"\n        public Task<{resp}> {call}Async({req} req, RpcOptions? options = null)\n";
         code += "        {\n";
         code +=
-            $"            return await InvokeRpcAsync(\"{call}\", req, {resp}.Parser, options);\n";
+            $"            return InvokeRpcAsync(\"{call}\", req, {resp}.Parser, options);\n";
         code += "        }\n";
     }
     code += "    }\n}";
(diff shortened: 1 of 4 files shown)


cretz (approved):
Looks great, will merge if/when CI passes (I am merging main into your branch first).

PR #269, opened 2024-06-07
Add strongly-typed extension methods for SignalWithStart

Blackburn29:
## What was changed

Added extension methods for calling `signalWithStart`.
## Why?

Adds support for strong typings when starting a workflow with a signal.

Currently this requires holding a second variable for `WorkflowOptions` and calling
`options.SignalWithStart(x => x.Method())`.

These extension methods give you the ability to short-hand the setup:
```csharp
var handle = await _temporalClient.SignalWorkflowWithStartAsync(
    workflow => workflow.RunAsync(),
    workflow => workflow.SignalMethod(),
    new WorkflowOptions
    {
       //...
    });
```
## Checklist

1. Closes 
N/A

2. How was this tested:
I have been using these extension methods in some of my applications for awhile, but I figured they would be useful to others.

3. Any docs updates needed?
N/A

### DIFF
--- a/src/Temporalio/Client/ITemporalClientExtensions.cs
+++ b/src/Temporalio/Client/ITemporalClientExtensions.cs
@@ -44,6 +44,49 @@ namespace Temporalio.Client
         }
 
         /// <summary>
+        /// Signal and (optionally) start a workflow via lambda invoking the signalWithStart method.
+        /// </summary>
+        /// <typeparam name="TWorkflow">Workflow class type.</typeparam>
+        /// <param name="client">Client to use.</param>
+        /// <param name="workflowRunCall">Invocation of workflow run method with a result.</param>
+        /// <param name="signalCall">Invocation of signal method without a result.</param>
+        /// <param name="options">Start workflow options. ID and TaskQueue are required.</param>
+        /// <returns>Workflow handle for the started workflow.</returns>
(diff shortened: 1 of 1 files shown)


cretz:
We may have future things like update with start and so we decided in newer SDKs that extra things on start are just options, so I am not sure we want to provide top-level shortcuts to save creating options in this case, but we of course support users doing so like you've done here.

As potential new multi-operation RPC calls come about some of this design may change. But if they don't change this much, we can merge this shortcut. But I think we want to wait on it to see those designs.

jmaeagle99:
Thank you for this contribution, and apologies for how long it's sat open. Circling back on the earlier note: we wanted to hold this until the "update with start" designs landed, since those could have reshaped how we approach multi-operation start calls. Those designs have since shipped, and they point towards keeping signal-with-start as it is.

The convention that emerged is that each of these APIs takes exactly one top-level operation lambda, with the ride-along operation expressed on the options object rather than as a second lambda parameter. The returned handle corresponds with the top-level lambda expression.

Rather than add a shortcut that departs from that convention, we're going to keep signal-with-start expressed through options and close this out. That said, and as you referenced, the pattern you’ve built is a perfectly good one for your own codebase and is fully supported.

Really appreciate you taking time to share it, and sorry again for the delay in giving you a clear answer.

PR #270, opened 2024-06-10
Workflow completion command reordering

### LINKED ISSUE #249: [Feature Request] Do not set workflow completion until after all coroutines have settled in the task

See https://github.com/temporalio/features/issues/481

cretz:
## What was changed

See https://github.com/temporalio/features/issues/481. Like other core-based SDKs, .NET put the workflow complete command on the command list immediately when it occurred from workflow return. And then it truncated anything else after that. So coroutines that completed in the same task but after workflow method return were dropped.

This issue changes that behavior to enure same-task-post-complete coroutine commands are preserved. The approach that is taken now is that, if we are not replaying (or replaying w/ SDK flag set) and there are commands after return, move the return to the end and set an SDK flag. Otherwise do as done today. This not only ensures that workflow completion comes last, but the checking for whether any post-complete commands even exist allow us to not set this flag.

(there is also a somewhat unrelated fix to a flaky user-client-replacement test)

## Checklist

1. Closes #249

### DIFF
--- a/src/Temporalio/Worker/WorkflowInstance.cs
+++ b/src/Temporalio/Worker/WorkflowInstance.cs
@@ -67,6 +67,7 @@ namespace Temporalio.Worker
         private readonly Action<WorkflowInstance> onTaskStarting;
         private readonly Action<WorkflowInstance, Exception?> onTaskCompleted;
         private readonly IReadOnlyCollection<Type>? workerLevelFailureExceptionTypes;
+        private readonly bool disableCompletionCommandReordering;
         private WorkflowActivationCompletion? completion;
         // Will be set to null after last use (i.e. when workflow actually started)
         private Lazy<object?[]>? startArgs;
@@ -187,6 +188,7 @@ namespace Temporalio.Worker
             Random = new(details.Start.RandomnessSeed);
             TracingEventsEnabled = !details.DisableTracingEvents;
             workerLevelFailureExceptionTypes = details.WorkerLevelFailureExceptionTypes;
+            disableCompletionCommandReordering = details.DisableCompletionCommandReordering;
         }
 
         /// <summary>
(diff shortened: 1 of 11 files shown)


src/Temporalio/Worker/WorkflowInstance.cs
```diff
@@ -1402,6 +1385,71 @@ private string GetStackTrace()
...
+            }
+            if (IsReplaying && !act.AvailableInternalFlags.Contains((uint)WorkflowLogicFlag.ReorderWorkflowCompletion))
+            {
+                return;
+            }
+
+            // We know we're on a newer SDK and can move completion to the end if we need to. First,
+            // find the completion command and the number of trailing query commands.
+            var completionCommandIndex = -1;
+            var trailingQueryCommandCount = 0;
```
cretz: Yes, I know technically I could have just moved past trailing query commands (and now queries come in their own activation IIRC), but in my head it only makes sense to move until before those. But I can simplify the logic by just always moving to the end if it others think it's worth it.

src/Temporalio/Worker/WorkflowInstance.cs
```diff
@@ -1402,6 +1385,71 @@ private string GetStackTrace()
             }).Where(s => !string.IsNullOrEmpty(s)).Select(s => $"Task waiting at:\n{s}"));
         }
 
+        private void ApplyCompletionCommandReordering(
```
cretz, line 1388: Looking for review of this method

src/Temporalio/Worker/WorkflowInstance.cs
```diff
@@ -1402,6 +1385,71 @@ private string GetStackTrace()
...
+            // We know we're on a newer SDK and can move completion to the end if we need to. First,
+            // find the completion command and the number of trailing query commands.
+            var completionCommandIndex = -1;
+            var trailingQueryCommandCount = 0;
+            for (var i = 0; i < completion.Successful.Commands.Count; i++)
+            {
+                var cmd = completion.Successful.Commands[i];
+                // Set completion index if the command is a completion
+                if (cmd.CompleteWorkflowExecution != null ||
+                    cmd.ContinueAsNewWorkflowExecution != null ||
```
Quinn-With-Two-Ns, line 1417: What  about cancelled?

src/Temporalio/Worker/WorkflowInstance.cs
Sushisource: Yeah I just wouldn't bother with this

src/Temporalio/Worker/WorkflowInstance.cs
Sushisource: Yep you'll definitely want that here.

src/Temporalio/Worker/WorkflowInstance.cs
cretz: Concur, will simplify

src/Temporalio/Worker/WorkflowInstance.cs
cretz: Ah thanks! The logic I took this from didn't account for cancelled either (but core takes off post-complete commands anyways if we don't in lang, so no big deal). Fixing.

src/Temporalio/Worker/WorkflowInstance.cs
cretz: Fixed, made logic a tad simpler

src/Temporalio/Worker/WorkflowInstance.cs
cretz: Fixed

Sushisource (approved):

Quinn-With-Two-Ns (approved):

cretz:
Thanks for looking at this! Not sure anyone else is, so I am merging.

PR #275, opened 2024-06-17
Download cURL in older Linux containers on binary build

cretz:
## What was changed

Newer GH actions fail on older Linux containers due to older cURL so we update with a statically linked one (tested while this PR was in draft)


Sushisource (approved):
lol

mjameswh (approved):
LGTM

PR #283, opened 2024-06-20
Add workflow update RPC timeout/canceled exception

### LINKED ISSUE #250: Wrap GRPC::CANCELED and DEADLINE_EXCEEDED in an SDK Timeout exception for Update
So that users only have to catch one predictable exception

cretz:
## What was changed

Added `WorkflowUpdateRpcTimeoutOrCanceledException` and ensured that it was thrown on update RPC calls that are cancelled or timeout

## Checklist

1. Closes #250

### DIFF
--- a/src/Temporalio/Client/TemporalClient.Workflow.cs
+++ b/src/Temporalio/Client/TemporalClient.Workflow.cs
@@ -289,8 +289,18 @@ namespace Temporalio.Client
                 UpdateWorkflowExecutionResponse resp;
                 do
                 {
-                    resp = await Client.Connection.WorkflowService.UpdateWorkflowExecutionAsync(
-                        req, DefaultRetryOptions(input.Options.Rpc)).ConfigureAwait(false);
+                    try
+                    {
+                        resp = await Client.Connection.WorkflowService.UpdateWorkflowExecutionAsync(
+                            req, DefaultRetryOptions(input.Options.Rpc)).ConfigureAwait(false);
+                    }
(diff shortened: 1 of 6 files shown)


src/Temporalio/Client/TemporalClient.Workflow.cs
```diff
@@ -289,8 +289,18 @@ public override async Task<TResult> QueryWorkflowAsync<TResult>(QueryWorkflowInp
...
+                    {
+                        resp = await Client.Connection.WorkflowService.UpdateWorkflowExecutionAsync(
+                            req, DefaultRetryOptions(input.Options.Rpc)).ConfigureAwait(false);
+                    }
+                    catch (Exception e) when (e is OperationCanceledException ||
+                        (e is RpcException rpcErr && (
+                            rpcErr.Code == RpcException.StatusCode.DeadlineExceeded ||
+                            rpcErr.Code == RpcException.StatusCode.Cancelled)))
+                    {
+                        throw new WorkflowUpdateRpcTimeoutOrCanceledException(e);
```
Sushisource, line 302: Do we not want to keep retrying the start in these cases? Or, I guess right now that doesn't matter since there's no explicit overall "update timeout", only the rpc timeout?

cretz: No, these are usually caused by user-provided timeout or user-initiated cancel. Right, there is no update timeout yet which would be an application level concern and would be considered an update failure as opposed to an RPC failure.

Sushisource (approved):

PR #284, opened 2024-06-21
Fix improper casts during clone

### LINKED ISSUE #274: Wrong casting in several classes
Within the class ScheduleOptions method Clone() casting is done via the class WorkflowOptions, and I think maybe ScheduleOptions should be used:
https://github.com/temporalio/sdk-dotnet/blob/d3f38172112ddf0c855ad1ab9a9878432eb5dc4b/src/Temporalio/Client/Schedules/ScheduleOptions.cs#L45

Maybe I misunderstood something, but there are also a number of strange castings in these other classes:
https://github.com/temporalio/sdk-dotnet/blob/d3f38172112ddf0c855ad1ab9a9878432eb5dc4b/src/Temporalio/Client/Schedules/ScheduleTriggerOptions.cs#L26

https://github.com/temporalio/sdk-dotnet/blob/d3f38172112ddf0c855ad1ab9a9878432eb5dc4b/src/Temporalio/Client/WorkflowHistoryFetchOptions.cs#L34

https://github.com/temporalio/sdk-dotnet/blob/d3f38172112ddf0c855ad1ab9a9878432eb5dc4b/src/Temporalio/Client/Schedules/ScheduleListOptions.cs#L22

https://github.com/temporalio/sdk-dotnet/blob/d3f38172112ddf0c855ad1ab9a9878432eb5dc4b/src/Temporalio/Client/WorkflowCancelOptions.cs#L21

cretz:
## What was changed

Fix bad casting during clone and write test that confirms clone works with all of our hand-written options classes

## Checklist

1. Closes #274

### DIFF
--- a/src/Temporalio/Client/Schedules/ScheduleListOptions.cs
+++ b/src/Temporalio/Client/Schedules/ScheduleListOptions.cs
@@ -19,7 +19,7 @@ namespace Temporalio.Client.Schedules
         /// <returns>A shallow copy of these options and any transitive options fields.</returns>
         public virtual object Clone()
         {
-            var copy = (WorkflowListOptions)MemberwiseClone();
+            var copy = (ScheduleListOptions)MemberwiseClone();
             if (Rpc != null)
             {
                 copy.Rpc = (RpcOptions)Rpc.Clone();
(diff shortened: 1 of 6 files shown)


Sushisource (approved):
Heh, that's a pretty fun bit of reflection

cretz:
> Heh, that's a pretty fun bit of reflection

And one that many runtimes can't do. Getting all classes on all JARs in Java is rough and JVM/Python/Node/etc are often based on lazy loading so some things may not be there. But .NET assemblies give you this information up front which is nice.

Issue #281, opened 2024-06-18
Workflow-friendly concurrency control


drewhoskins-temporal:
We need a mechanism for concurrency control (mutex/semaphore) that fits with the language but is reentrant and deterministic within workflow coroutines.  To be used within signal and update handlers and the main workflow coroutine.

cretz:
Was solved in #287

PR #285, opened 2024-06-24
Confirmation that SemaphoreSlim can be used in workflows

cretz:
## What was changed

Added test for `SemaphoreSlim` which shows what works and what doesn't

## Checklist

1. Relates to #281 (will be closed via a sample separately)

### DIFF
--- a/tests/Temporalio.Tests/Worker/WorkflowWorkerTests.cs
+++ b/tests/Temporalio.Tests/Worker/WorkflowWorkerTests.cs
@@ -4,6 +4,7 @@
 namespace Temporalio.Tests.Worker;
 
 using System.Collections.Concurrent;
+using System.Linq.Expressions;
 using System.Threading.Tasks.Dataflow;
 using Microsoft.Extensions.Logging;
 using Temporalio.Activities;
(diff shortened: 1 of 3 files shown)


Sushisource (approved):

cretz:
This is annoying to use because you have to use only `WaitAsync` and not with timeout and probably not even with cancellation token. And it is disposable which is confusing for workflows to implement. So while we will still keep these tests to confirm behavior, I think we're going to write our own semaphore impl.

tests/Temporalio.Tests/Worker/WorkflowWorkerTests.cs
```diff
@@ -4789,6 +4790,215 @@ public async Task ExecuteWorkflowAsync_CoroutinesAfterComplete_GetProcessed()
...
+                    new() { StartToCloseTimeout = TimeSpan.FromMinutes(10) }));
+            }
+            finally
+            {
+                semaSlim.Release();
+            }
+        }
+
+        [WorkflowUpdate]
+        public async Task UpdateWithSyncSlim()
```
drewhoskins-temporal, line 4832: Curious why somebody would use this approach in an async handler.

tests/Temporalio.Tests/Worker/WorkflowWorkerTests.cs
```diff
@@ -4789,6 +4790,215 @@ public async Task ExecuteWorkflowAsync_CoroutinesAfterComplete_GetProcessed()
...
+            }
+        }
+
+        public void Dispose()
+        {
+            Dispose(true);
+            GC.SuppressFinalize(this);
+        }
+
+        protected virtual void Dispose(bool disposing)
```
drewhoskins-temporal, line 4880: Ah, I can see why we might want to wrap for convenience.

tests/Temporalio.Tests/Worker/WorkflowWorkerTests.cs
cretz: They wouldn't, this will deadlock. The test is just confirming these sync calls are no good.

tests/Temporalio.Tests/Worker/WorkflowWorkerTests.cs
cretz: Right, this is a bit annoying (not to mention the other bad calls on `SemaphoreSlim`).

cretz:
I am closing this PR. I will be opening a separate one that contains a new `Semaphore` implementation

PR #287, opened 2024-06-25
Implementation of Temporalio.Workflows.Semaphore

cretz:
## What was changed

* Added `Temporalio.Workflows.Semaphore` deterministic semaphore implementation
* Added tests for that implementation
* Added tests to confirm when existing standard library semaphore does/doesn't work
* Updated README with guidance

Opening as draft due to a bug that is occurring when cancelling the workflow while waiting on this semaphore (see TODO in test).

## Checklist

1. Relates to #281

### DIFF
--- a/README.md
+++ b/README.md
@@ -702,6 +702,10 @@ with .NET tasks inside of workflows:
   * Use `Workflow.WhenAnyAsync` instead.
   * Technically this only applies to an enumerable set of tasks with results or more than 2 tasks with results. Other
     uses are safe. See [this issue](https://github.com/dotnet/runtime/issues/87481).
+* Do not use `System.Threading.Semaphore` or `System.Threading.SemaphoreSlim`.
+  * Use `Temporalio.Workflows.Semaphore` instead.
+  * _Technically_ `SemaphoreSlim` does work if only the async form of `WaitAsync` is used without no timeouts and
+    `Release` is used. But anything else can deadlock the workflow and its use is cumbersome since it must be disposed.
 * Be wary of additional libraries' implicit use of the default scheduler.
   * For example, while there are articles for `Dataflow` about
     [using a specific scheduler](https://learn.microsoft.com/en-us/dotnet/standard/parallel-programming/how-to-specify-a-task-scheduler-in-a-dataflow-block),
--- a/src/Temporalio/Worker/ActivityWorker.cs
+++ b/src/Temporalio/Worker/ActivityWorker.cs
@@ -366,13 +366,14 @@ namespace Temporalio.Worker
                     Headers: tsk.Start.HeaderFields)).ConfigureAwait(false);
 
                 completion.Result.Completed = new();
-                // As a special case, ValueTuple is considered "void"
-                if (result?.GetType() != typeof(ValueTuple))
+                // As a special case, ValueTuple is considered "void" which is null
+                if (result?.GetType() == typeof(ValueTuple))
                 {
-                    completion.Result.Completed.Result =
-                        await worker.Client.Options.DataConverter.ToPayloadAsync(
-                            result).ConfigureAwait(false);
+                    result = null;
                 }
+                completion.Result.Completed.Result =
+                    await worker.Client.Options.DataConverter.ToPayloadAsync(
+                        result).ConfigureAwait(false);
             }
             catch (CompleteAsyncException)
             {
--- a/tests/Temporalio.Tests/Worker/WorkflowWorkerTests.cs
+++ b/tests/Temporalio.Tests/Worker/WorkflowWorkerTests.cs
@@ -4,6 +4,7 @@
 namespace Temporalio.Tests.Worker;
 
 using System.Collections.Concurrent;
+using System.Linq.Expressions;
 using System.Threading.Tasks.Dataflow;
 using Microsoft.Extensions.Logging;
 using Temporalio.Activities;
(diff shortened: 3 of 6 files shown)


README.md
```diff
@@ -702,6 +702,10 @@ with .NET tasks inside of workflows:
   * Use `Workflow.WhenAnyAsync` instead.
   * Technically this only applies to an enumerable set of tasks with results or more than 2 tasks with results. Other
     uses are safe. See [this issue](https://github.com/dotnet/runtime/issues/87481).
+* Do not use `System.Threading.Semaphore` or `System.Threading.SemaphoreSlim`.
+  * Use `Temporalio.Workflows.Semaphore` instead.
+  * _Technically_ `SemaphoreSlim` does work if only the async form of `WaitAsync` is used without no timeouts and
+    `Release` is used. But anything else can deadlock the workflow and its use is cumbersome since it must be disposed.
 * Be wary of additional libraries' implicit use of the default scheduler.
```
dandavison, line 709:
```suggestion
  * _Technically_ `SemaphoreSlim` does work if only the async form of `WaitAsync` is used without timeouts and
    `Release` is used. But any other usage pattern can deadlock the workflow and its use is cumbersome since it must be disposed.
* Be wary of additional libraries' implicit use of the default scheduler.
```

src/Temporalio/Worker/ActivityWorker.cs
```diff
@@ -366,13 +366,14 @@ private void StartActivity(Bridge.Api.ActivityTask.ActivityTask tsk)
...
-                if (result?.GetType() != typeof(ValueTuple))
+                // As a special case, ValueTuple is considered "void" which is null
+                if (result?.GetType() == typeof(ValueTuple))
                 {
-                    completion.Result.Completed.Result =
-                        await worker.Client.Options.DataConverter.ToPayloadAsync(
-                            result).ConfigureAwait(false);
+                    result = null;
                 }
+                completion.Result.Completed.Result =
```
Quinn-With-Two-Ns, line 374: Should we add an explicit test to cover this? I guess the semaphore tests cover it as well?

Quinn-With-Two-Ns (approved):

src/Temporalio/Worker/ActivityWorker.cs
cretz: Yes, the semaphore test hit this because it was our first `Task`-no-result activity (as opposed to just `void`). I can add an explicit test.

cretz: Hrmm, so `ExecuteActivityAsync_SimpleAsyncVoidMethod_Succeeds` tested this and succeeds in CI but I don't know why, digging...

cretz: Ok, there are different routes to creating an activity, and that other test didn't test this route. This semaphore test should be sufficient.

tests/Temporalio.Tests/Worker/WorkflowWorkerTests.cs
```diff
@@ -4789,6 +4790,483 @@ public async Task ExecuteWorkflowAsync_CoroutinesAfterComplete_GetProcessed()
...
+                            updateExpr,
+                            new(id: "update-0", waitForStage: WorkflowUpdateStage.Accepted)));
+                    _ = Task.Run(() =>
+                        handle.StartUpdateAsync(
+                            updateExpr,
+                            new(id: "update-1", waitForStage: WorkflowUpdateStage.Accepted)));
+
+                    // Send handle back to check
+                    handleCompletion.SetResult(handle);
+                    // This will never complete
```
dandavison, line 4981: > because the update handlers attempt to execute an activity with a semaphore permit held. Since we are attempting to perform more updates than there are available semaphore permits, this will deadlock.

a comment like that would have saved me time reading the tests.

tests/Temporalio.Tests/Worker/WorkflowWorkerTests.cs
```diff
@@ -4789,6 +4790,483 @@ public async Task ExecuteWorkflowAsync_CoroutinesAfterComplete_GetProcessed()
...
+            [Activity]
+            public Task DoActivityAsync(int number) => UpdateCompletions[number].Task;
+        }
+    }
+
+    [Fact]
+    public async Task ExecuteWorkflowAsync_Semaphore_MultipleWaiters()
+    {
+        // This is a basic test of semaphore usage. We will make a semaphore with 2 allowed and 5
+        // provided and confirm as we complete some that others can go.
```
dandavison, line 5114: Nice comment, would appreciate one of those on the stdlib semaphore tests.

tests/Temporalio.Tests/Worker/WorkflowWorkerTests.cs
```diff
@@ -4789,6 +4790,483 @@ public async Task ExecuteWorkflowAsync_CoroutinesAfterComplete_GetProcessed()
...
+                acts.UpdateCompletions[waiting1[0]].SetResult();
+                acts.UpdateCompletions[waiting1[1]].SetResult();
+                var waiting2 = await AssertMore.EventuallyAsync(async () =>
+                {
+                    var waiting = await handle.QueryAsync(wf => wf.Waiting);
+                    Assert.True(waiting.Count == 2);
+                    Assert.DoesNotContain(waiting, waiting1.Contains);
+                    return waiting;
+                });
+                Assert.Contains(await handle.QueryAsync(wf => wf.Completed), waiting1.Contains);
```
dandavison, line 5144: This assertion seems to be saying

> Assert that at least one of the `waiting1` items is now completed.

If so, it seems that we could make it stronger and assert that both of them are completed. Clearer to use set operations?

tests/Temporalio.Tests/Worker/WorkflowWorkerTests.cs
```diff
@@ -4789,6 +4790,483 @@ public async Task ExecuteWorkflowAsync_CoroutinesAfterComplete_GetProcessed()
...
+                }
+
+                // Confirm completions
+                var completed = await AssertMore.EventuallyAsync(async () =>
+                {
+                    var completed = await handle.QueryAsync(wf => wf.Completed);
+                    Assert.True(completed.Count == 5);
+                    return completed;
+                });
+                Assert.Contains(completed.GetRange(0, 2), waiting1.Contains);
```
dandavison, line 5162: As above, I think we really want to assert that the set of `waiting1` items equals the set of items in `GetRange(0, 2)`, but we're asserting something weaker IIUC.

dandavison (commented):
This all LGTM. Just submitting some minor comments I made while reviewing.

tests/Temporalio.Tests/Worker/WorkflowWorkerTests.cs
cretz: > This assertion seems to be saying
> > Assert that at least one of the waiting1 items is now completed.

It's not. `Assert.Contains(collection, predicate)` means that every item in the collection must match the predicate. This is basically unordered collection comparison (assuming there are no duplicates which I think is safe to assume). Could have used sets though assertion equality/utilities is not as helpful for .NET hash sets.

cretz: > it seems that we could make it stronger and assert that both of them are completed

Yes, I meant this to be `Assert.All` or `Assert.True` w/ `.All` on the query result (set equality doesn't have as simple of assertion tools in .NET). Will set reminder to change.

tests/Temporalio.Tests/Worker/WorkflowWorkerTests.cs
cretz: Same deal here, I misunderstood `Contains(collection, predicate)`, the intent was to check all

tests/Temporalio.Tests/Worker/WorkflowWorkerTests.cs
cretz: https://github.com/temporalio/sdk-dotnet/pull/296

tests/Temporalio.Tests/Worker/WorkflowWorkerTests.cs
cretz: https://github.com/temporalio/sdk-dotnet/pull/296

PR #289, opened 2024-06-26
Accept search attributes for dev server

### LINKED ISSUE #282: [Feature Request] Accept search attributes in dev server startup

See https://github.com/temporalio/features/issues/494. With https://github.com/temporalio/cli/pull/593 deployed, users should be allowed to set search attributes in dev server option that translate to `--search-attribute` calls.

cretz:
## What was changed

* `Temporalio.Testing.WorkflowEnvironmentStartLocalOptions` now accepts `SearchAttributes` option
* Test for the option

~There is an issue with the dev server using search attributes, currently blocked on https://github.com/temporalio/temporal/issues/6195. Leaving as draft for now.~ EDIT: Now ready for review

## Checklist

1. Closes #282


cretz:
This is now ready for review

mjameswh (approved):

mjameswh:
Don't you have any preexisting tests in the repo that were using custom search attributes, and should be updated to use this?

cretz:
> Don't you have any preexisting tests in the repo that were using custom search attributes, and should be updated to use this?

Yes, though not sure they _need_ to be updated since they work today, but they _can_ be. This may not always be the case for other SDKs because if their tests are interchangeable with the time-skipping server, that only works with the operator service approach still.

cretz:
The way the test harness code is setup, it separates the server setup from the search attribute creation. I think it's worth keeping that separation.

PR #293, opened 2024-06-28
Expose resource based autotuning

Sushisource:
## What was changed
Expose resource based auto-tuner

## Why?
Adding in all SDKs

## Checklist

1. Closes 

2. How was this tested:
 Added some tests, manual verification with omes / existing core tests

3. Any docs updates needed?
 Will be after experimental phase


src/Temporalio/Bridge/OptionsExtensions.cs
```diff
@@ -452,20 +469,14 @@ internal static class OptionsExtensions
                     throw new ArgumentException("Unable to get assembly manifest ID for build ID");
                 buildId = entryAssembly.ManifestModule.ModuleVersionId.ToString();
             }
-            var nonDetWorkflowFailForTypes = options.Workflows.Where(
-                w => w.FailureExceptionTypes?.Any(
-                    t => t.IsAssignableFrom(typeof(WorkflowNondeterminismException))) ?? false).
-                Select(w => w.Name).ToArray();
```
Sushisource, line 458: This was unused

src/Temporalio/Bridge/include/temporal-sdk-bridge.h
```diff
@@ -353,6 +353,45 @@ typedef struct WorkerOrFail {
...
+  SlotSupplier_Tag tag;
+  union {
+    struct {
+      struct FixedSizeSlotSupplier fixed_size;
+    };
+    struct {
+      struct ResourceBasedSlotSupplier resource_based;
+    };
+  };
+} SlotSupplier;
```
cretz, line 387: I think this makes our header file a bit uglier to consume from C compared to mutually exclusive options (technically doesn't matter though)

src/Temporalio/Bridge/src/worker.rs
```diff
@@ -49,6 +48,43 @@ pub struct WorkerOptions {
...
+pub struct ResourceBasedSlotSupplier {
+    minimum_slots: usize,
+    maximum_slots: usize,
+    ramp_throttle_ms: u64,
+    tuner_config: ResourceBasedTunerConfig,
+}
+
+#[repr(C)]
+#[derive(Clone, Copy, PartialEq, Debug)]
+pub struct ResourceBasedTunerConfig {
```
cretz:
```suggestion
pub struct ResourceBasedTunerOptions {
```

src/Temporalio/Bridge/src/worker.rs
```diff
@@ -49,6 +48,43 @@ pub struct WorkerOptions {
     nondeterminism_as_workflow_fail_for_types: ByteArrayRefArray,
 }
 
+#[repr(C)]
+pub struct TunerHolder {
```
cretz, line 52:
```suggestion
pub struct WorkerTuner {
```

I think this is a better name, but doesn't matter much to have this separate/different "holder" terminology in the bridge only I suppose so long as user's don't have to see it

src/Temporalio/Worker/TemporalWorkerOptions.cs
```diff
@@ -124,9 +125,9 @@ public TemporalWorkerOptions()
 
         /// <summary>
         /// Gets or sets the maximum number of activities that will ever be given to this worker
-        /// concurrently. Default is 100.
+        /// concurrently. Default is 100. Mutually exclusive with <see cref="Tuner"/>.
         /// </summary>
-        public int MaxConcurrentActivities { get; set; } = 100;
+        public int? MaxConcurrentActivities { get; set; }
```
cretz, line 130: This is technically a backwards incompatible change (people could use this property in some log or something). I wonder if it'd be better to leave these and just say this is ignored if `Tuner` is non-null? I don't have a super strong opinion here, we can probably break this kind of compat if we must.

src/Temporalio/Worker/Tuning/WorkerTuner.cs
```diff
@@ -0,0 +1,106 @@
...
+                workflowTaskSlotSupplier,
+                activityTaskSlotSupplier,
+                localActivitySlotSupplier);
+        }
+
+        /// <summary>
+        /// Gets a slot supplier for workflow tasks.
+        /// </summary>
+        /// <returns>A slot supplier for workflow tasks.</returns>
+        public abstract ISlotSupplier GetWorkflowTaskSlotSupplier();
```
cretz:
```suggestion
        public abstract ISlotSupplier WorkflowTaskSlotSupplier { get; }
```

Lets make these abstract properties instead of getters

src/Temporalio/Worker/Tuning/WorkerTuner.cs
```diff
@@ -0,0 +1,106 @@
+namespace Temporalio.Worker.Tuning
+{
+    /// <summary>
+    /// WorkerTuners allow for the dynamic customization of some aspects of worker configuration.
+    /// </summary>
+    public abstract class WorkerTuner
```
cretz: I have an alternative suggestion here. .NET prefers interfaces and the default worker tuner can just be composite, so how about something like:

```csharp
public interface IWorkerTuner
{
    ISlotSupplier WorkflowTaskSlotSupplier { get; }
    ISlotSupplier ActivityTaskSlotSupplier { get; }
    ISlotSupplier LocalActivitySlotSupplier { get; }
}

// Not a record on purpose
public class WorkerTuner : IWorkerTuner
{
    public static WorkerTuner CreateResourceBased(...)

    public static WorkerTuner CreateFixed(...)

    public WorkerTuner(
        ISlotSupplier workflowTaskSlotSupplier,
        ISlotSupplier activityTaskSlotSupplier,
        ISlotSupplier localActivitySlotSupplier)
    {
        WorkflowTaskSlotSupplier = workerTaskSlotSupplier;
        ActivityTaskSlotSupplier = activityTaskSlotSupplier;
        LocalActivitySlotSupplier = localActivitySlotSupplier;
    }

    public ISlotSupplier WorkflowTaskSlotSupplier { get; init; }
    public ISlotSupplier ActivityTaskSlotSupplier { get; init; }
    public ISlotSupplier LocalActivitySlotSupplier { get; init; }
}
```

And get rid of `CompositeTuner` and accept `IWorkerTuner` in the worker options. I think this is more idiomatic .NET (have interface and common impl as the primary impl). Arguably it'd be more idiomatic in .NET (unlike Python and others) to also make `ResourceBasedTuner` a class with a constructor instead of a static helper here. So could be something like:

```csharp
public interface IWorkerTuner
{
    ISlotSupplier WorkflowTaskSlotSupplier { get; }
    ISlotSupplier ActivityTaskSlotSupplier { get; }
    ISlotSupplier LocalActivitySlotSupplier { get; }
}

// Not a record on purpose
public class WorkerTuner : IWorkerTuner
{
    public WorkerTuner(
        ISlotSupplier workflowTaskSlotSupplier,
        ISlotSupplier activityTaskSlotSupplier,
        ISlotSupplier localActivitySlotSupplier)
    {
        WorkflowTaskSlotSupplier = workerTaskSlotSupplier;
        ActivityTaskSlotSupplier = activityTaskSlotSupplier;
        LocalActivitySlotSupplier = localActivitySlotSupplier;
    }

    public ISlotSupplier WorkflowTaskSlotSupplier { get; init; }
    public ISlotSupplier ActivityTaskSlotSupplier { get; init; }
    public ISlotSupplier LocalActivitySlotSupplier { get; init; }
}

public class ResourceBasedTuner : WorkerTuner
{
    public ResourceBasedTuner(
        double targetMemoryUsage,
        double targetCpuUsage,
        ResourceBasedSlotSupplierOptions? workflowOptions = null,
        ResourceBasedSlotSupplierOptions? activityOptions = null,
        ResourceBasedSlotSupplierOptions? localActivityOptions = null)
        : base(
            new ResourceBasedSlotSupplier(workflowOptions ?? new(), new(targetMemoryUsage, targetCpuUsage),
            new ResourceBasedSlotSupplier(activityOptions ?? new(), new(targetMemoryUsage, targetCpuUsage),
            new ResourceBasedSlotSupplier(localActivityOptions ?? new(), new(targetMemoryUsage, targetCpuUsage))
    {
    }
}

public class FixedSizeTuner : WorkerTuner
{
    public FixedSizeTuner(int workflowTaskSlots, int activityTaskSlots, int localActivitySlots)
        : base(
            new FixedSizeSlotSupplier(workflowTaskSlots),
            new FixedSizeSlotSupplier(activityTaskSlots),
            new FixedSizeSlotSupplier(localActivitySlots))
    {
    }
}
```

I think that last one is the most idiomatic

src/Temporalio/Worker/Tuning/WorkerTuner.cs
```diff
@@ -0,0 +1,106 @@
...
+        }
+
+        /// <summary>
+        /// Create a fixed-size tuner with the given slot capacities.
+        /// </summary>
+        /// <param name="numWorkflowTaskSlots">The number of available workflow task slots.</param>
+        /// <param name="numActivityTaskSlots">The number of available activity task slots.</param>
+        /// <param name="numLocalActivitySlots">The number of available local activity slots.</param>
+        /// <returns>The tuner.</returns>
+        public static WorkerTuner CreateFixed(
```
cretz:
```suggestion
        public static WorkerTuner CreateFixedSize(
```

To be consistent

src/Temporalio/Bridge/OptionsExtensions.cs
```diff
@@ -483,6 +494,86 @@ internal static class OptionsExtensions
...
+                activity_slot_supplier =
+                    tuner.GetActivityTaskSlotSupplier().ToInteropSlotSupplier(false),
+                local_activity_slot_supplier =
+                    tuner.GetLocalActivitySlotSupplier().ToInteropSlotSupplier(false),
+            };
+        }
+
+        private static Interop.SlotSupplier ToInteropSlotSupplier(
+                this Temporalio.Worker.Tuning.ISlotSupplier supplier,
+                bool isWorkflow)
```
cretz:
```suggestion
            this Temporalio.Worker.Tuning.ISlotSupplier supplier,
            bool isWorkflow)
```

Inconsistent spacing

src/Temporalio/Worker/Tuning/WorkerTuner.cs
```diff
@@ -0,0 +1,106 @@
...
+            uint numWorkflowTaskSlots,
+            uint numActivityTaskSlots,
+            uint numLocalActivitySlots)
```
cretz: While it makes more sense in Rust, it's not that common in .NET to use uints unless you _really_ need to (note our options don't really). Can check for negative at runtime if needed. Same type of thing as Go, most people pass around the common integer type and usually they aren't expected to convert for value-safety reasons.

cretz (commented):
Mostly LGTM. I do have a suggestion about reworking the worker tuner hierarchy in a comment below.

src/Temporalio/Bridge/OptionsExtensions.cs
cretz: Thanks! Wonder why there isn't an unused-local-var check in .NET, maybe I have it turned off

src/Temporalio/Worker/Tuning/FixedSizeSlotSupplier.cs
```diff
@@ -0,0 +1,8 @@
+namespace Temporalio.Worker.Tuning
+{
+    /// <summary>
+    /// A slot supplier that will only ever issue at most a fixed number of slots.
+    /// </summary>
+    /// <param name="NumSlots">The maximum number of slots that will ever be issued.</param>
+    public sealed record FixedSizeSlotSupplier(uint NumSlots) : ISlotSupplier;
```
cretz:
```suggestion
    public sealed record FixedSizeSlotSupplier(int SlotCount) : ISlotSupplier;
```

Besides the int/uint thing, I think .NET would prefer either `Count` suffix or just `Slots` over "num".

src/Temporalio/Bridge/include/temporal-sdk-bridge.h
Sushisource: :man_shrugging: Eh, not really a problem IMO. If we care about the header being as easy as possible we can do that when we have another consumer, but honestly I think this is pretty easy to use as-is.

src/Temporalio/Bridge/src/worker.rs
Sushisource: This matches the thing it's getting turned into Core side so I think it's worth keeping

src/Temporalio/Worker/TemporalWorkerOptions.cs
Sushisource: I think it's probably worth it. It's more clear at the type level about what's going on and I think is fairly unlikely to break anyone and easy to fix if it is.

src/Temporalio/Bridge/OptionsExtensions.cs
Sushisource: Dotnet format seems to just... not format stuff I would expect it to fairly often

src/Temporalio/Worker/Tuning/WorkerTuner.cs
Sushisource: I'm not sure what would rise to the definition of "really need to" in the sense that things that shouldn't be negative shouldn't be negative and that's a fairly binary distinction.

It's annoying as hell that unsigned types exist in these languages but aren't used for whatever reason. Oh well though, I'll change it.

src/Temporalio/Bridge/OptionsExtensions.cs
cretz: Yeah, I don't use it to rewrite code, I just use it to know CI will pass

src/Temporalio/Worker/Tuning/ResourceBasedSlotSupplierOptions.cs
```diff
@@ -0,0 +1,26 @@
...
+    /// This value matters because how many resources a task will use cannot be determined ahead of
+    /// time, and thus the system should wait to see how much resources are used before issuing more
+    /// slots.
+    /// </param>
+    /// <remarks>
+    /// WARNING: Resource based tuning is currently experimental.
+    /// </remarks>
+    public sealed record ResourceBasedSlotSupplierOptions(
+        uint? MinimumSlots = null,
+        uint? MaximumSlots = null,
```
cretz: Can we switch these to ints too?

src/Temporalio/Worker/Tuning/WorkerTuner.cs
cretz: Yeah, that is a bit annoying. I think the reasoning is more about the specific-bit-size integers and just carried over to here. For those, people avoid them unless they need specific binary sizes or are in a space-conscious place (e.g. embedded). And I think the general "use int for integers unless there is a good use case not to" was kinda the general accepted thing and applied to unsigned as well as specific-bit-size types. The only reason I bring up here is because in cases where uints or other things are used, .NET has a whole thing about unchecked conversions and it's a bit of a pain for devs using ints everywhere else.

src/Temporalio/Worker/Tuning/WorkerTuner.cs
```diff
@@ -0,0 +1,106 @@
...
+        /// Create a fixed-size tuner with the given slot capacities.
+        /// </summary>
+        /// <param name="numWorkflowTaskSlots">The number of available workflow task slots.</param>
+        /// <param name="numActivityTaskSlots">The number of available activity task slots.</param>
+        /// <param name="numLocalActivitySlots">The number of available local activity slots.</param>
+        /// <returns>The tuner.</returns>
+        public static WorkerTuner CreateFixedSize(
+            int numWorkflowTaskSlots,
+            int numActivityTaskSlots,
+            int numLocalActivitySlots)
```
cretz:
```suggestion
            int workflowTaskSlots,
            int activityTaskSlots,
            int localActivitySlots)
```

Not a big deal, but if you do end up making another commit, might as well (or suffix with `Count` or whatever). Matters a bit more here than in Java because .NET supports named parameters.

cretz: Btw, for single-line method implementations, can use `=>` instead of a block + return. Makes no difference though.

cretz (approved):
May need to downgrade proto to 23.x and regen those

cretz (commented):
LGTM, will leave unapproved pending further investigation for the `worker.is_null()` thing on `worker_free` and there may be some whitespace issues.

PR #298, opened 2024-07-01
Add Temporalio.Workflows.Mutex

### LINKED ISSUE #297: [Feature Request] Add Temporalio.Workflows.Mutex as thin facade over single-count Temporalio.Workflows.Semaphore

Add Temporalio.Workflows.Mutex as thin facade over single-count Temporalio.Workflows.Semaphore. Maybe add methods/overloads to match the existing threading mutex method names (`WaitOne` and `ReleaseMutex`).

cretz:
## What was changed

Added `Temporalio.Workflows.Mutex` which is a thin layer over the semaphore

1. Closes #297


src/Temporalio/Workflows/Mutex.cs
```diff
@@ -0,0 +1,93 @@
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Temporalio.Workflows
+{
+    /// <summary>
+    /// Mutex is a thin wrapper around <see cref="Semaphore"/> with a single count.
```
Sushisource: Probably mention "And it is a safe alternative to System.Threading.Mutex"

I'd add that language to the semaphores too if they don't have it.

Sushisource (approved):
Would it make sense to add support for using this withing a `using` block? Seems like that could be nice.

cretz:
> Would it make sense to add support for using this withing a using block

In this case we are matching existing .NET APIs which do not leverage `IDisposable` to release mutexes/semaphores. I think .NET chose explicit release vs using block (akin to a Rust guard) for .NET developers to be clear about when to release (may not otherwise be clear especially for those with `using` as a statement instead of a block).

src/Temporalio/Workflows/Mutex.cs
cretz: Yes, it was already on semaphore. Added it here.

dandavison (approved):
LGTM!

Sushisource:
> > Would it make sense to add support for using this withing a using block
> 
> In this case we are matching existing .NET APIs which do not leverage `IDisposable` to release mutexes/semaphores. I think .NET chose explicit release vs using block (akin to a Rust guard) for .NET developers to be clear about when to release (may not otherwise be clear especially for those with `using` as a statement instead of a block).

FYI it very much does work with disposable https://learn.microsoft.com/en-us/dotnet/api/system.threading.mutex?view=net-8.0#remarks

PR #306, opened 2024-07-15
Doc fixes for cref's so that rider doesn't show red warnings

jakejscott:
## What was changed

In Rider, when you open the solutions there's a bunch of red warning showing up for cref docs so I've updated them.

## Why?

This fixes the solution wide analysis that Rider runs

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/src/Temporalio/Activities/ActivityInfo.cs
+++ b/src/Temporalio/Activities/ActivityInfo.cs
@@ -55,7 +55,7 @@ namespace Temporalio.Activities
     {
         /// <summary>
         /// Gets the value that is set on
-        /// <see cref="Microsoft.Extensions.Logging.ILogger.BeginScope" /> before this activity is
+        /// <see cref="Microsoft.Extensions.Logging.ILogger.BeginScope{TState}" /> before this activity is
         /// started.
         /// </summary>
         internal Dictionary<string, object> LoggerScope { get; } = new()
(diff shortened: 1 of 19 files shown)


cretz:
Thanks!

So there seems two issues with Rider that the .NET team and docfx don't consider issues: 1) Not requiring type arguments when not needed for overload ambiguity (why is Rider ok w/ not providing parameters in these situations but not ok w/ not providing type arguments), and 2) inability to resolve references to things in the scope they are in.

I think these are both Rider bugs if https://learn.microsoft.com/en-us/dotnet/csharp/misc/cs1574 does not warn like it does with `cref`s .NET/Roslyn actually can't resolve. Otherwise, these are just going to keep happening because not all contributors use this IDE and this IDE is the only one with these problems.

rstm-sf:
@cretz hello

In the end, what decision will we come to? I fixed it too, without being sure that anyone was doing this...

cretz:
I think if `dotnet` and `docfx` consider these valid, this is an IDE specific issue. We count on the compiler to warn on this and unless we use Rider-specific analyzers in CI, we can't guarantee future PRs won't cause more Rider-only warnings. I wonder if there is already an open bug for JetBrains we can link or if we should create one?

I would rather not merge this temporary fix for a single IDE knowing that it will just happen again. I think the IDE should fix their inconsistency with the .NET tooling.

PR #315, opened 2024-07-25
Command reordering in Core

### LINKED ISSUE #312: [Feature Request] Update core and confirm new post-complete command reordering

Core now has command reordering to move first workflow completion last. We had a poor attempt at this before in two ways: first was only truncate after _last_ completion command (as opposed to new Core which is after _first_ completion command), and second didn't account for multiple completion commands. Still discussing how we want to retain compatibility.

cretz:
## What was changed

* Update core
* Have .NET command-reorder behavior only apply if replaying with that flag (i.e. no longer apply and set the flag for new executions)
* Removed internal-for-test-use-only option to disable .NET reordering
* Add tests for post-completion commands including replays for pre-.NET-flag, post-.NET-flag, and current (post-Core-flag)
* Add tests for multi-completion success first including replays for pre-.NET-flag, post-.NET-flag, and current (post-Core-flag)
* Add tests for multi-completion failure first including replays for pre-.NET-flag, post-.NET-flag, and current (post-Core-flag)

## Checklist

1. Closes #312


src/Temporalio/Worker/WorkflowInstance.cs
```diff
@@ -1459,21 +1450,22 @@ private string GetStackTrace()
             }
             workflowComplete = completionCommandIndex >= 0;
 
-            // This only applies for successful activations that have a completion not at the end,
-            // don't have completion reordering disabled, and that are either not replaying or have
-            // the flag set.
+            // In a previous version of .NET, if this was a successful activation completion with
```
dandavison: nit: (there might be a few other places in the repo)

```suggestion
            // In a previous version of .NET SDK, if this was a successful activation completion with
```

src/Temporalio/Worker/WorkflowInstance.cs
```diff
@@ -1425,20 +1423,11 @@ private string GetStackTrace()
...
         {
-            // In earlier versions of the SDK we allowed commands to be sent after workflow
-            // completion. These ended up being removed effectively making the result of the
-            // workflow function mean any other later coroutine commands be ignored. To match
-            // Go/Java, we are now going to move workflow completion to the end so that
-            // same-task-post-completion commands are still accounted for.
-            //
-            // Note this only applies for successful activations that don't have completion
-            // reordering disabled and that are either not replaying or have the flag set.
-
```
dandavison, line 1441: It would be worth leaving a comment for any future readers explaining that the code below assumes there is only one completion command. Otherwise having "the completion" in comments could confuse readers, making them forget that there could be multiple.

src/Temporalio/Worker/WorkflowInstance.cs
```diff
@@ -1459,21 +1448,22 @@ private string GetStackTrace()
...
                 return;
             }
 
-            // Now we know the completion is in the wrong spot and we're on a newer SDK, so set the
-            // SDK flag and move it
+            // Now we know that we're replaying w/ the flag set and the completion in the wrong
+            // spot, so set the SDK flag and move it
             completion.Successful.UsedInternalFlags.Add((uint)WorkflowLogicFlag.ReorderWorkflowCompletion);
             var compCmd = completion.Successful.Commands[completionCommandIndex];
             completion.Successful.Commands.RemoveAt(completionCommandIndex);
```
dandavison: So just to try to recap this logic for myself and anyone else. It would require a fairly complicated diagram to set it all out, but here's one story:

- This release of .NET will also be the first .NET release that has the new core command reordering code.

- So, for example, let `C` be a completion command and `N` be a non-completion, and consider a workflow being processed today by a .NET worker on the recently released code. Suppose the raw lang sequence is `N C1 N C2 N`. The .NET flag is set, so .NET will change that to `N C1 N N C2`. And core will change that to `N C1`, since it's old core.

- Now, if we release .NET workers based on this PR, the same thing will happen on replay, because the .NET flag will be set, and the new core internal flag will not be set, so both .NET and core will do their legacy behavior.

Sushisource (approved):

dandavison (approved):
LGTM. I'll look at the tests later when I work on Python.

src/Temporalio/Worker/WorkflowInstance.cs
dandavison: Now consider a workflow born after this PR is released. Again the raw commands collected in lang are `N C1 N C2 N`. lang does not do anything to them, since lang's flag is not set, and core receives them unchanged. Core then does its new behavior, and changes them to `N N N C1`, and sets core's flag, so that this will also be done on replay.

cretz: Yes, everything in your statements is correct

src/Temporalio/Worker/WorkflowInstance.cs
cretz: I can change the variable name to `lastCompletionCommandIndex` if that helps

src/Temporalio/Worker/WorkflowInstance.cs
cretz: A bit pedantic for a code comment, but I understand the ambiguity to the reader. Will change.

dandavison: If a reviewer prefixes their comment with "nit", you can't complain it's pedantic :) Universal law of code review!

src/Temporalio/Worker/WorkflowInstance.cs
dandavison: Sure yes, and ideally "the last completion" in comments rather than "the completion" but, that's prefixed by "nit"!

src/Temporalio/Worker/WorkflowInstance.cs
cretz: Very true, I definitely missed that prefacing. My bad (and fixed anyways).

src/Temporalio/Worker/WorkflowInstance.cs
cretz: Updated var name, yeah "the completion in question" is hopefully implied now

PR #318, opened 2024-08-01
HTTP CONNECT proxy support

### LINKED ISSUE #215: [Feature Request] HTTP CONNECT proxy support

* Support HTTP connect proxy options added in https://github.com/temporalio/sdk-core/pull/714. See how Python did it at
https://github.com/temporalio/sdk-python/pull/501
* Add test in features repo

adeal:
## What was changed
This change adds HTTP CONNECT proxy support.

- No need to update core submodule
- Added `Temporalio.Client.HttpConnectProxyConfig`
- Updated client to accept that config

## Checklist

1. Closes #215 

2. How was this tested:
  This is challenging to test w/o a local proxy setup, but I'm open to recommendations. If I'm able, I'll publish a local release and see if I can test this against an environment where proxy resources exist.

4. Any docs updates needed?

### DIFF
--- a/src/Temporalio/Bridge/Interop/Interop.cs
+++ b/src/Temporalio/Bridge/Interop/Interop.cs
@@ -561,7 +561,7 @@ namespace Temporalio.Bridge.Interop
     {
         public SlotSupplier_Tag tag;
 
-        [NativeTypeName("__AnonymousRecord_temporal-sdk-bridge_L380_C3")]
+        [NativeTypeName("__AnonymousRecord_temporal-sdk-bridge_L387_C3")]
         public _Anonymous_e__Union Anonymous;
 
         internal ref FixedSizeSlotSupplier fixed_size
@@ -590,11 +590,11 @@ namespace Temporalio.Bridge.Interop
         internal unsafe partial struct _Anonymous_e__Union
         {
             [FieldOffset(0)]
-            [NativeTypeName("__AnonymousRecord_temporal-sdk-bridge_L381_C5")]
+            [NativeTypeName("__AnonymousRecord_temporal-sdk-bridge_L388_C5")]
             public _Anonymous1_e__Struct Anonymous1;
 
             [FieldOffset(0)]
-            [NativeTypeName("__AnonymousRecord_temporal-sdk-bridge_L384_C5")]
+            [NativeTypeName("__AnonymousRecord_temporal-sdk-bridge_L391_C5")]
             public _Anonymous2_e__Struct Anonymous2;
 
             internal partial struct _Anonymous1_e__Struct
--- a/src/Temporalio/Bridge/Interop/Interop.cs
+++ b/src/Temporalio/Bridge/Interop/Interop.cs
@@ -141,6 +141,18 @@ namespace Temporalio.Bridge.Interop
         public ulong timeout_millis;
     }
 
+    internal partial struct ClientHttpConnectProxyConfig
+    {
+        [NativeTypeName("struct ByteArrayRef")]
+        public ByteArrayRef target_addr;
+
+        [NativeTypeName("struct ByteArrayRef")]
+        public ByteArrayRef username;
+
+        [NativeTypeName("struct ByteArrayRef")]
+        public ByteArrayRef password;
+    }
+
     internal unsafe partial struct ClientOptions
     {
         [NativeTypeName("struct ByteArrayRef")]
@@ -169,6 +181,9 @@ namespace Temporalio.Bridge.Interop
 
         [NativeTypeName("const struct ClientKeepAliveOptions *")]
         public ClientKeepAliveOptions* keep_alive_options;
+
+        [NativeTypeName("const struct ClientHttpConnectProxyConfig *")]
+        public ClientHttpConnectProxyConfig* http_connect_proxy_options;
     }
 
     internal unsafe partial struct ByteArray
--- a/src/Temporalio/Bridge/Interop/Interop.cs
+++ b/src/Temporalio/Bridge/Interop/Interop.cs
@@ -144,7 +144,7 @@ namespace Temporalio.Bridge.Interop
     internal partial struct ClientHttpConnectProxyOptions
     {
         [NativeTypeName("struct ByteArrayRef")]
-        public ByteArrayRef target_addr;
+        public ByteArrayRef target_host;
 
         [NativeTypeName("struct ByteArrayRef")]
         public ByteArrayRef username;
--- a/src/Temporalio/Bridge/Interop/Interop.cs
+++ b/src/Temporalio/Bridge/Interop/Interop.cs
@@ -141,7 +141,7 @@ namespace Temporalio.Bridge.Interop
         public ulong timeout_millis;
     }
 
-    internal partial struct ClientHttpConnectProxyConfig
+    internal partial struct ClientHttpConnectProxyOptions
     {
         [NativeTypeName("struct ByteArrayRef")]
         public ByteArrayRef target_addr;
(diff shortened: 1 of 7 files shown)


src/Temporalio/Client/HttpConnectProxyConfig.cs
```diff
@@ -0,0 +1,26 @@
...
+{
+    /// <summary>
+    /// HTTP connect proxy options for Temporal connections.
+    /// </summary>
+    public class HttpConnectProxyConfig : ICloneable
+    {
+        /// <summary>
+        /// Gets or sets the target host to proxy through as a host:port string.
+        /// </summary>
+        public string? TargetHost { get; set; }
```
adeal, line 13: This probably shouldn't be nullable.

src/Temporalio/Client/TemporalConnectionOptions.cs
```diff
@@ -59,6 +60,14 @@ public TemporalConnectionOptions()
         /// </remarks>
         public KeepAliveOptions? KeepAlive { get; set; } = new();
 
+        /// <summary>
+        /// Gets or sets HTTP connect proxy options for this connection.
+        /// </summary>
+        /// <remarks>
+        /// Default enabled, set to null to disable.
```
adeal: default behavior should be flipped.

update remarks

src/Temporalio/Client/TemporalConnectionOptions.cs
```diff
@@ -59,6 +60,14 @@ public TemporalConnectionOptions()
...
+        /// </remarks>
+        public HttpConnectProxyConfig? HttpConnectProxy { get; set; } = new();
```
adeal: @cretz should this be:

```suggestion
        public HttpConnectProxyConfig? HttpConnectProxyConfig { get; set; } = new();
```

to align with python?

src/Temporalio/Client/TemporalConnectionOptions.cs
```diff
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using Temporalio.Bridge.Interop;
```
adeal: rm

src/Temporalio/Bridge/Interop/Interop.cs
```diff
@@ -141,6 +141,18 @@ internal partial struct ClientKeepAliveOptions
         public ulong timeout_millis;
     }
 
+    internal partial struct ClientHttpConnectProxyConfig
+    {
+        [NativeTypeName("struct ByteArrayRef")]
+        public ByteArrayRef target_addr;
```
adeal:
```suggestion
        public ByteArrayRef target_host;
```

cretz: This file should be auto-generated via `ClangSharpPInvokeGenerator` so it should match the Rust names

src/Temporalio/Bridge/include/temporal-sdk-bridge.h
```diff
@@ -106,6 +106,7 @@ typedef struct ClientOptions {
   const struct ClientTlsOptions *tls_options;
   const struct ClientRetryOptions *retry_options;
   const struct ClientKeepAliveOptions *keep_alive_options;
+  const HttpConnectProxyOptions *http_connect_proxy_options;
```
cretz: This is auto-generated as part of the Rust build and therefore should match the Rust constructs

src/Temporalio/Client/HttpConnectProxyConfig.cs
```diff
@@ -0,0 +1,26 @@
+using System;
+
+namespace Temporalio.Client
+{
+    /// <summary>
+    /// HTTP connect proxy options for Temporal connections.
+    /// </summary>
+    public class HttpConnectProxyConfig : ICloneable
```
cretz:
```suggestion
    public class HttpConnectProxyOptions : ICloneable
```

Matches the term we've used elsewhere

cretz: In addition to a parameterless constructor, we should have one that accepts `TargetHost` and we should make sure in `ToInteropOptions` that this is non-null (i.e. it is "required" but not required on class construction).

adeal: I mentioned this in the description a bit, but it seems like you all established a pattern to diffferentiate the core types from the lang types. Since the core type is already `HttpConnectProxyOptions`, I opted for the above. Plus, I wasn't sure if being consistent with the python sdk was important.

Will update.

adeal: That makes sense.

src/Temporalio/Client/TemporalConnectionOptions.cs
cretz: No, it's more important to align with ourselves here IMO (e.g. keep alive settings)

src/Temporalio/Client/HttpConnectProxyConfig.cs
cretz: :+1: I think keep alive options can be used as a good template here

adeal: Updated

src/Temporalio/Client/TemporalConnectionOptions.cs
adeal: I agree, consistency within a single sdk is more important than across sdks.

Kept as-is, thanks for the input.

src/Temporalio/Client/HttpConnectProxyConfig.cs
adeal: Updated.

src/Temporalio/Bridge/Interop/Interop.cs
cretz: Unresolving. This doesn't seem to match the Rust name (this file is autogenerated)

src/Temporalio/Bridge/include/temporal-sdk-bridge.h
adeal: I'm getting an inconsistent experience, but this is ultimately what is gen'd locally for me.

src/Temporalio/Client/HttpConnectProxyOptions.cs
```diff
@@ -0,0 +1,32 @@
...
+    /// <summary>
+    /// HTTP connect proxy options for Temporal connections.
+    /// </summary>
+    public class HttpConnectProxyOptions : ICloneable
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="HttpConnectProxyOptions"/> class.
+        /// </summary>
+        /// <param name="targetHost">A 'host:port' string representing the target to proxy through.</param>
+        public HttpConnectProxyOptions(string targetHost) => TargetHost = targetHost;
```
cretz, line 24: There needs to be a default no-arg constructor too for how others use it, the single-arg constructor is an overload, not the only constructor

src/Temporalio/Bridge/Interop/Interop.cs
adeal: Sorry, misunderstood you. I thought you were veto'ing my proposed name change comment above.

src/Temporalio/Client/HttpConnectProxyOptions.cs
adeal: Updated, thanks for the education on this!

src/Temporalio/Bridge/include/temporal-sdk-bridge.h
cretz: No prob, I may have just assumed it was going to generate `const struct`. If CI passes w/ what is generated here we're all good. Can ignore.

src/Temporalio/Bridge/Interop/Interop.cs
adeal: Updated.

cretz (approved):
Everything looks good. I think CI is failing because of the line endings of the `Interop.cs` file maybe? Unsure, I will try to look shortly (and may try out via merge to your fork branch if that's ok), but everything in this PR in general LGTM.

cretz:
There was an issue where the FFI struct referenced `HttpConnectProxyOptions` instead of `ClientHttpConnectProxyOptions`. Fixed and regen'd. When CI passes, I will update PR from main then merge into main.

adeal:
> There was an issue where the FFI struct referenced `HttpConnectProxyOptions` instead of `ClientHttpConnectProxyOptions`. Fixed and regen'd. When CI passes, I will update PR from main then merge into main.

AH, obvious in hindsight. Thanks for tracking it down and making the fix.

PR #320, opened 2024-08-03
Allow updating search attributes from schedule updates

### LINKED ISSUE #314: [Feature Request] Support schedule search attribute update

See https://github.com/temporalio/features/issues/512 and the Go implementation at https://github.com/temporalio/sdk-go/issues/1561

Sushisource:
## What was changed
Added ability to update the SAs for schedules
Also updated Core, largely to eliminate some compiler warnings from new Rust version

## Why?
Parity

## Checklist

1. Closes https://github.com/temporalio/sdk-dotnet/issues/314
2. How was this tested:
Added tests

3. Any docs updates needed?

### DIFF
--- a/src/Temporalio/Bridge/src/runtime.rs
+++ b/src/Temporalio/Bridge/src/runtime.rs
@@ -55,14 +55,6 @@ pub struct ForwardedLog {
 }
 
 #[repr(C)]
-pub struct ForwardedLogDetails {
-    target: ByteArray,
-    message: ByteArray,
-    timestamp_millis: u64,
-    fields_json: ByteArray,
-}
-
-#[repr(C)]
 pub enum ForwardedLogLevel {
     Trace = 0,
     Debug,
--- a/src/Temporalio/Client/Schedules/ScheduleUpdate.cs
+++ b/src/Temporalio/Client/Schedules/ScheduleUpdate.cs
@@ -1,8 +1,25 @@
+using Temporalio.Common;
+
 namespace Temporalio.Client.Schedules
 {
     /// <summary>
     /// An update returned from an updater.
     /// </summary>
     /// <param name="Schedule">Schedule to update.</param>
-    public record ScheduleUpdate(Schedule Schedule);
+    /// <param name="SearchAttributes">Optional indexed attributes that can be used while querying
+    /// schedules via the list schedules APIs. The key and value type must be registered with
+    /// Temporal server.
+    ///
+    /// <ul>
+    /// <li>If null, the search attributes will not be updated.</li>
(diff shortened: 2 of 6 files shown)


Sushisource:
I looked at the Go impl while doing this and I think the language here needs to be changed: https://github.com/temporalio/sdk-go/pull/1562/files#diff-03e17ff1d7fdefff7ba8a41ea26b6f3a3135e4b084e58cef60574e60bfd9a9f2R493

It is distinctly _not_ an upsert operation. It is straight-up replacement. An upsert would preserve unchanged attributes if they are not included in the new collection, but they are wiped out unless you manually copy the existing attributes.

src/Temporalio/Bridge/src/runtime.rs
```diff
@@ -54,14 +54,6 @@ pub struct ForwardedLog {
     fields_json: Arc<Mutex<Option<String>>>,
 }
 
-#[repr(C)]
-pub struct ForwardedLogDetails {
```
Sushisource, line 58: This was unused

Quinn-With-Two-Ns:
Yeah would be in favor of fixing the Go SDK docs as well

src/Temporalio/Client/Schedules/ScheduleUpdate.cs
```diff
@@ -1,8 +1,25 @@
...
+    /// <li>If present but empty, the search attributes will be cleared.</li>
+    /// <li>If present and non-empty, all search attributes will be updated to the provided
+    ///     collection. This means existing attributes which do not exist in the provided collection
+    ///     will be wiped out. You can copy attributes from the existing <see
+    ///     cref="ScheduleDescription"/> to avoid this.</li>
+    /// </ul>
+    /// </param>
+    public record ScheduleUpdate(
+        Schedule Schedule,
+        SearchAttributeCollection? SearchAttributes = null);
```
cretz:
```suggestion
        SearchAttributeCollection? TypedSearchAttributes = null);
```

This matches the name on `ScheduleOptions` when creating. I considered accepting `ScheduleOptions` when the inevitable memo update comes about, but it seemed more bad than good.

src/Temporalio/Client/TemporalClient.Schedules.cs
```diff
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
+using Google.Protobuf;
```
cretz:
```suggestion
```

Sushisource: Stupid ide

cretz (approved):

PR #335, opened 2024-08-24
Rename solution file Temporalio.sln -> sdk-dotnet.sln

adeal:
## What was changed

Update solution file name Temporalio.sln -> sdk-dotnet.sln

## Why?
Feel free to reject, I'm not confident whether this breaks any downstream processes like build, packaging, etc.
This is just something I noticed while making a bug fix.

From an IDE perspective, most (all?) editors display the solution name as the IDE window name. Having `sdk-dotnet` makes it easier to identify that this project is the sdk rather than other Temporalio projects (e.g. samples). This also aligns with other sdk clients.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


cretz:
> This also aligns with other sdk clients.

Can you clarify this bit? Which other SDK clients? I don't have a strong opinion on solution name, but a quick look at https://github.com/HangfireIO/Hangfire, https://github.com/Azure/azure-functions-durable-extension, https://github.com/dotnet/orleans, and others, the solution file is named the same as the root namespace of the projects within.

adeal:
> > This also aligns with other sdk clients.
> 
> Can you clarify this bit? Which other SDK clients? I don't have a strong opinion on solution name, but a quick look at https://github.com/HangfireIO/Hangfire, https://github.com/Azure/azure-functions-durable-extension, https://github.com/dotnet/orleans, and others, the solution file is named the same as the root namespace of the projects within.


@cretz Good call on those other open source examples. From experience, I don't know I've ever followed a "root namespace" as a strict convention.

Other sdks that don't have a top level "solution" container will typically just render the root directory as the window name. In those cases it would be sdk-core, sdk-python, etc.

This is not a change I feel strongly about, so feel free to reject.

cretz:
Thanks! Yeah looking at other projects in the space, the current naming approach is quite common.

PR #336, opened 2024-08-24
Fix HTTP CONNECT proxy options not passing to core

adeal:
## Why?
Follow-up fix to #318.

The proxy settings were not being pushed down to core. Although the ClientHttpConnectProxyOptions were properly translated to core's HttpConnectProxyOptions, the TryFrom trait on the CoreClientOptions was not updated to include the new ClientHttpConnectProxyOptions.

## Checklist

1. Closes 
N/A

2. How was this tested:

Unfortunately, there still is no viable path for automated test coverage as-is. However, I validated this change by configuring a temporal app to point to an invalid proxy location and verified that a connection was not successful. Previously, no error would occur since the connection was not being routed to the proxy target.

```csharp
var client = await TemporalClient.ConnectAsync(new TemporalClientConnectOptions()
{
    TargetHost = "<valid_host>",
    HttpConnectProxy = new HttpConnectProxyOptions("localhost:9392"), // Arbitrary invalid location
});
```

```
Unhandled exception. System.InvalidOperationException: Connection failed: Server connection error: tonic::transport::Error(Transport, ConnectError(client error (Connect)

Caused by:
    Connection refused (os error 61)))
   at Temporalio.Bridge.Client.ConnectAsync(Runtime runtime, TemporalConnectionOptions options)
   at Temporalio.Client.TemporalConnection.GetBridgeClientAsync()
   at Temporalio.Client.TemporalConnection.ConnectAsync(TemporalConnectionOptions options)
   at Temporalio.Client.TemporalClient.ConnectAsync(TemporalClientConnectOptions options)
   at Program.<Main>$(String[] args) in /Users/adeal/src/samples-dotnet/src/ActivitySimple/Program.cs:line 15
   at Program.<Main>(String[] args)

```

3. Any docs updates needed?

### DIFF
--- a/src/Temporalio/Bridge/src/client.rs
+++ b/src/Temporalio/Bridge/src/client.rs
@@ -525,7 +525,8 @@ impl TryFrom<&ClientOptions> for CoreClientOptions {
             } else {
                 Some(opts.metadata.to_string_map_on_newlines())
             })
-            .api_key(opts.api_key.to_option_string());
+            .api_key(opts.api_key.to_option_string())
+            .http_connect_proxy(unsafe { opts.http_connect_proxy_options.as_ref() }.map(Into::into));
         if let Some(tls_config) = unsafe { opts.tls_options.as_ref() } {
             opts_builder.tls_cfg(tls_config.try_into()?);
         }


src/Temporalio/Bridge/src/client.rs
```diff
@@ -589,7 +590,7 @@ impl From<&ClientHttpConnectProxyOptions> for HttpConnectProxyOptions {
     fn from(opts: &ClientHttpConnectProxyOptions) -> Self {
         HttpConnectProxyOptions {
             target_addr: opts.target_host.to_string(),
-            basic_auth: if opts.username.size != 0 && opts.password.size != 0 {
+            basic_auth: if (int)opts.username.size != 0 && (int)opts.password.size != 0 {
```
adeal, line 593: not required, but slipped it into this PR.

cretz (approved):

cretz:
We will prioritize working on https://github.com/temporalio/features/issues/521

PR #348, opened 2024-09-15
Improve xml-doc

rstm-sf:
## What was changed
Improve xml-doc

## Why?
When I open the solution in the Rider, it highlights all the rough edges
image

## Checklist

1. Closes 

2. How was this tested:
image

3. Any docs updates needed?

### DIFF
--- a/src/Temporalio/Activities/ActivityInfo.cs
+++ b/src/Temporalio/Activities/ActivityInfo.cs
@@ -55,7 +55,7 @@ namespace Temporalio.Activities
     {
         /// <summary>
         /// Gets the value that is set on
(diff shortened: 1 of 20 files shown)


rstm-sf:
Ooops.... https://github.com/temporalio/sdk-dotnet/pull/306

cretz:
:+1: Yeah, is it ok if we close this one? We may be closing the other one too because we can't really make single-IDE-only fixes unless we have those checks in CI to prevent them from continually reoccurring. Clearly Rider is inconsistent with .NET here.

rstm-sf:
Yes, then this can be closed, I think

PR #355, opened 2024-10-08
Add update ID and name in log scope

### LINKED ISSUE #351: Show update handler and ID in logging context
See https://github.com/temporalio/features/issues/466 for more details

cretz:
## What was changed

Add `UpdateId` and `UpdateName` to the logger scope for the update handler

(features CI will fail until #352 merged)

## Checklist

1. Closes #351


src/Temporalio/Worker/WorkflowInstance.cs
cretz, line 1: Would strongly recommend setting "Hide Whitespace" in the view options here. The only thing we did here is move most update code to another method so we can wrap it easier.

Quinn-With-Two-Ns (approved):

PR #369, opened 2024-11-22
Store `MethodInfo` when creating an `ActivityDefinition`

### LINKED ISSUE #367: [Feature Request] Allow access to optional `MethodInfo` on `ActivityDefinition`

If the `ActivityDefinition` was built via a method, there should be no harm in exposing this method to callers (e.g. via a `MethodInfo?` property). This will allow users to do reflective things like check attributes.

tgrieger:
## What was changed
In order to access other attributes for a given activity, the MethodInfo used to create the activity is needed. There are only certain situations where a MethodInfo is used so the new property is added but made nullable.

## Why?
The main reason is to allow activity interceptors to get reflection information about the method that is being called so it can use things like Attributes to make decisions, if needed.

## Checklist

1. Closes https://github.com/temporalio/sdk-dotnet/issues/367

2. How was this tested:

New tests are added to verify each public Create method properly sets or does not set the MethodInfo.

3. Any docs updates needed?

### DIFF
--- a/tests/Temporalio.Tests/Activities/ActivityDefinitionTests.cs
+++ b/tests/Temporalio.Tests/Activities/ActivityDefinitionTests.cs
@@ -1,7 +1,6 @@
-using System.Reflection;
-
 namespace Temporalio.Tests.Activities;
 
+using System.Reflection;
 using System.Threading.Tasks;
 using Temporalio.Activities;
 using Temporalio.Converters;
--- a/tests/Temporalio.Tests/Activities/ActivityDefinitionTests.cs
+++ b/tests/Temporalio.Tests/Activities/ActivityDefinitionTests.cs
@@ -1,3 +1,5 @@
+using System.Reflection;
+
 namespace Temporalio.Tests.Activities;
 
 using System.Threading.Tasks;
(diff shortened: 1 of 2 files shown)


tests/Temporalio.Tests/Activities/ActivityDefinitionTests.cs
```diff
@@ -1,3 +1,5 @@
+using System.Reflection;
+
 namespace Temporalio.Tests.Activities;
 
 using System.Threading.Tasks;
```
cretz:
```suggestion
namespace Temporalio.Tests.Activities;

using System.Reflection;
using System.Threading.Tasks;
```

Sorry, I know it's pedantic, and we are inconsistent ourselves I've seen across files, but hopefully at least consistent within files

cretz (approved):
This looks great, only a minor suggestion

tests/Temporalio.Tests/Activities/ActivityDefinitionTests.cs
tgrieger: Not a problem, totally missed that it was out of order since it was auto added! Updated.

cretz:
Thanks!

PR #372, opened 2024-11-25
Custom slot suplier

Sushisource:
## What was changed
Add user-implementable slot suppliers

## Why?
Parity. Cool feature.

## Checklist

1. Closes 

2. How was this tested:
Added tests

3. Any docs updates needed?
Will need to follow up with docs for custom interface in all langs


Sushisource:
Let's merge https://github.com/temporalio/sdk-dotnet/pull/376 first then I will rebase

src/Temporalio/Bridge/CustomSlotSupplier.cs
```diff
@@ -0,0 +1,179 @@
...
+        {
+            var handle = GCHandle.FromIntPtr(new IntPtr(tokenSrc));
+            var cancelTokenSrc = (CancellationTokenSource)handle.Target!;
+            cancelTokenSrc.Cancel();
+            handle.Free();
+        }
+
+        private void SafeReserve(Interop.SlotReserveCtx ctx, IntPtr sender)
+        {
+            var reserveTask = Task.Run(async () =>
```
cretz:
```suggestion
            Task.Run(async () =>
```

Were you getting an warning/error here if the variable wasn't assigned? If so, may be able to do `var _` or something (older versions of .NET won't support `_ =` maybe)

src/Temporalio/Bridge/CustomSlotSupplier.cs
```diff
@@ -0,0 +1,179 @@
...
+                            new(ctx), cancelTokenSrc.Token).ConfigureAwait(false);
+                        var usedPermitId = AddPermitToMap(permit);
+                        unsafe
+                        {
+                            Interop.Methods.complete_async_reserve(sender.ToPointer(), new(usedPermitId));
+                        }
+                        cancelTokenSrc.Dispose();
+                        return;
+                    }
+                    catch (OperationCanceledException)
```
cretz:
```suggestion
                    catch (OperationCanceledException) when (cancelTokenSrc.IsCancellationRequested)
```

I assume there will be a missed reservation if a user raised this exception for whatever reason unrelated to our cancellation. In fact, should we either 1) _always_ call `complete_async_reserve` before returning, even on cancel, and/or 2) _always_ avoid calling `complete_async_reserve` if cancellation is requested (user can choose to ignore cancellation token)

src/Temporalio/Bridge/CustomSlotSupplier.cs
```diff
@@ -0,0 +1,179 @@
...
+            var cancelTokenSrc = (CancellationTokenSource)handle.Target!;
+            cancelTokenSrc.Cancel();
+            handle.Free();
+        }
+
+        private void SafeReserve(Interop.SlotReserveCtx ctx, IntPtr sender)
+        {
+            var reserveTask = Task.Run(async () =>
+            {
+                var cancelTokenSrc = new System.Threading.CancellationTokenSource();
```
cretz: Can you use a `using` block here instead of all of the manual `Dispose` calls? Note, you will have to do a `using (...)` + block instead of `using var ...` because of our requirement to support older .NET versions.

src/Temporalio/Bridge/CustomSlotSupplier.cs
```diff
@@ -0,0 +1,179 @@
...
+                }
+                catch (Exception e)
+                {
+                    Console.WriteLine($"Error setting cancel token on ctx: {e}");
+                    throw;
+                }
+            }
+        }
+
+        private unsafe void Reserve(Interop.SlotReserveCtx ctx, void* sender)
```
cretz:
```suggestion
        private unsafe void Reserve(Interop.SlotReserveCtx* ctx, void* sender)
```

I would have expected these all to be pointers so they aren't copied (so `*const SlotReserveCtx` on the Rust side). I think it's clearer that these short-lived Rust-owned things be passed by reference to Rust, so you don't need to re-capture a fixed pointer later.

src/Temporalio/Bridge/CustomSlotSupplier.cs
```diff
@@ -0,0 +1,179 @@
...
+            PinCallbackHolder(interopCallbacks);
+        }
+
+        private static void SetCancelTokenOnCtx(ref SlotReserveCtx ctx, CancellationTokenSource cancelTokenSrc)
+        {
+            unsafe
+            {
+                try
+                {
+                    var handle = GCHandle.Alloc(cancelTokenSrc);
```
cretz: Is this only freed if `CancelReserve` is called? Consider inlining the logic of this method with `Reserve` and freeing this in a `finally`. Once the try/catch is removed and the need for a fixed pointer is removed (see other comments), it's a short method anyways.

EDIT: Looking at Rust code, it appears `CancelReserve` is _always_ called even if cancellation doesn't need to occur. Unsure if it's worth adding that comment somewhere on the .NET side. Regardless, so this is technically fine to free inside of callback, though there is still probably clarity in inlining this method.

src/Temporalio/Bridge/CustomSlotSupplier.cs
```diff
@@ -0,0 +1,179 @@
...
+                    fixed (Interop.SlotReserveCtx* p = &ctx)
+                    {
+                        Interop.Methods.set_reserve_cancel_target(p, GCHandle.ToIntPtr(handle).ToPointer());
+                    }
+                }
+                catch (Exception e)
+                {
+                    Console.WriteLine($"Error setting cancel token on ctx: {e}");
```
cretz:
```suggestion
```

I assume leftover debug code? Can just remove the whole try/catch and let it bubble

src/Temporalio/Bridge/CustomSlotSupplier.cs
```diff
@@ -0,0 +1,179 @@
...
+                        return;
+                    }
+#pragma warning disable CA1031 // We are ok catching all exceptions here
+                    catch (Exception e)
+                    {
+#pragma warning restore CA1031
+                        logger.LogError(e, "Error reserving slot");
+                    }
+                    // Wait for a bit to avoid spamming errors
+                    await Task.Delay(1000, cancelTokenSrc.Token).ConfigureAwait(false);
```
cretz: This and all error handling logic needs to be made clear on the docs of `ICustomSlotSupplier.Reserve`. Basically we need to tell people not to throw from it and the consequences of doing so. Same with other calls, though all of those can just say errors are logged and ignored.

src/Temporalio/Bridge/CustomSlotSupplier.cs
```diff
@@ -0,0 +1,179 @@
...
+            }
+            var usedPermitId = AddPermitToMap(maybePermit);
+            return new(usedPermitId);
+        }
+
+        private void MarkUsed(Interop.SlotMarkUsedCtx ctx)
+        {
+            try
+            {
+                userSupplier.MarkSlotUsed(new(ctx, permits[ctx.slot_permit.ToUInt32()]));
```
cretz: Do not access the `permits` dictionary outside of lock (same for the use in release)

src/Temporalio/Bridge/CustomSlotSupplier.cs
```diff
@@ -0,0 +1,179 @@
...
+            {
+                userSupplier.ReleaseSlot(new(ctx, permits[permitId]));
+            }
+#pragma warning disable CA1031 // We are ok catching all exceptions here
+            catch (Exception e)
+            {
+#pragma warning restore CA1031
+                logger.LogError(e, "Error releasing slot");
+            }
+            permits.Remove(permitId);
```
cretz: This should be in a `finally`

src/Temporalio/Bridge/NativeInvokeableClass.cs
```diff
@@ -0,0 +1,66 @@
...
+        /// Gets the pointer to the native callback holder.
+        /// </summary>
+        internal unsafe T* Ptr { get; private set; }
+
+        /// <summary>
+        /// Pin the native type in memory and add it to the handle list. Call this after adding
+        /// the callbacks via <see cref="FunctionPointer"/>. Also adds `this` to the handle list.
+        /// </summary>
+        /// <param name="value">The native type to pin.</param>
+        internal void PinCallbackHolder(T value)
```
cretz: Technically these calls only invoked by subclasses should be `private protected` but not a big deal

src/Temporalio/Bridge/NativeInvokeableClass.cs
```diff
@@ -0,0 +1,66 @@
...
+using System.Collections.Generic;
+using System.Runtime.InteropServices;
+
+namespace Temporalio.Bridge
+{
+    /// <summary>
+    /// Extend this class to help with making a class that has callbacks which are invoked by Rust.
+    /// </summary>
+    /// <typeparam name="T">The native type that holds the function ptrs for callbacks to C#.</typeparam>
+    internal class NativeInvokeableClass<T>
```
cretz: Technically can/should be abstract, but not a big deal

src/Temporalio/Bridge/src/worker.rs
```diff
@@ -56,27 +63,231 @@ pub struct TunerHolder {
...
+    unsafe extern "C" fn(ctx: SlotReserveCtx, sender: *mut libc::c_void);
+type CustomCancelReserveCallback = unsafe extern "C" fn(token_source: *mut libc::c_void);
+/// Must return C#-tracked id for the permit. A zero value means no permit was reserved.
+type CustomTryReserveSlotCallback = unsafe extern "C" fn(ctx: SlotReserveCtx) -> usize;
+type CustomMarkSlotUsedCallback = unsafe extern "C" fn(ctx: SlotMarkUsedCtx);
+type CustomReleaseSlotCallback = unsafe extern "C" fn(ctx: SlotReleaseCtx);
+
+#[repr(C)]
+#[derive(Clone, Copy)]
+pub struct CustomSlotSupplierCallbacksImpl(*const CustomSlotSupplierCallbacks);
```
cretz, line 109: IMO this needs a `Drop` impl that calls a separate callback called `free` that invokes `Free` on the .NET side similar to how the custom metric meter does.

src/Temporalio/Bridge/Worker.cs
```diff
@@ -16,11 +17,16 @@ internal class Worker : SafeHandle
...
+        ///   anything in the bridge that needs it, since it's guaranteed to be set.</param>
         /// <exception cref="Exception">
         /// If any of the options are invalid including improperly defined workflows/activities.
         /// </exception>
         public Worker(
-            Client client, string namespace_, Temporalio.Worker.TemporalWorkerOptions options)
+            Client client,
+            string namespace_,
+            Temporalio.Worker.TemporalWorkerOptions options,
+            ILoggerFactory loggerFactory)
```
cretz, line 29: This logger factory is already on the client and technically overridable on the worker. So you can probably remove this param and add something like `var loggerFactory = options.LoggerFactory ?? client.Options.LoggerFactory`.

src/Temporalio/Worker/Tuning/SlotReserveContext.cs
```diff
@@ -0,0 +1,57 @@
...
+
+namespace Temporalio.Worker.Tuning
+{
+    /// <summary>
+    /// Context for reserving a slot from a <see cref="ICustomSlotSupplier"/>.
+    /// </summary>
+    /// <remarks>
+    /// WARNING: Custom slot suppliers are currently experimental.
+    /// </remarks>
+    public class SlotReserveContext
```
cretz: I think this class and the other context/info classes should be `record`s with public properties in their constructor and the logic for conversion from a bridge context to this remain with the bridge code. We should allow people to create these things themselves if they want to test their implementations.

src/Temporalio/Worker/Tuning/SlotReserveContext.cs
```diff
@@ -0,0 +1,57 @@
...
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="SlotReserveContext"/> class.
+        /// </summary>
+        /// <param name="ctx">The bridge version of the slot reserve context.</param>
+        internal SlotReserveContext(Temporalio.Bridge.Interop.SlotReserveCtx ctx)
+        {
+            this.SlotType = ctx.slot_type switch
```
cretz: A bit surprised a switch expression works in older .NET (or maybe it doesn't, didn't check CI)

src/Temporalio/Worker/Tuning/ICustomSlotSupplier.cs
```diff
@@ -0,0 +1,68 @@
...
+
+namespace Temporalio.Worker.Tuning
+{
+    /// <summary>
+    /// This class can be implemented to provide custom slot supplier behavior.
+    /// </summary>
+    /// <remarks>
+    /// WARNING: Custom slot suppliers are currently experimental.
+    /// </remarks>
+    public interface ICustomSlotSupplier : ISlotSupplier
```
cretz: Could be discussed whether this is better as an abstract class, but I am fine with this too

src/Temporalio/Worker/Tuning/ICustomSlotSupplier.cs
```diff
@@ -0,0 +1,68 @@
...
+        /// </summary>
+        /// <remarks>
+        /// This method will be called concurrently from multiple threads, so it must be thread-safe.
+        /// </remarks>
+        /// <param name="ctx">The context for slot reservation.</param>
+        /// <param name="cancellationToken">A cancellation token that the SDK may use
+        ///  to cancel the operation.</param>
+        /// <returns>A permit to use the slot which may be populated with your own data.</returns>
+        /// <exception cref="OperationCanceledException">Cancellation requested.</exception>
+        public Task<ISlotPermit> ReserveSlotAsync(SlotReserveContext ctx, CancellationToken cancellationToken);
```
cretz:
```suggestion
        Task<ISlotPermit> ReserveSlotAsync(SlotReserveContext ctx, CancellationToken cancellationToken);
```

Implied

src/Temporalio/Worker/Tuning/ISlotPermit.cs
cretz, line 1: Arguably this could just not exist and we could accept any `object` from the user. Alternatively we could just have a sealed `SlotPermit` that has user data inside it like Java has. I think either would be better than this, because with this there is no easy/default implementation for people who don't care about user data.

src/Temporalio/Worker/Tuning/ISlotSupplier.cs
```diff
@@ -6,10 +6,13 @@ namespace Temporalio.Worker.Tuning
...
     ///
-    /// Currently you cannot implement your own slot supplier, but you can use the provided <see
-    /// cref="FixedSizeSlotSupplier"/> and <see cref="ResourceBasedSlotSupplier"/> slot suppliers.
+    /// Pre-built slot suppliers are available as
+    /// <see cref="FixedSizeSlotSupplier"/> and <see cref="ResourceBasedSlotSupplier"/>.
+    ///
+    /// In order to implement your own slot supplier, you can implement the
+    /// <see cref="ICustomSlotSupplier"/> interface.
     /// </summary>
     public interface ISlotSupplier
```
cretz: If you do change `ICustomSlotSupplier` to an abstract class, this could also be an abstract class with an `internal` constructor thereby making it where people _can't_ accidentally provide an invalid instance.

src/Temporalio/Bridge/CustomSlotSupplier.cs
Sushisource: It won't matter - because if the token is cancelled Rust has already given up waiting for anything to come back through the completion channel, so if something is sent or not, either way it's ignored.

src/Temporalio/Bridge/CustomSlotSupplier.cs
Sushisource: Ah, yeah, that's much nicer, forgot that existed.

src/Temporalio/Bridge/CustomSlotSupplier.cs
Sushisource: Yeah, I was thinking about doing that, makes sense.

src/Temporalio/Bridge/CustomSlotSupplier.cs
Sushisource: Yeah, leftover. Though, interestingly, looks like exceptions thrown from calls into C# from Rust don't bubble. They just silently break.

src/Temporalio/Bridge/Worker.cs
Sushisource: Options aren't accessible on that Client type

src/Temporalio/Bridge/CustomSlotSupplier.cs
Sushisource: Thanks. Rust makes you forget about this stuff sometimes lol

src/Temporalio/Bridge/CustomSlotSupplier.cs
Sushisource: Added comment

src/Temporalio/Bridge/src/worker.rs
Sushisource: Yes, just that was the last thing on my TODO list. Added.

src/Temporalio/Worker/Tuning/ISlotPermit.cs
Sushisource: Yeah, good point about not being able to create something easily. I will go with the Java style way because I don't like the idea of just returning any old object if for no other reason than that it's nice to know what is a permit and what isn't.

There was some reason I had to change this to an interface (it was a class before) that I've forgotten now and I'm sure is going to bite me again, but, I'm sure it's not blocking.

src/Temporalio/Bridge/CustomSlotSupplier.cs
cretz: I am saying `OperationCanceledException` can be thrown from `ReserveSlotAsync` without it being the result of _this_ cancellation token, hence the `when` suggestion here.

But in addition to that, remember that `Box::into_raw(Box::new(tx))` means that Rust will never drop the tx side of the oneshot until it's `Box::from_raw`'d in `complete_async_reserve`. So to not have memory leaks, you probably need to call `complete_async_reserve` even on cancel, but just give it a user permit of 0 or something.

src/Temporalio/Bridge/CustomSlotSupplier.cs
cretz: When Rust panics I have seen that converted to `System.Runtime.InteropServices.SEHException`, but that's about it

src/Temporalio/Bridge/Worker.cs
cretz: My mistake, got my clients mixed up

src/Temporalio/Worker/Tuning/ISlotSupplier.cs
Sushisource: This one doesn't work so much because the built-in impls are records and thus can't extend from a class, only interfaces. 

We could change them to classes, but that's technically breaking, though that's fine at this point and I think is probably worth it.

cretz: :+1: I don't have a strong opinion. I'm ok leaving this as an interface or doing the compat break on the fixed/resource suppliers.

src/Temporalio/Bridge/CustomSlotSupplier.cs
Sushisource: Ah, yeah I get what you mean now

src/Temporalio/Bridge/CustomSlotSupplier.cs
```diff
@@ -0,0 +1,246 @@
...
+                        var srcHandle = GCHandle.Alloc(cancelTokenSrc);
+                        Interop.Methods.set_reserve_cancel_target(
+                            (Interop.SlotReserveCtx*)ctx.ToPointer(),
+                            GCHandle.ToIntPtr(srcHandle).ToPointer());
+                    }
+                    while (true)
+                    {
+                        try
+                        {
+                            ConfiguredTaskAwaitable<Temporalio.Worker.Tuning.SlotPermit> reserveTask;
```
cretz: Never seen this type used. Docs say it's an internal compiler thing. I think it should be `Task<`.

src/Temporalio/Bridge/CustomSlotSupplier.cs
```diff
@@ -0,0 +1,246 @@
...
+        {
+            SafeReserve(new IntPtr(ctx), new IntPtr(sender));
+        }
+
+        // Note that this is always called by Rust, either because the call is cancelled or because
+        // it completed. Therefore the GCHandle is always freed.
+        private unsafe void CancelReserve(void* tokenSrc)
+        {
+            var handle = GCHandle.FromIntPtr(new IntPtr(tokenSrc));
+            var cancelTokenSrc = (CancellationTokenSource)handle.Target!;
```
cretz, line 72: To confirm, this can never be called before `set_reserve_cancel_target` returns and never called after `complete_async_reserve` returns right?

src/Temporalio/Bridge/OptionsExtensions.cs
```diff
@@ -602,10 +607,21 @@ private static Interop.SlotSupplier ToInteropSlotSupplier(
...
+                        tag = Interop.SlotSupplier_Tag.Custom,
+                        custom = new Interop.CustomSlotSupplierCallbacksImpl() { _0 = wrapped.Ptr },
+                    };
+                }
+            }
             else
             {
-                throw new ArgumentException(
-                    "ISlotSupplier must be one of the types provided by the library");
+                throw new ArgumentException("ISlotSupplier is an invalid type");
```
cretz:
```suggestion
                throw new ArgumentException("Unknown slot supplier");
```

src/Temporalio/Worker/Tuning/SlotPermit.cs
```diff
@@ -0,0 +1,33 @@
+namespace Temporalio.Worker.Tuning
+{
+    /// <summary>
+    /// A permit to use a slot for a workflow/activity/local activity task.
```
cretz: Can tell users they can extend this class if they want (unlike what we allowed in Java)

src/Temporalio/Worker/Tuning/SlotPermit.cs
```diff
@@ -0,0 +1,33 @@
...
+    /// </remarks>
+    public class SlotPermit
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="SlotPermit"/> class with no associated data.
+        /// </summary>
+        public SlotPermit()
+        {
+            UserData = null;
+        }
```
cretz, line 18:
```suggestion
        public SlotPermit() => UserData = null;
```

Just an FYI, lots of these single-line methods can be shortened. But this is pedantic/unimportant.

src/Temporalio/Worker/Tuning/SlotReserveContext.cs
```diff
@@ -0,0 +1,26 @@
...
+    /// constructor, only use "with" clauses.
+    /// </remarks>
+    public record SlotReserveContext(
+        SlotType SlotType,
+        string TaskQueue,
+        string WorkerIdentity,
+        string WorkerBuildId,
+        bool IsSticky)
+    {
+    }
```
cretz:
```suggestion
        bool IsSticky);
```

src/Temporalio/Worker/Tuning/CustomSlotSupplier.cs
```diff
@@ -0,0 +1,78 @@
...
+
+namespace Temporalio.Worker.Tuning
+{
+    /// <summary>
+    /// This class can be implemented to provide custom slot supplier behavior.
+    /// </summary>
+    /// <remarks>
+    /// WARNING: Custom slot suppliers are currently experimental.
+    /// </remarks>
+    public abstract class CustomSlotSupplier : SlotSupplier
```
cretz, line 13: I wonder if you have to have a public, parameterless constructor here for accessibility outside the project. The tests package doesn't do a good job of testing this because we set `System.Runtime.CompilerServices.InternalsVisibleTo` to the test project so it can access some visible-to-testing only things (I forget what). May need just a quick confirmation this can be extended outside the project. Can just be some scratch code in `tests/Temporalio.SmokeTests`, doesn't have to be committed.

cretz (commented):
LGTM, only minor things

src/Temporalio/Bridge/CustomSlotSupplier.cs
Sushisource: It won't compile as just `Task`, but I can move the configure await to where it's awaited

src/Temporalio/Bridge/CustomSlotSupplier.cs
Sushisource: It shouldn't be possible, no

src/Temporalio/Bridge/CustomSlotSupplier.cs
cretz: Yeah, assuming it's because of the `unsafe` block, I'd say something like this should work:

```csharp
SlotReserveContext ctx;
unsafe { ctx = ReserveCtxFromBridge((Interop.SlotReserveCtx*)ctx.ToPointer()); }
var permit = await userSupplier.ReserveSlotAsync(ctx, cancelTokenSrc.Token).ConfigureAwait(false);
```

src/Temporalio/Worker/Tuning/CustomSlotSupplier.cs
Sushisource: It appears to work just fine without it. The bench project is easier to use since it doesn't require nupkg mumbojumbo but also doesn't do internals visible.

cretz (approved):

PR #374, opened 2024-11-27
Fix CI after GHA's drop of node16 actions

### LINKED ISSUE #373: [Bug] CI build is broken due to GHA dropping support for node16-based actions
### Describe the bug

GitHub Action’s just started enforcing the long-announced EOL on Node16-based actions. Basically, the latest release of GHA’s Runner agent ignores the fact that an action is requesting node16, and instead just calls node20 anyway. 

Until now, we were relying on GLIBC 2.24 Docker images or runners in very specific places to ensure that the native artifacts included in our official SDK releases would run on some older Linux distros that some of our customers were still using.

However, given that Node20 is incompatible with GLIBC 2.27 or earlier, this no longer works, and our CI workflows are now failing.

mjameswh:
## What changed

- Modified our GHA workflows so that when building the release-grade native artifacts on Linux targets, only the actual Rust compilation step runs inside a GLIBC 2.24-based Docker container; all other operations are performed directly on the runner, which can now be any recent Ubuntu.

  This was required because GHA no longer allows actions based on Node16 (i.e. actions are forced to use Node20 instead), but Node20 requires at least GLIBC 2.28. It is therefore not possible to have a single execution environment that can both execute Node20-based actions, yet produces native libraries that can then execute in the GLIBC 2.24-based environments that some of our users have.

  Fixes #373.


Sushisource (approved):

mjameswh:
@cretz Leaving that to you. Just merge if you are ok with it.

cretz:
Moving this to draft while I test it...

cretz:
Confirmed working after minor YAML change, thanks @mjameswh! Updating branch then merging after CI complete...

PR #376, opened 2024-12-04
Update Core

Sushisource:
Updating core to separate those changes from the PR for slot suppliers https://github.com/temporalio/sdk-dotnet/pull/372

Which I will rebase after this is merged


tests/Temporalio.Tests/Worker/WorkflowWorkerTests.cs
```diff
@@ -1227,7 +1227,7 @@ await ExecuteWorkerAsync<MiscHelpersWorkflow>(
                 // Now query after close to check that is replaying worked
                 var isReplayingValues = await Env.Client.GetWorkflowHandle<MiscHelpersWorkflow>(
                     workflowId).QueryAsync(wf => wf.GetEventsForIsReplaying());
-                Assert.Equal(new[] { false, true }, isReplayingValues);
+                Assert.Equal(new[] { false, true, true }, isReplayingValues);
```
Sushisource, line 1230: This is also due to the evict-on-finish change

src/Temporalio/Worker/WorkflowInstance.cs
```diff
@@ -538,8 +538,25 @@ public WorkflowActivationCompletion Activate(WorkflowActivation act)
                     {
                         // We must set the sync context to null so work isn't posted there
                         SynchronizationContext.SetSynchronizationContext(null);
+                        // TODO: Temporary workaround in lieu of https://github.com/temporalio/sdk-dotnet/issues/375
+                        var sortedJobs = act.Jobs.OrderBy(j =>
```
cretz, line 542: I am a bit surprised we didn't order before. But I assume this order by matches previous Core versions before this update? Meaning we feel fairly confident that replays will work the same on .NET SDK upgrade here?

src/Temporalio/Worker/WorkflowInstance.cs
```diff
@@ -1292,7 +1309,7 @@ await inbound.Value.HandleSignalAsync(new(
             }));
         }
 
-        private void ApplyStartWorkflow(StartWorkflow start)
+        private void ApplyStartWorkflow(InitializeWorkflow init)
```
cretz:
```suggestion
        private void ApplyInitializeWorkflow(InitializeWorkflow init)
```

Pedantic, don't have to change

src/Temporalio/Worker/WorkflowInstance.cs
```diff
@@ -1715,6 +1732,7 @@ public override Task DelayAsync(DelayAsyncInput input)
                         {
                             Seq = seq,
                             StartToFireTimeout = Google.Protobuf.WellKnownTypes.Duration.FromTimeSpan(delay),
+                            Summary = instance.PayloadConverter.ToPayload(string.Empty),
```
cretz:
```suggestion
```

I think we want this to be null instead of a full payload (and we will add if summary is present as part of #359). Is there some kind of check Core-side forcing us to set this?

tests/Temporalio.Tests/Worker/WorkflowWorkerTests.cs
cretz: Took me a sec to grok what was happening here. But now I see that on line 1214 above we issue a post-complete query which now causes full replay where it didn't before

src/Temporalio/Worker/WorkflowInstance.cs
Sushisource: Yes, this doesn't change anything. It matches the previous Core ordering

src/Temporalio/Worker/WorkflowInstance.cs
Sushisource: No, I was confused by something going on in one of the tests and forgot to clean this up

cretz (approved):

PR #377, opened 2024-12-05
Add limit to list workflows

### LINKED ISSUE #332: [Feature Request] `ReplayWorkflowsAsync` should take `limit` parameter
Refer to https://github.com/temporalio/sdk-python/issues/610 - same justification.

In this case only the version using `IAsyncEnumerable` really needs it, but, we should probably add it to both versions for consistency.

Sushisource:
## What was changed
Title

## Why?
Convenience functionality since there's no built-in `take` function or similar in .NET for async iterators

## Checklist

1. Closes https://github.com/temporalio/sdk-dotnet/issues/332

2. How was this tested:
Added test

3. Any docs updates needed?

### DIFF
--- a/src/Temporalio/Client/TemporalClient.Workflow.cs
+++ b/src/Temporalio/Client/TemporalClient.Workflow.cs
@@ -458,6 +458,7 @@ namespace Temporalio.Client
                 // Need to combine cancellation token
                 var rpcOptsAndCancelSource = DefaultRetryOptions(input.Options?.Rpc).
                     WithAdditionalCancellationToken(cancellationToken);
+                var yielded = 0;
                 try
                 {
                     var req = new ListWorkflowExecutionsRequest()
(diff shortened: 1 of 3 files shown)


src/Temporalio/Client/TemporalClient.Workflow.cs
```diff
@@ -472,6 +473,11 @@ private async IAsyncEnumerable<WorkflowExecution> ListWorkflowsInternalAsync(
                             req, rpcOptsAndCancelSource.Item1).ConfigureAwait(false);
                         foreach (var exec in resp.Executions)
                         {
+                            if (input.Options != null && input.Options.Limit > 0 &&
+                                yielded++ >= input.Options.Limit)
```
cretz, line 477: Not the biggest fan of side effects in conditionals like this since postfix increment operator precedence is not always easy for a code reader to understand, but not a blocker.

cretz (approved):

PR #378, opened 2024-12-09
User metadata and workflow metadata support

### LINKED ISSUE #359: [Feature Request] Support user metadata

See https://github.com/temporalio/features/issues/486. Specifically, after https://github.com/temporalio/sdk-core/issues/830:

* Accept optional "static summary " and "static details" on all ways to start a workflow (start, signal with start, schedule, and child workflow) and populate user metadata with them
* Accept optional "summary" when creating timers and populate user metadata with it
* Accept optional "summary" when creating activities and populate user metadata with it
* Handle `__temporal_workflow_metadata` query (see features issue and https://github.com/temporalio/sdk-go/pull/1597)
* Mark experimental

cretz:
## What was changed

* Added `WorkflowOptions.StaticSummary` and `WorkflowOptions.StaticDetails` for setting static summary/details on workflow start, workflow signal with start, and schedule workflow action
* Added `ActivityOptions.Summary` for setting a summary for activity executions
* Added `ChildWorkflowOptions.StaticSummary` and `ChildWorkflowOptions.StaticDetails` for setting summary/details on child workflow start
* Added `Workflow.DelayWithOptionsAsync` and `DelayOptions` for an options-based overload that supports timer `Summary`
  * It was decided to avoid compilation confusion with type inference to name this separately, but it does not affect interceptors
* Added `Workflow.WaitConditionWithOptionsAsync` and `WaitConditionOptions` for an options-based overload that supports timeout timer `Summary`
  * It was decided to avoid compilation confusion to name this separately
* Added `WorkflowExecutionDescription.StaticSummary` and `WorkflowExecutionDescription.StaticDetails` for viewing summary/details on workflow describe
* Added `Workflow.CurrentDetails` property with get/set, so it can be set with a string visible on workflow metadata
* Added `Definition` support for signal, query, and update attributes/definitions
* Added support for `__temporal_workflow_metadata` query in workflow instance

Can ignore commits prior to the one for this PR, they are in main already.

## Checklist

1. Closes #359


src/Temporalio/Worker/Interceptors/DelayAsyncInput.cs
```diff
@@ -8,11 +9,23 @@ namespace Temporalio.Worker.Interceptors
     /// </summary>
     /// <param name="Delay">Delay duration.</param>
     /// <param name="CancellationToken">Optional cancellation token.</param>
+    /// <param name="Summary">Summary for the delay.</param>
```
Sushisource, line 12: Maybe worth saying "which will appear in the workflow history" or something

For new users it might be really nonobvious why a "delay" would have a summary.

src/Temporalio/Workflows/ActivityOptions.cs
```diff
@@ -72,6 +72,13 @@ public class ActivityOptions : ICloneable
         /// </summary>
         public string? TaskQueue { get; set; }
 
+        /// <summary>
+        /// Gets or sets a single-line fixed summary for this activity that may appear in UI/CLI.
+        /// This can be in single-line Temporal markdown format.
+        /// </summary>
+        /// <remarks>WARNING: This setting is experimental.</remarks>
+        public string? Summary { get; set; }
```
Sushisource, line 80: Is this one not static?

src/Temporalio/Workflows/WorkflowQueryAttribute.cs
```diff
@@ -37,6 +37,12 @@ public WorkflowQueryAttribute(string name)
         /// </summary>
         public string? Name { get; }
 
+        /// <summary>
+        /// Gets or sets a short description for this query that may appear in UI/CLI.
```
Sushisource: Maybe worth mentioning that this is returned when asking what queries a workflow supports, and that _that_ is what may appear in UI/CLI

src/Temporalio/Workflows/WorkflowUpdateAttribute.cs
```diff
@@ -41,6 +41,12 @@ public WorkflowUpdateAttribute()
         /// </summary>
         public string? Name { get; }
 
+        /// <summary>
+        /// Gets or sets a short description for this update that may appear in UI/CLI.
```
Sushisource: Same about when this is returned

Sushisource (approved):
Makes sense to me! Not much to think about really. Left a few comment suggestions

src/Temporalio/Worker/Interceptors/DelayAsyncInput.cs
cretz: This is the interceptor field (mostly because .NET linting makes me document it). For the one the user calls, in `DelayOptions`, it states:

> Gets or sets a simple string identifying this timer that may be visible in UI/CLI. While it can be normal text, it is best to treat as a timer ID.

src/Temporalio/Workflows/ActivityOptions.cs
cretz: No, and neither is the timer one. And technically in history, neither is the workflow one. We only call the workflow one `StaticSummary` because we want to use the term `StaticDetails` and to clarify they are immutable/static and differentiate those from the concept of `CurrentDetails` which is mutable. So workflows have "static" summary/details even though they are really just details, but we don't want confused with "current" details. We could not find better terminology for these when we designed them.

PR #381, opened 2024-12-16
Update with start

### LINKED ISSUE #346: UpdateWithStart SDK API
Expose update-with-start operations via the SDK

cretz:
## What was changed

Added `StartUpdateWithStartWorkflowAsync` and `ExecuteUpdateWithStartWorkflowAsync` client calls and all that entails. Marked as experimental.

## Checklist

1. Closes #346

### DIFF
--- a/src/Temporalio.Extensions.OpenTelemetry/TracingInterceptor.cs
+++ b/src/Temporalio.Extensions.OpenTelemetry/TracingInterceptor.cs
@@ -125,6 +125,27 @@ namespace Temporalio.Extensions.OpenTelemetry
         }
 
         /// <summary>
+        /// Create tag collection for the given workflow and update ID.
+        /// </summary>
+        /// <param name="workflowId">Workflow ID.</param>
+        /// <param name="updateId">Update ID.</param>
+        /// <returns>Tags.</returns>
+        protected virtual IEnumerable<KeyValuePair<string, object?>> CreateUpdateTags(
+            string workflowId, string? updateId)
+        {
+            var ret = new List<KeyValuePair<string, object?>>(2);
+            if (Options.TagNameWorkflowId is string wfName)
+            {
+                ret.Add(new(wfName, workflowId));
+            }
+            if (Options.TagNameUpdateId is string updateName && updateId is { } nonNullUpdateId)
+            {
+                ret.Add(new(updateName, nonNullUpdateId));
+            }
+            return ret;
+        }
+
+        /// <summary>
         /// Create tag collection from the current workflow environment. Must be called within a
         /// workflow.
         /// </summary>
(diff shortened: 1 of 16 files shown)


src/Temporalio.Extensions.OpenTelemetry/TracingInterceptor.cs
```diff
@@ -208,6 +229,48 @@ public override async Task<WorkflowHandle<TWorkflow, TResult>> StartWorkflowAsyn
...
+                StartUpdateWithStartWorkflowInput input)
+            {
+                // Ignore if for some reason the start operation is not set by this interceptor
+                if (input.Options.StartWorkflowOperation == null)
+                {
+                    return await base.StartUpdateWithStartWorkflowAsync<TUpdateResult>(input).ConfigureAwait(false);
+                }
+
+                using (var activity = ClientSource.StartActivity(
+                    $"UpdateWithStartWorkflow:{input.Options.StartWorkflowOperation.Workflow}",
```
cretz, line 242: @dandavison - note this differs from Go. In Go, this is `UpdateWithStartWorkflow:<update-name>` whereas here it is `UpdateWithStartWorkflow:<workflow-name>`. This latter behavior matches signal with start in both SDKs (i.e. it's `SignalWithStartWorkflow:<workflow-name>` not `SignalWithStartWorkflow:<signal-name>`), and of course inbound update does get its own span worker side. Not sure we even properly set tracing in any other update-with-start SDK.

src/Temporalio/Client/TemporalClient.Workflow.cs
```diff
@@ -115,6 +155,161 @@ public override async Task<WorkflowHandle<TWorkflow, TResult>> StartWorkflowAsyn
...
+                    if (input.Options.StartWorkflowOperation.Options.Rpc != null)
+                    {
+                        throw new ArgumentException("Cannot set RPC options on start options, set them on the update options");
+                    }
+
+                    // Single atomic to prevent reuse
+                    if (!input.Options.StartWorkflowOperation.TryMarkUsed())
+                    {
+                        throw new ArgumentException("Start operation already used");
+                    }
```
cretz: This is the one condition+exception I should move above the `try` block (well, and maybe the null check of course). Right now reusing the operation could set this exception before the actual exception/result of the original use of the operation. Will fix...

src/Temporalio/Client/TemporalClient.Workflow.cs
```diff
@@ -115,6 +155,161 @@ public override async Task<WorkflowHandle<TWorkflow, TResult>> StartWorkflowAsyn
...
+                                null,
+                                runId,
+                                runId)!);
+                        }
+                        if (resp.Responses.Count > 1)
+                        {
+                            updateResp = resp.Responses[1].UpdateWorkflow;
+                        }
+                    }
+                    while (updateResp == null || (
```
cretz:
```suggestion
                        updateResp = resp.Responses[1].UpdateWorkflow;
                    }
                    while (
```

Looking at the Python PR, this response is always present even if the update is not yet done

src/Temporalio/Client/TemporalClient.Workflow.cs
```diff
@@ -115,6 +155,161 @@ public override async Task<WorkflowHandle<TWorkflow, TResult>> StartWorkflowAsyn
...
+                        updateResp.Stage < (UpdateWorkflowExecutionLifecycleStage)input.Options.WaitForStage &&
+                        updateResp.Stage < UpdateWorkflowExecutionLifecycleStage.Accepted));
```
dandavison: What's the purpose of having the user `waitForStage` in these conditions? I think we should simplify them to `while < Accepted`. I'm going to suggest the same thing for Python also. If/when we get durable-on-admitted, we'll have to do a pass over the SDKs and we can make the conditions more complex then.

src/Temporalio/Client/TemporalClient.Workflow.cs
```diff
@@ -115,6 +155,161 @@ public override async Task<WorkflowHandle<TWorkflow, TResult>> StartWorkflowAsyn
...
+                        input.Headers).ConfigureAwait(false);
+                    updateReq.WorkflowExecution = new() { WorkflowId = startReq.WorkflowId };
+                    var req = new ExecuteMultiOperationRequest() { Namespace = Client.Options.Namespace };
+                    req.Operations.Add(
+                        new ExecuteMultiOperationRequest.Types.Operation() { StartWorkflow = startReq });
+                    req.Operations.Add(
+                        new ExecuteMultiOperationRequest.Types.Operation() { UpdateWorkflow = updateReq });
+
+                    // Continually try to start until an exception occurs, the user-asked stage is
+                    // reached, or the stage is accepted. But we will set the workflow handle as soon as
```
dandavison, line 216: See comment below, I think `while < Accepted` is all that's needed.

dandavison (approved):
LGTM, I didn't see anything missing or out of line with the plan.

src/Temporalio/Client/ITemporalClient.Workflow.cs
```diff
@@ -92,6 +92,84 @@ WorkflowHandle<TWorkflow> GetWorkflowHandle<TWorkflow>(
         WorkflowHandle<TWorkflow, TResult> GetWorkflowHandle<TWorkflow, TResult>(
             string id, string? runId = null, string? firstExecutionRunId = null);
 
+        /// <summary>
+        /// Start an update via a call to a WorkflowUpdate attributed method, possibly starting the
```
dandavison, line 96: "via a call to a WorkflowUpdate attributed method" means "this will end up causing an update handler method to be executed in a workflow"?

src/Temporalio/Client/TemporalClient.Workflow.cs
cretz: Sure, it makes no real difference at this time, agreed (it was leftover from older update logic where we considered returning immediately for "admitted" IIRC).

src/Temporalio/Client/ITemporalClient.Workflow.cs
cretz: Not exactly. What "via a call" means is that this is a lambda expression that makes a call to a method with the WorkflowUpdate attribute. So it looks like `myClient.StartUpdateWithStartWorkflow((MyWorkflow wf) => wf.MyUpdateMethodWithAttribute(myArg), ...`.

src/Temporalio/Client/TemporalClient.Workflow.cs
cretz: Fixed

src/Temporalio/Client/TemporalClient.Workflow.cs
cretz: Fixed

src/Temporalio/Client/TemporalClient.Workflow.cs
cretz: Updated here and in regular update logic

src/Temporalio/Client/TemporalClient.Workflow.cs
cretz: Updated here and in regular update logic

PR #382, opened 2024-12-19
Minor update fixes

cretz:
## What was changed

* Fix issue where proper run ID wasn't set on update handle
* Fix issue where null and OK statuses weren't being ignored when obtaining multi-op failure

### DIFF
--- a/src/Temporalio/Client/TemporalClient.Workflow.cs
+++ b/src/Temporalio/Client/TemporalClient.Workflow.cs
@@ -253,7 +253,8 @@ namespace Temporalio.Client
                 }
                 catch (Exception e)
                 {
-                    // If this is a multi-operation failure, set exception to the first non-aborted
+                    // If this is a multi-operation failure, set exception to the first present,
+                    // non-OK, non-aborted error
                     if (e is RpcException rpcErr)
                     {
                         var status = rpcErr.GrpcStatus.Value;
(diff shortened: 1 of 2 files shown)


src/Temporalio/Client/TemporalClient.Workflow.cs
```diff
@@ -462,7 +472,7 @@ public async override Task<WorkflowUpdateHandle<TResult>> StartWorkflowUpdateAsy
                 // If the requested stage is completed, wait for result, but discard the update
                 // exception, that will come when _they_ call get result
                 var handle = new WorkflowUpdateHandle<TResult>(
-                    Client, req.Request.Meta.UpdateId, input.Id, input.RunId)
+                    Client, req.Request.Meta.UpdateId, input.Id, resp.UpdateRef.WorkflowExecution.RunId)
```
dandavison, line 475: We should really add a test for this. (I know the other fix is hard to test).

src/Temporalio/Client/TemporalClient.Workflow.cs
```diff
@@ -253,19 +253,29 @@ public override async Task<WorkflowUpdateHandle<TUpdateResult>> StartUpdateWithS
...
                         if (status != null && status.Details.Count == 1)
                         {
                             if (status.Details[0].TryUnpack(out Api.ErrorDetails.V1.MultiOperationExecutionFailure failure))
                             {
-                                var nonAborted = failure.Statuses.FirstOrDefault(s => s.Details.Count == 0 ||
-                                    !s.Details[0].Is(Api.Failure.V1.MultiOperationExecutionAborted.Descriptor));
-                                var grpcStatus = new GrpcStatus() { Code = nonAborted.Code, Message = nonAborted.Message };
-                                grpcStatus.Details.AddRange(nonAborted.Details);
-                                e = new RpcException(grpcStatus);
+                                var nonAborted = failure.Statuses.FirstOrDefault(s =>
```
dandavison, line 265: variable name could be `failureStatus` or something, but not blocking.

dandavison (approved):
LGTM. I'm going to try to add a test for the run ID issue in Python.

src/Temporalio/Client/TemporalClient.Workflow.cs
cretz: Added assertion to existing test that would have failed before (basically just ensures run ID now exists, because it was null before)

PR #384, opened 2025-01-02
Run event loop after job application

### LINKED ISSUE #375: [Feature Request] Align activation job application with TS changes
.NET currently applies jobs without reordering them (or, as I'm writing this, potentially will as a temporary workaround until we address this), but wakes up the event loop after each job.

In TS, we changed this to apply jobs to state and then run the event loop: https://github.com/temporalio/sdk-typescript/pull/1488

There is a really nice test [here](https://github.com/temporalio/sdk-typescript/pull/1488/files#diff-edcae2d23017bc07e3a859b68e980d151947323c04128f2bc08147bda4a1d7deR2512) @mjameswh added that I think all async languages should have some form of that helps us understand exactly how routines will proceed.

I'm not sure if making this change will actually have consequences to the _realized_ .NET behavior. In any case it should probably be protected with a flag if we make it.

See https://github.com/temporalio/sdk-python/issues/606 for python equivalent

Sushisource:
## What was changed
Run event loop once after applying jobs, rather than after each job

## Why?
Align with TS & Core's intended method of job application

## Checklist

1. Closes https://github.com/temporalio/sdk-dotnet/issues/375

2. How was this tested:
Existing & new tests

3. Any docs updates needed?

### DIFF
--- a/src/Temporalio/Worker/WorkflowInstance.cs
+++ b/src/Temporalio/Worker/WorkflowInstance.cs
@@ -541,31 +541,18 @@ namespace Temporalio.Worker
                     {
                         // We must set the sync context to null so work isn't posted there
                         SynchronizationContext.SetSynchronizationContext(null);
-                        // TODO: Temporary workaround in lieu of https://github.com/temporalio/sdk-dotnet/issues/375
-                        var sortedJobs = act.Jobs.OrderBy(j =>
-                        {
-                            switch (j.VariantCase)
-                            {
-                                case WorkflowActivationJob.VariantOneofCase.NotifyHasPatch:
-                                case WorkflowActivationJob.VariantOneofCase.UpdateRandomSeed:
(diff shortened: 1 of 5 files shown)


Sushisource:
Unforunately, unlike python, this one cannot be done without a flag because the loop is run after every job rather than after each batch.

This means that, ex, two signals that came in and both have more than one immediately resolved await point will execute through those await points completely before moving to the next signal, rather than interleaving their wakeups.

Because of that, I'm not sure this change is worth it, unless we think we could actually remove the flag after some number of years or whatever.

cretz:
Concur it may not be worth it

Sushisource:
We won't bother with this. This PR is kept referenced in the ticket so that it might be used if we later introduce a flag for fixing wait-condition behavior.

cretz:
Deleting branch

PR #399, opened 2025-02-03
Added namespace name to workflow/activity tags

petrkoutnycz:
## What was changed
OpenTelemetry tags were extended with a namespace name.

## Why?
Motivation is to get a namespace name as part of OTel data so we can do some grouping based on that in our observability tool.

### DIFF
--- a/src/Temporalio.Extensions.OpenTelemetry/TracingInterceptor.cs
+++ b/src/Temporalio.Extensions.OpenTelemetry/TracingInterceptor.cs
@@ -163,6 +163,10 @@ namespace Temporalio.Extensions.OpenTelemetry
             {
                 ret.Add(new(runName, info.RunId));
             }
+            if (Options.TagNameNamespace is string namespaceName)
+            {
+                ret.Add(new(namespaceName, info.Namespace));
+            }
             return ret;
         }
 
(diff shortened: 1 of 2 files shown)


src/Temporalio.Extensions.OpenTelemetry/TracingInterceptor.cs
```diff
@@ -163,6 +163,10 @@ protected virtual IDictionary<string, Payload> HeadersFromContext(
             {
                 ret.Add(new(runName, info.RunId));
             }
+            if (Options.TagNameNamespace is string namespaceName)
+            {
+                ret.Add(new(namespaceName, info.Namespace));
+            }
```
cretz, line 169: We match our other SDKs like Go and Python with the default tag names which doesn't include `temporalNamespace`. But we made these `protected virtual` on purpose so anyone can extend the interceptor, override these, and set any additional tags they'd like.

petrkoutnycz: But there can be a plan to make it work cross-SDK...right? 😈

cretz: We have opened https://github.com/temporalio/features/issues/592. If we decide to do it, yes, but in the meantime this can be done today extending the class.

PR #407, opened 2025-02-05
Update Core & add HTTP Otel / Dev UI Port

### LINKED ISSUE #392: [Feature Request] OTel metrics over HTTP

Expose ability to have Core OTel metrics use HTTP instead of default gRPC protocol. Would recommend a simple default-false http boolean. See https://github.com/temporalio/sdk-core/issues/820.

### LINKED ISSUE #396: [Feature Request] Allow customization of dev server UI port

See https://github.com/temporalio/features/issues/588

Sushisource:
## What was changed
* Added ability to specify UI port in dev server startup
* Added ability to select HTTP protocol for Otel metrics export

## Checklist

1. Closes https://github.com/temporalio/sdk-dotnet/issues/396 
    closes https://github.com/temporalio/sdk-dotnet/issues/392

3. How was this tested:
Core tests should be sufficient. Manually verified UI port is set properly.

4. Any docs updates needed?


src/Temporalio/Bridge/src/testing.rs
```diff
@@ -208,6 +209,11 @@ impl TryFrom<&DevServerOptions> for ephemeral_server::TemporalDevServerConfig {
             .port(test_server_options.port())
             .db_filename(options.database_filename.to_option_string())
             .ui(options.ui)
+            .ui_port(if options.ui_port == 0 {
```
cretz:
```suggestion
            .ui_port(if options.ui_port == 0 || !options.ui {
```

I wonder if we need this here

src/Temporalio/Runtime/OpenTelemetryProtocol.cs
cretz, line 1: Not required, but could make this a nested enum inside the `OpenTelemetryOptions` as just `Protocol` if you'd like, but here is fine too

cretz (commented):
Nice, thanks for knocking out some of my issues while here. Only a couple of minor comments.

src/Temporalio/Runtime/OpenTelemetryProtocol.cs
Sushisource: I stuck with what temporality did

src/Temporalio/Worker/WorkflowCodecHelper.cs
```diff
@@ -174,29 +215,19 @@ await EncodeAsync(
                     await EncodeAsync(
                         codec, cmd.SignalExternalWorkflowExecution.Headers).ConfigureAwait(false);
                     break;
+                case WorkflowCommand.VariantOneofCase.ScheduleNexusOperation:
+                    await EncodeAsync(
+                        codec, cmd.ScheduleNexusOperation.Input).ConfigureAwait(false);
```
cretz: Can we surround this with the if-not-null check? Hopefully as of #405 CI will break anyways. (EDIT: it does break in CI, whew)

Sushisource: Will do. IMO these should all just check for nullity internally. Feels a bit silly to protect all of them individually.

cretz: I'd be ok with that change if you want to make it or open an issue for us to make it later. I think originally I was unnecessarily trying to reduce call counts.

omnisharp.json
cretz, line 1: Hrmm, do we have to have this as a new top-level file in the repo? We can if we must, though ideally we can avoid specific editor settings in the repo.

cretz (approved):
Approving. Don't like the new top-level file, but not a big deal.

PR #410, opened 2025-02-09
Reserve __temporal prefix

### LINKED ISSUE #391: [Feature Request] Special behavior for Temporal built-in prefixes

See https://github.com/temporalio/features/issues/576

Sushisource:
## What was changed
See https://github.com/temporalio/features/issues/576

## Why?
Future flexibility for us to implement built-in handlers

## Checklist

1. Closes https://github.com/temporalio/sdk-dotnet/issues/391

2. How was this tested:
New tests & test upgrades

3. Any docs updates needed?


.editorconfig
```diff
@@ -47,6 +47,9 @@ dotnet_diagnostic.CA2237.severity = none
 # Warn on unused imports
 dotnet_diagnostic.IDE0005.severity = warning
 
+# Don't warn on using var
+dotnet_diagnostic.IDE0008.severity = none
```
cretz, line 51: If you remember, am a bit curious where in code this was triggered

src/Temporalio/Worker/WorkflowInstance.cs
```diff
@@ -2450,5 +2462,51 @@ public record Handler(
                 string? UpdateId,
                 HandlerUnfinishedPolicy UnfinishedPolicy);
         }
+
+        /// <summary>
+        /// This interceptor ensures that handlers using the reserved prefix bypass user
+        /// interceptors.
+        /// </summary>
+        private class InboundReservedPrefixInterceptor : WorkflowInboundInterceptor
```
cretz: I do not believe this should be implemented as an interceptor/wrapper that now becomes the outer interceptor always just for this small reason. I believe in `ApplyQueryWorkflow` we need to just make the two reserved queries call query logic directly without going through the interceptor. And it is there we should fail if it's reserved and not one of those. Can also fail update and signal in their apply-level places too.

src/Temporalio/Workflows/WorkflowDefinition.cs
```diff
@@ -427,6 +428,35 @@ public static WorkflowDefinition Create(
...
+                            {
+                                errs.Add($"{type} handler name {key} cannot start with {reservedQ}");
+                            }
+                        }
+                    }
+                }
+            }
+            CheckNamesForPrefix(signals.Keys, "Signal");
+            CheckNamesForPrefix(queries.Keys, "Query");
+            CheckNamesForPrefix(updates.Keys, "Update");
```
cretz: I think you can just check signal name prefix in signal definition, update in update definition, and query in query definition. It is clearer to keep validation with the definition. Can just do it in those constructors if you want.

src/Temporalio/Common/Constants.cs
```diff
@@ -0,0 +1,13 @@
...
+{
+    /// <summary>
+    /// Constants needed by the Temporal SDK.
+    /// </summary>
+    internal static class Constants
+    {
+        /// <summary>
+        /// Prefix for reserved handler and definition names.
+        /// </summary>
+        internal const string ReservedNamePrefix = "__temporal";
```
cretz: If you can think of any other class, would be ok moving there (e.g. even `TemporalRuntime` since it's internal) instead of a whole new class/file for a single constant.

src/Temporalio/Workflows/Workflow.cs
```diff
@@ -18,6 +18,15 @@ namespace Temporalio.Workflows
     /// </summary>
     public static class Workflow
     {
+        /// <summary>
+        /// All known reserved query handler prefixes.
+        /// </summary>
+        internal static readonly string[] ReservedQueryHandlerPrefixes =
```
cretz: Would move to `WorkflowQueryDefinition`

src/Temporalio/Common/Constants.cs
cretz: _Technically_ `__temporal_`, but I am ok with this without trailing underscore

.editorconfig
Sushisource: All over. I think it's just a matter of how the default "low" severity shows up in different lang servers. All the low severity stuff shows up in omnisharp, which can be kinda nice to auto fix some stuff sometimes, but, this one is obviously a little silly.

src/Temporalio/Common/Constants.cs
Sushisource: I think we should probably actually make it specifically be without that. Seems very odd you could register `__temporalmything` but not `__temporal_mything`

Sushisource: Well there's always going to be something that'd be the first reason to have a common place for constants - no where else felt like an accurate fit, but I can move it to Runtime.

src/Temporalio/Worker/WorkflowInstance.cs
Sushisource: Sure, I did this because it would be robust in the face of different ways to handle built-in operations. If you're not worried about that then I'll get rid of it.

src/Temporalio/Workflows/WorkflowDefinition.cs
Sushisource: The downside of this is now you need a specific bypassing constructor -- see my last commit

src/Temporalio/Worker/WorkflowInstance.cs
```diff
@@ -1140,13 +1145,13 @@ private void ApplyQueryWorkflow(QueryWorkflow query)
...
                         Func<string> getter = GetStackTrace;
-                        queryDefn = WorkflowQueryDefinition.CreateWithoutAttribute(
+                        queryDefn = WorkflowQueryDefinition.CreateWithoutAttributeReservedName(
                             "__stack_trace", getter);
                     }
                     else if (query.QueryType == "__temporal_workflow_metadata")
                     {
                         Func<Api.Sdk.V1.WorkflowMetadata> getter = GetWorkflowMetadata;
-                        queryDefn = WorkflowQueryDefinition.CreateWithoutAttribute(
+                        queryDefn = WorkflowQueryDefinition.CreateWithoutAttributeReservedName(
```
cretz: For these two, we need to not go through interceptors to invoke them. Would be ok if they aren't even "query definitions" and just invoked directly or something (also saves you from having that bypass constructor param for query definition)

src/Temporalio/Workflows/WorkflowQueryDefinition.cs
```diff
@@ -10,11 +11,31 @@ namespace Temporalio.Workflows
...
+            "__enhanced_stack_trace",
+        };
+
         private static readonly ConcurrentDictionary<MethodInfo, WorkflowQueryDefinition> MethodDefinitions = new();
         private static readonly ConcurrentDictionary<PropertyInfo, WorkflowQueryDefinition> PropertyDefinitions = new();
 
-        private WorkflowQueryDefinition(string? name, string? description, MethodInfo? method, Delegate? del)
+        private WorkflowQueryDefinition(string? name, string? description, MethodInfo? method, Delegate? del, bool bypassReserved = false)
         {
+            if (!bypassReserved && name != null)
```
cretz: If you want to save lines, add `using System.Linq;` import and then:

```suggestion
            if (!bypassReserved && name != null && ReservedQueryHandlerPrefixes.Any(p => name.StartsWith(p))
```

src/Temporalio/Workflows/WorkflowQueryDefinition.cs
```diff
@@ -10,11 +11,31 @@ namespace Temporalio.Workflows
     /// </summary>
     public class WorkflowQueryDefinition
     {
+        /// <summary>
+        /// All known reserved query handler prefixes.
+        /// </summary>
+        internal static readonly string[] ReservedQueryHandlerPrefixes =
```
cretz:
```suggestion
        private static readonly string[] ReservedQueryHandlerPrefixes =
```

Doesn't matter much though

src/Temporalio/Common/Constants.cs
cretz: Yeah, was unsure here because in spec I added underscore not thinking and other SDKs already merged theirs, though I doubt an inconsistency here matters.

src/Temporalio/Workflows/WorkflowQueryDefinition.cs
Sushisource: Don't save all that much if you want to keep the prefix name in the error which I do, but, changed.

src/Temporalio/Workflows/WorkflowQueryDefinition.cs
```diff
@@ -10,11 +12,26 @@ namespace Temporalio.Workflows
...
+            TemporalRuntime.ReservedNamePrefix,
+            "__stack_trace",
+            "__enhanced_stack_trace",
+        };
+
         private static readonly ConcurrentDictionary<MethodInfo, WorkflowQueryDefinition> MethodDefinitions = new();
         private static readonly ConcurrentDictionary<PropertyInfo, WorkflowQueryDefinition> PropertyDefinitions = new();
 
-        private WorkflowQueryDefinition(string? name, string? description, MethodInfo? method, Delegate? del)
+        private WorkflowQueryDefinition(string? name, string? description, MethodInfo? method, Delegate? del, bool bypassReserved = false)
```
cretz: Can you now remove this constructor parameter for bypassing since it is never true anymore?

cretz (commented):
LGTM, one minor change about removing constructor param

src/Temporalio/Workflows/WorkflowQueryDefinition.cs
Sushisource: Woops, thanks

src/Temporalio/Runtime/TemporalRuntime.cs
```diff
@@ -13,6 +13,11 @@ namespace Temporalio.Runtime
     /// </remarks>
     public sealed class TemporalRuntime
     {
+        /// <summary>
+        /// Prefix for reserved handler and definition names.
+        /// </summary>
+        internal const string ReservedNamePrefix = "__temporal";
```
Quinn-With-Two-Ns, line 19: Shouldn't it be `__temporal_` per https://github.com/temporalio/features/issues/576

Sushisource: See up here: https://github.com/temporalio/sdk-dotnet/pull/410#discussion_r1949907394

Yeah, that's what's written there, but, also that just seems wrong. Probably we should change that spec.

cretz (approved):

PR #431, opened 2025-03-18
add protected constructors for metric instrumentation in Temporalio.Common namespace

### LINKED ISSUE #430: [Feature Request] Allow metric meters to be constructed by users for tests

Today in `Temporalio.Testing.ActivityEnvironment` a `MetricMeter` is accepted which is an abstract class, but the returned metrics from abstract methods there cannot be user constructed due to having `internal` constructors. We should make `protected` constructors on each that accept `string name, string? unit, string? description` that just delegate to the `internal` ones using `MetricDetails`.

robcao:
## What was changed

Added new `protected` constructors for `MetricCounter`, `MetricGauge`, and `MetricHistogram` in the `Temporalio.Common` namespace. This will allow users to supply their own implementations when passing in a custom `MetricMeter` on the activity environment for tests.

Previously the only constructors for these classes were internal, so they were not instantiable from outside of the `Temporalio` library.
## Why?

Users want to be able to test their metric instrumentation within an activity using the `ActivityEnvironment` abstraction, without having to run their activities inside a full `TemporalWorker`.

It's fairly common to use an in-memory exporter to validate metric data points: https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Exporter.InMemory/README.md, so users want to be able to have the ability to supply a custom metric meter hooked up to an in memory exporter to write tests validating metrics.

## Checklist

1. Closes https://github.com/temporalio/sdk-dotnet/issues/430

2. How was this tested:

No logic was added, just an additional constructor. It would be difficult to write a test for this without creating a new test assembly where internals are not visible.

In C#, protected constructors are guaranteed to be accessible by any derived class outside of the assembly: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/accessibility-levels

3. Any docs updates needed?

I updated the section on activity testing to inform users of the presence of the `MetricMeter` on the `ActivityEnvironment`, to be consistent with the rest of the public properties.

I can re-order the properties in this list to be alphabetical, but they currently roughly appear to be in the order as the properties appear on the ActivityExecutionContext, so placing the property at the bottom to be roughly consistent

### DIFF
--- a/README.md
+++ b/README.md
@@ -1115,6 +1115,7 @@ activity context:
 * `Heartbeater` - Callback invoked each heartbeat.
 * `WorkerShutdownTokenSource` - Token source for issuing worker shutdown.
 * `PayloadConverter` - Defaulted to default payload converter.
+* `MetricMeter` - Defaulted to noop meter.
 
 ### OpenTelemetry Tracing Support
 
--- a/src/Temporalio/Common/MetricCounter.cs
+++ b/src/Temporalio/Common/MetricCounter.cs
@@ -19,6 +19,17 @@ namespace Temporalio.Common
         }
 
         /// <summary>
+        /// Initializes a new instance of the <see cref="MetricCounter{T}" /> class.
+        /// </summary>
+        /// <param name="name">The name of the counter.</param>
+        /// <param name="unit">The optional unit of measurement for the values recorded by the counter.</param>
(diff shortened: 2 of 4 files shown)


cretz (approved):
This is great, thanks! Will merge when CI passes.

PR #432, opened 2025-03-18
Apply modern event loop algorithm with new SDK flag

### LINKED ISSUE #375: [Feature Request] Align activation job application with TS changes

### LINKED ISSUE #427: [Bug] Replay after UnhandledCommand can cause main workflow body to complete before signals are handled

I have a workflow that continues as new, and also receives signals. We see that sometimes when the workflow tries to continue as new while new signals come in for that workflow, the continue as new output from the workflow does not contain all of the signals

### Describe the bug

The problem is that the main workflow function is completing before all of the signal handlers run. Admittedly, I am not certain if this is a bug in the SDK or in the workflow implementation, but if it is the latter, this is very unintuitive behavior.

The problem happens when multiple signals come in when the workflow sends the continue as new command back to the server. 

This causes the workflow to replay after receiving all of the new signals, but wait conditions are being checked after every individual signal handlers runs.

This can cause the following sequence of events.

- Workflow main function is blocked on a wait condition
- First signal handler runs
- Workflow main function is now unblocked, workflow main function runs to completion
- Second signal handler runs
  - Because the workflow main function is already completed, the results of this handler will not be included in scenarios like Continue as New

### Minimal Reproduction

I created a repository here with the reproduction steps. There is a README file with instructions.

https://github.com/robcao/temporal-dotnet-missing-signal-sample

For ease of viewing, here is the workflow definition:

What appears to be happening is after the first signal handler runs, both of the wait conditions in the main function body are re-evaluated, both evaluate to true, and so the main function body runs before the second signal handler:

```csharp
await Workflow.WhenAnyAsync(Workflow.DelayAsync(TimeSpan.FromSeconds(5)), Workflow.WaitConditionAsync(() => Signals.Count > 0));
await Workflow.WaitConditionAsync(() => Workflow.AllHandlersFinished);
```

```csharp
[Workflow]
public class SleepThenReturn
{
	internal List<string> Signals { get; init; } = new();

	[WorkflowRun]
	public async Task<string[]> RunAsync(string[] input)
	{
		if (!string.IsNullOrWhiteSpace(Workflow.Info.ContinuedRunId))
		{
			Workflow.Logger.LogInformation("Now continuing as new, there are {count} signals.", input.Length);

			return input;
		}

		await Workflow.WhenAnyAsync(Workflow.DelayAsync(TimeSpan.FromSeconds(5)), Workflow.WaitConditionAsync(() => Signals.Count > 0));

		await Workflow.WaitConditionAsync(() => Workflow.AllHandlersFinished);

		List<string> next = new();

		foreach (string signal in Signals)
		{
			next.Add(signal);
		}

		throw Workflow.CreateContinueAsNewException<SleepThenReturn>(wf => wf.RunAsync(next.ToArray()));
	}

	[WorkflowSignal]
	public Task SendSignal(string signal)
	{
		Workflow.Logger.LogInformation("Handling signal input {signal}.", signal);
		Signals.Add(signal);
		return Task.CompletedTask;
	}
}

```

### Environment/Versions

- OS and processor: x64 Windows
- Temporal Version: 1.5.0
- Are you using Docker or Kubernetes or building Temporal from source? Running directly on Windows

### Additional context

It seems like what we desire here is for handlers to be marked in progress earlier than they currently are, before the event loop runs, perhaps somewhere here: https://github.com/temporalio/sdk-dotnet/blob/main/src/Temporalio/Worker/WorkflowInstance.cs#L562

Another user noticed similar behavior to what we see in this Slack thread: https://temporalio.slack.com/archives/C012SHMPDDZ/p1738636421721339

cretz:
## What was changed

(description as mostly taken from a comment above the SDK flag)

Before this flag, workflow async logic did the following (in no particular order):
* Reorder jobs itself
* Schedule primary workflow method when the initialize job was seen
* Tick the event loop after each job processed
* Check all conditions at once and resolve all that are satisfied

We have since learned that all of these are wrong in one way or another. So if this flag is set on the first activation/task of a workflow, the following logic now applies respectively:
* Leave Core's job ordering as is
* Do not schedule the primary workflow method until after all jobs have been applied and it's not already present.
* Tick the event loop only once after everything applied
* Only resolve the first condition that is satisfied and re-run the entire event loop

We know this solves at least #427 where all-handlers-finished was improperly assumed true too early.

## Checklist

1. Closes #427
2. Closes #375

### DIFF
--- a/src/Temporalio/Worker/WorkflowInstance.cs
+++ b/src/Temporalio/Worker/WorkflowInstance.cs
@@ -85,6 +85,8 @@ namespace Temporalio.Worker
         private WorkflowQueryDefinition? dynamicQuery;
         private WorkflowSignalDefinition? dynamicSignal;
         private WorkflowUpdateDefinition? dynamicUpdate;
+        private bool workflowInitialized;
+        private bool applyModernEventLoopLogic;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="WorkflowInstance"/> class.
@@ -529,6 +531,20 @@ namespace Temporalio.Worker
                 IsReplaying = act.IsReplaying;
                 UtcNow = act.Timestamp.ToDateTime();
 
+                // If the workflow has not been initialized, we are in the first activation and we
+                // need to set the modern-event-loop-logic flag
+                if (!workflowInitialized)
+                {
+                    // If we're not replaying or we are replaying and the flag is already set, set
+                    // to true and mark flag. Otherwise we leave false.
+                    if (!IsReplaying ||
+                        act.AvailableInternalFlags.Contains((uint)WorkflowLogicFlag.ApplyModernEventLoopLogic))
+                    {
+                        applyModernEventLoopLogic = true;
+                        completion.Successful.UsedInternalFlags.Add((uint)WorkflowLogicFlag.ApplyModernEventLoopLogic);
+                    }
+                }
+
                 // Starting callback
                 onTaskStarting(this);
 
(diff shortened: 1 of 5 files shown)


src/Temporalio/Worker/WorkflowInstance.cs
```diff
@@ -541,29 +557,59 @@ public WorkflowActivationCompletion Activate(WorkflowActivation act)
...
+                            }).ToList();
+                        }
+
+                        // Apply each job
+                        foreach (var job in jobs)
                         {
                             Apply(job);
-                            // Run scheduler once. Do not check conditions when patching or querying.
-                            var checkConditions = job.NotifyHasPatch == null && job.QueryWorkflow == null;
+                            // We only run the scheduler after each job in legacy event loop logic
```
Sushisource, line 591:
```suggestion
                            // We run the scheduler after each job in legacy event loop logic
```

Sushisource (approved):

src/Temporalio/Worker/WorkflowInstance.cs
cretz: Meant to say "we only do X in legacy" to clarify we don't do it in non-legacy

src/Temporalio/Worker/WorkflowInstance.cs
```diff
@@ -541,29 +557,59 @@ public WorkflowActivationCompletion Activate(WorkflowActivation act)
...
+                            if (!applyModernEventLoopLogic)
```
THardy98, line 592: Presumably, this can't change within an activation right? If so, can we bring this outside the loop over the jobs instead of checking for each job

src/Temporalio/Worker/WorkflowInstance.cs
```diff
@@ -541,29 +557,59 @@ public WorkflowActivationCompletion Activate(WorkflowActivation act)
...
+                        // For modern event loop logic, we initialize here if not initialized
+                        // already
+                        if (applyModernEventLoopLogic && !workflowInitialized)
+                        {
+                            InitializeWorkflow();
+                        }
+
+                        // For modern event loop logic, we run the event loop only after applying
+                        // everything, and we check conditions if there are any non-query jobs
+                        if (applyModernEventLoopLogic)
```
THardy98, line 610: nit:
```
// For modern event loop logic
if (applyModernEventLoopLogic) {
    // We initialize here if not initialized already
    if (!workflowInitialized) { ... }
    // We run the event loop only after applying everything, and we check conditions if there are any non-query jobs.
    var checkConditions = jobs.Any(j => j.VariantCase != WorkflowActivationJob.VariantOneofCase.QueryWorkflow);
    RunOnce(checkConditions); 
}
```
avoid the double check for `applyModernEventLoopLogic`, not a strong opinion, feel free to ignore

THardy98 (approved):
Not blocking, small comments/changes.

and just to verify my understanding:
`Check all conditions at once and resolve all that are satisfied.`
->
`Only resolve the first condition that is satisfied and re-run the entire event loop.`

we're doing this so that we can give subsequent signals (i.e. if multiple signals are sent) a chance to be handled, in the case that resolved conditions from the first signal cause the workflow to complete/terminate/finish

src/Temporalio/Worker/WorkflowInstance.cs
cretz: The legacy behavior is `RunOnce` inside-loop after applying each job (so we have to preserve it here) and the new behavior is `RunOnce` after-loop (which is seen below). So we have to leave the inside-loop here for older workflows.

src/Temporalio/Worker/WorkflowInstance.cs
cretz: I considered combining, but the legibility of separating initialize from run-once is worth the top-level separation

cretz:
> and just to verify my understanding:
> Check all conditions at once and resolve all that are satisfied.
> ->
> Only resolve the first condition that is satisfied and re-run the entire event loop.
>
> we're doing this so that we can give subsequent signals (i.e. if multiple signals are sent) a chance to be handled, in the case that resolved conditions from the first signal cause the workflow to complete/terminate/finish

There's a few reasons why you don't want to tick the event loop after each job. I should have linked this to #375 which I am doing now, but also see https://github.com/temporalio/sdk-python/issues/606 and we have had internal discussions on this. Basically waking up the event loop mid-job-apply causes conditions and other things to be confusingly applied too early. It becomes hard to reason about when a coroutine in a handler may run. This also aligns the event loop with other SDKs (in addition to fixing a few other things).

src/Temporalio/Worker/WorkflowInstance.cs
THardy98: Oh I missed the `Apply(job);` line above, at the beginning of the loop 👍

THardy98:
> There's a few reasons why you don't want to tick the event loop after each job. I should have linked this to #375 which I am doing now, but also see [temporalio/sdk-python#606](https://github.com/temporalio/sdk-python/issues/606) and we have had internal discussions on this. Basically waking up the event loop mid-job-apply causes conditions and other things to be confusingly applied too early. It becomes hard to reason about when a coroutine in a handler may run. This also aligns the event loop with other SDKs (in addition to fixing a few other things).

Yeah I found these issues trying to gather some more context on this, very helpful. In effect, this boils down to predictable handler behavior (at least wrt to the principle bug fix). Thanks for the explanation :)

vchirikov:
Hi @cretz , when it will be released?

cretz:
Next .NET SDK release, but we do not have a firm date at this time, should not be _too_ long

PR #433, opened 2025-03-18
Ready: change readme activity examples from scheduletoclose to starttoclose

GSmithApps:
## What was changed
Changed the activity invocation examples from `ScheduleToClose` to `StartToClose` 

## Why?
I have been seeing some examples in the courses that use `ScheduleToClose` where I think we mean `StartToClose` in activities, and I suspect it's a copy-paste thing from here.

In my understanding, `StartToClose` is more common than `ScheduleToClose` for activities, so I just figured I'd switch it.


cretz (approved):
Thanks! It is true that we encourage start-to-close more. Will merge after CI passes.

PR #434, opened 2025-03-18
Increase visibility on some client constructs and minor docs updates

### LINKED ISSUE #218: [Feature Request] Add visibility link in list workflow method XML doc
Link https://docs.temporal.io/visibility from the list workflow method XML doc

### LINKED ISSUE #228: [Feature Request] Make sure all client responses are user-instantiable/mockable
Today some things, such as workflow list result, are only instantiable via `internal` constructor which makes it difficult for users to test. Try to make everything that could come back from a client able to be instantiated by a user. What many clients do (e.g. just checked Azure clients) is make their classes simple classes with parameterless constructors and have all properties also have setters. This may be preferred over our current stance of getter-only properties that cannot be customized in test situations.

### LINKED ISSUE #419: [Feature Request] Requesting that WorkflowHandle's members are marked virtual or inherit from interface

I keep running into cases where if I have a service object that manipulates workflow handles, I cannot fake / mock it because `WorkflowHandle` is a record that doesn't inherit from an interface & the methods on it aren't `virtual`, so a dynamic interceptor (like the one FakeItEasy uses) cannot intercept these methods. 

I think my problems would be solved if we marked methods on WorkflowHandle as `virtual`, or had it implement an interface.

### Additional context

I understand there are some aspects of Temporal where faking/mocking is discouraged in favor of integration style tests. I think this is a fair point for testing Workflows & Activities. 

I'm not sure if that's the rationale for making WorkflowHandle unmockable, but if it is I don't think that philosophy should be applied to WorkflowHandle. 

For example, I may have a service object with a method that takes in a request, parses something out, then gets a `WorkflowHandle` and signals the workflow. As a unit test, it'd be great to be able to assert that "given some request to my service object, a call to a `WorkflowHandle` is made to signal with input parameters matching XYZ" but that currently seems quite difficult.

cretz:
## What was changed

* Exposed constructors as `protected internal` on multiple currently-internal-constructor-only classes returned from client calls so they can be mocked, but added warning that we may change them in incompatible ways in the future
  * We don't really want to commit to constructor stability here IMO at this time
  * Yes it is a bit annoying that they are internal and therefore you have to extend them to construct them, but we feel `public` encourages using these constructors too much, something we are actually discouraging
* Made final overloads of some client calls `virtual` so they can be overridden in subclasses
* Added docs as needed

## Checklist

1. Closes #218
1. Closes #228
1. Closes #419

### DIFF
--- a/src/Temporalio/Client/ITemporalClient.Workflow.cs
+++ b/src/Temporalio/Client/ITemporalClient.Workflow.cs
@@ -177,6 +177,7 @@ namespace Temporalio.Client
         /// <param name="query">Query to use for filtering.</param>
         /// <param name="options">Options for the list call.</param>
         /// <returns>Async enumerator for the workflows.</returns>
+        /// <seealso href="https://docs.temporal.io/visibility">Visibility docs.</seealso>
         public IAsyncEnumerable<WorkflowExecution> ListWorkflowsAsync(
             string query, WorkflowListOptions? options = null);
 #endif
@@ -187,6 +188,7 @@ namespace Temporalio.Client
         /// <param name="query">Query to use for counting.</param>
         /// <param name="options">Options for the count call.</param>
         /// <returns>Count information for the workflows.</returns>
+        /// <seealso href="https://docs.temporal.io/visibility">Visibility docs.</seealso>
         public Task<WorkflowExecutionCount> CountWorkflowsAsync(
             string query, WorkflowCountOptions? options = null);
     }
--- a/src/Temporalio/Client/Schedules/ScheduleDescription.cs
+++ b/src/Temporalio/Client/Schedules/ScheduleDescription.cs
@@ -17,7 +17,15 @@ namespace Temporalio.Client.Schedules
         private readonly Lazy<IReadOnlyDictionary<string, IEncodedRawValue>> memo;
         private readonly Lazy<SearchAttributeCollection> searchAttributes;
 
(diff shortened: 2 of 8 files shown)


neiljohari (commented):
Looks pretty good! Thank you!

**question:** so as a customer to use this, I'll likely extend and create fixture versions of things like WorkflowExecution in my unit tests project? Is that the goal? Seems reasonable if it is

Also is there a preview build of this package on NuGet I could use to try out and give feedback?

cretz:
> I'll likely extend and create fixture versions of things like WorkflowExecution in my unit tests project? Is that the goal? Seems reasonable if it is

Yes, you can extend `WorkflowExecution` with your own and call the base constructor with the advanced things needed to construct it since it is from raw proto. We discourage mocking clients usually because it can be hard to ever get the details of what the server does right, which is why we offer test environments that start real servers that we encourage people to use (i.e. `Temporalio.Testing.WorkflowEnvironment.StartLocalAsync()` which returns an environment with a fully usable `Client` on it).

> Also is there a preview build of this package on NuGet I could use to try out and give feedback?

No, it will be available in next stable release after merged, but we do not publish NuGet preview releases. Though you can build it yourself (see bottom of README) or once merged to main, the main merge commit GitHub CI run will have artifacts with the NuGet package on it.

yuandrew (approved):

PR #446, opened 2025-04-09
Serialization context for converters and codecs

### LINKED ISSUE #438: [Feature Request] Serialization context for codecs and converters

See https://github.com/temporalio/features/issues/434

cretz:
## What was changed

* Added `IWithSerializationContext<T>` that can be implemented by data converters (and is), payload converters, encoding-specific payload converters, failure converters, and payload codecs to provide context-specific versions of themselves
* Added `ISerializationContext` as a base/marker interface for the contexts and nested `Activity` and `Workflow` class implementations of it
* Updated workflow client, schedule client, async activity client, workflow worker, activity worker, workflow outbound child, and workflow outbound external code to use context-specific converters/codecs.
* Added test to test most of those paths

## Checklist

1. Closes #438

### DIFF
--- a/src/Temporalio/Worker/WorkflowWorker.cs
+++ b/src/Temporalio/Worker/WorkflowWorker.cs
@@ -87,7 +87,6 @@ namespace Temporalio.Worker
             }))
             {
                 var tasks = new ConcurrentDictionary<Task, bool?>();
-                var codec = options.DataConverter.PayloadCodec;
                 while (true)
                 {
                     var act = await options.BridgeWorker.PollWorkflowActivationAsync().ConfigureAwait(false);
@@ -98,7 +97,7 @@ namespace Temporalio.Worker
                     }
                     // Otherwise handle and put on task set
                     // TODO(cretz): Any reason for users to need to customize factory here?
-                    var task = Task.Run(() => HandleActivationAsync(codec, act));
+                    var task = Task.Run(() => HandleActivationAsync(act));
                     tasks[task] = null;
                     _ = task.ContinueWith(
                         task => { tasks.TryRemove(task, out _); }, TaskScheduler.Current);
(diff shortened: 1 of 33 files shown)


src/Temporalio/Worker/WorkflowWorker.cs
```diff
@@ -215,46 +246,62 @@ private async Task HandleActivationAsync(IPayloadCodec? codec, WorkflowActivatio
...
-                        async _ =>
+                await options.BridgeWorker.CompleteWorkflowActivationAsync(comp).ConfigureAwait(false);
+            }
+            catch (Exception e)
+            {
+                logger.LogError(e, "Failed completing activation on {RunId}", act.RunId);
+            }
+        }
+
+        private async Task HandleCacheEvictionAsync(WorkflowActivation act, RemoveFromCache job)
```
cretz, line 261: When we added data conversion details to the existing `HandleActivationAsync` it complicated the code a lot because eviction was inside it too. So we took this opportunity to refactor and move eviction here to a separate method.

Sushisource: The royal "we" here is a bit hilarious lol

cretz: Maybe I have multiple personality disorder

Sushisource (commented):
Wow, definitely a bit of a laborious change.

I'm tempted to say it could maybe be a bit cleaner by creating some extension methods like `asSerializationContext` on things like some of the activation jobs/inputs/etc so that the places you're extracting info from them and turning them into a serialization context can be somewhat-deduped, and, maybe more importantly, just not create so much inline noise in the places where they're happening.

That's stylistic though and nonblocking

Sushisource (approved):
That was supposed to be approve

cretz:
Yeah most of that inline noise is just .NET taking so many characters to type, not sure having single-use helpers will help too much. Yes this is a somewhat large effort, as it was [in Java a couple years ago](https://github.com/temporalio/sdk-java/pull/1695), and it's going to be really rough for whoever has to do this work in Python.

PR #455, opened 2025-04-16
Main

GSmithApps:
## What was changed
Minor things in the readme

## Why?
General housekeeping

## Note to maintainers
Please change whatever you see fit. Thank you!


cretz (approved):
Accepting mostly blindly because the ultimate goal is to remove this README in favor of proper docs.temporal.io. Can you confirm that any of your edits don't also need to be made there? Should prioritize getting whatever information you think was missing here added there.

PR #466, opened 2025-04-24
Worker Deployment Versioning

### LINKED ISSUE #437: [Feature Request] Support New Worker Versioning API
We need to add annotations to workflows to support the new versioning APIs. Refer to temporalio/features#548 for the Go equivalents.

There are some designs for all SDK languages (including .NET) here: https://docs.google.com/document/d/1QGacjnlFwrfFtAvv9besx6sEdr36kJ3zQ9J7bGLRwk4/edit?usp=sharing

Sushisource:
## What was changed
Added the annotations and options for worker-deployment-based versioning

## Why?
New feature in all SDKs

## Checklist

1. Closes https://github.com/temporalio/sdk-dotnet/issues/437

2. How was this tested:
New & existing tests

3. Any docs updates needed?


Sushisource:
Ready for review, but will have one small addition of version override options on start workflow after https://github.com/temporalio/api/pull/579 is merged

src/Temporalio.Extensions.Hosting/TemporalWorkerServiceId.cs
```diff
@@ -0,0 +1,12 @@
+namespace Temporalio.Extensions.Hosting
+{
+    /// <summary>
+    /// Represents the unique identifier for a hosted Temporal Worker service.
+    /// </summary>
+    public record TemporalWorkerServiceId(
```
cretz: Pedantic I know, but can we call this "Identifier" so we don't get it confused with our user uses of "ID" and "identity"?

src/Temporalio.Extensions.Hosting/TemporalWorkerServiceId.cs
```diff
@@ -0,0 +1,12 @@
...
+        string TaskQueue,
+        string? Version,
+        bool IsBuildId)
+    {
+    }
```
cretz:
```suggestion
        bool IsBuildId);
```

I think this should work, unsure

src/Temporalio.Extensions.Hosting/TemporalWorkerService.cs
```diff
@@ -113,27 +114,107 @@ public TemporalWorkerService(
...
+        : this(
+                new TemporalWorkerServiceId(
+                    taskQueueAndDeploymentVersion.TaskQueue,
+                    taskQueueAndDeploymentVersion.DeploymentVersion?.ToCanonicalString(),
+                    false),
+                optionsMonitor,
+                existingClient,
+                loggerFactory)
+        {
+        }
```
cretz: I think this whole constructor can be deleted. The only reason the other existed was for DI (though users could technically call it directly), so I think the next constructor is good enough.

src/Temporalio.Extensions.Hosting/TemporalWorkerServiceId.cs
```diff
@@ -0,0 +1,12 @@
+namespace Temporalio.Extensions.Hosting
+{
+    /// <summary>
+    /// Represents the unique identifier for a hosted Temporal Worker service.
+    /// </summary>
+    public record TemporalWorkerServiceId(
+        string TaskQueue,
+        string? Version,
+        bool IsBuildId)
```
cretz:
```suggestion
        bool VersionIsBuildId)
```

Made it a bit clearer to me, but not a big deal. Alternatively could accept optional and mutually exclusive build ID _and_ `WorkerDeploymentVersion`, e.g.:

```csharp
public record TemporalWorkerServiceIdentifier(
    string TaskQueue,
    string? UnversionedBuildId,
    WorkerDeploymentVersion? Version)
{
    public TemporalWorkerServiceIdentifier(string taskQueue, string? unversionedBuildId = null)
        : this(taskQueue, unversionedBuildId, null)
    {
    }

    public TemporalWorkerServiceIdentifier(string taskQueue, WorkerDeploymentVersion version)
        : this(taskQueue, null, version)
    {
    }
}
```

But not a big deal

src/Temporalio/Worker/TemporalWorkerOptions.cs
```diff
@@ -109,15 +109,26 @@ public TemporalWorkerOptions()
         /// If unset, the default is
         /// <c>Assembly.GetEntryAssembly().ManifestModule.ModuleVersionId</c>.
         /// </summary>
+        /// <remarks>Exclusive with <see cref="DeploymentOptions"/>.</remarks>
+        [Obsolete("Use DeploymentOptions instead")]
```
cretz, line 113: Is this obsolete? People may still want this build ID even without versioning, correct?

src/Temporalio/Worker/TemporalWorkerOptions.cs
```diff
@@ -109,15 +109,26 @@ public TemporalWorkerOptions()
...
         public bool UseWorkerVersioning { get; set; }
 
+        /// <summary>
+        /// Gets or sets the deployment options for this worker.
+        /// </summary>
+        /// <remarks>WARNING: Deployment-based versioning is experimental and APIs may
+        /// change.</remarks>
+        public WorkerDeploymentOptions? DeploymentOptions { get; set; }
```
cretz, line 131:
```suggestion
        public WorkerDeploymentOptions? Deployment { get; set; }
```

In cases where we have sub-options, we don't usually put the `Options` suffix on the property name (but maybe we do inconsistently somewhere I didn't check)

src/Temporalio/Worker/TemporalWorkerOptions.cs
cretz, line 1: Can you update `Clone` to deep-clone the deployment options? Can see `TemporalConnectionOptions/Clone` for an example of how we've done this

src/Temporalio/Bridge/src/worker.rs
```diff
@@ -48,14 +47,43 @@ pub struct WorkerOptions {
...
-    use_worker_versioning: bool,
     max_concurrent_workflow_task_polls: u32,
     nonsticky_to_sticky_poll_ratio: f32,
     max_concurrent_activity_task_polls: u32,
     nondeterminism_as_workflow_fail: bool,
     nondeterminism_as_workflow_fail_for_types: ByteArrayRefArray,
 }
 
+#[repr(C)]
+pub enum WorkerVersioningStrategy {
```
cretz, line 58: I personally think how Rust/cbingen translates Rust enums to C tagged unions is a bit gross from a pure C API POV (unless there is a real memory concern making us want to use unions). Same with slot suppliers at #372 though I didn't think to look at the time. Not a big deal of course, though I do prefer mutually exclusive fields checked at runtime (e.g. how we take three mutually exclusive types of metrics in `MetricsOptions`). Don't have to change anything.

src/Temporalio/Bridge/src/worker.rs
```diff
@@ -807,13 +835,35 @@ impl TryFrom<&WorkerOptions> for temporal_sdk_core::WorkerConfig {
...
+                                    deployment_name: dopts.version.deployment_name.to_string(),
+                                    build_id: dopts.version.build_id.to_string(),
+                                },
+                                use_worker_versioning: dopts.use_worker_versioning,
+                                default_versioning_behavior: match dopts.default_versioning_behavior {
+                                    0 => None,
+                                    1 => Some(temporal_sdk_core_protos::temporal::api::enums::v1::VersioningBehavior::Pinned),
+                                    2 => Some(temporal_sdk_core_protos::temporal::api::enums::v1::VersioningBehavior::AutoUpgrade),
+                                    _ => bail!("Invalid default versioning behavior {}", dopts.default_versioning_behavior),
+                                }
```
cretz: Is there some int-to-enum thing we can do here dynamically instead of us remembering to come back and update this match?

src/Temporalio/Bridge/OptionsExtensions.cs
```diff
@@ -450,16 +450,66 @@ public static Interop.WorkerOptions ToInteropOptions(
...
-                }
                 var entryAssembly = Assembly.GetEntryAssembly() ??
                     throw new ArgumentException("Unable to get assembly manifest ID for build ID");
                 buildId = entryAssembly.ManifestModule.ModuleVersionId.ToString();
+                buildIdAutoDetermined = true;
+            }
+            Interop.WorkerVersioningStrategy versioningStrategy;
+            if (options.DeploymentOptions != null)
+            {
+                if (options.DeploymentOptions.Version == null)
```
cretz, line 467: Should we also check that `options.UseWorkerVersioning` is `false`?

cretz, line 467: Should we also check that `buildIdAutodetermined` is `true`? We're not even using the `buildId` local var when using deployment options so why default it above? Revisiting this, I think the logic needs to change to more clearly differentiate deployment options vs non-deployment options.

src/Temporalio/Workflows/WorkflowDefinition.cs
```diff
@@ -104,6 +109,16 @@ private WorkflowDefinition(
...
 
+        /// <summary>
+        /// Gets the versioning behavior.
+        /// </summary>
+        public VersioningBehavior? VersioningBehavior { get; private init; }
+
+        /// <summary>
+        /// Gets the dynamic configuration method.
+        /// </summary>
+        public Func<object?, WorkflowDefinitionOptions>? DynamicConfigMethod { get; private init; }
```
cretz:
```suggestion
        public Func<object?, WorkflowDefinitionOptions>? DynamicOptionsMethod { get; private init; }
```

cretz: What happens if I access `Workflow.Instance` inside this method? I think we should update `WorkflowInstance` to have `WorkflowDefinitionsOptions` be an internal property that is lazily created on first access. And that property can be used by all uses of failure exception types and versioning behaviors instead of having special fields for dynamic/runtime forms of those.

src/Temporalio/Workflows/WorkflowDefinition.cs
```diff
@@ -286,6 +304,27 @@ public static WorkflowDefinition Create(
                         runMethod = method;
                     }
                 }
+                if (method.IsDefined(typeof(WorkflowDynamicOptionsAttribute), false))
```
cretz, line 307: Did we decide whether any workflow can define this or if only dynamic workflows can? If any, I wonder if the term "dynamic" is confusing. I wonder if `[WorkflowOptionsProvider]` or something is a better name.

src/Temporalio/Workflows/WorkflowDefinitionOptions.cs
```diff
@@ -0,0 +1,35 @@
...
+    /// <summary>
+    /// Additional options for a workflow definition that may be returned by
+    /// a function annotated with <see cref="WorkflowDynamicOptionsAttribute"/>.
+    /// </summary>
+    public class WorkflowDefinitionOptions
+    {
+        /// <summary>
+        /// Gets the types of exceptions that, if a workflow-thrown exception extends, will
+        /// cause the workflow/update to fail instead of suspending the workflow via task failure.
+        /// These are applied in addition to
```
cretz, line 15: Somewhere need to clarify if set, this replaces any value set on the `WorkflowAttribute`

src/Temporalio/Workflows/WorkflowDefinitionOptions.cs
```diff
@@ -0,0 +1,35 @@
...
+        /// all others. But this default may change in the future.
+        /// </remarks>
+        public Type[]? FailureExceptionTypes { get; init; }
+
+        /// <summary>
+        /// Gets the versioning behavior to use for this workflow.
+        /// </summary>
+        /// <remarks>WARNING: Deployment-based versioning is experimental and APIs may
+        /// change.</remarks>
+        public VersioningBehavior VersioningBehavior { get; init; }
```
cretz: Need to clarify if not left unspecified, this overwrites any value set on the `WorkflowAttribute`

src/Temporalio/Workflows/WorkflowDefinition.cs
```diff
@@ -286,6 +304,27 @@ public static WorkflowDefinition Create(
...
+                    if (method.ReturnType != typeof(WorkflowDefinitionOptions))
+                    {
+                        throw new ArgumentException($"WorkflowDynamicOptions method {method} must return WorkflowDefinitionOptions");
+                    }
+                    var parameters = method.GetParameters();
+                    if (parameters.Length != 0)
+                    {
+                        throw new ArgumentException($"WorkflowDynamicOptions method {method} must take no parameters");
+                    }
+                    dynamicOptionsMethod = instance => (WorkflowDefinitionOptions)method.Invoke(instance, null)!;
```
cretz, line 334: Add one more check here that makes sure `dynamicOptionsMethod` isn't already set (i.e. we only allow one of these)

tests/Temporalio.Tests/TestUtils.cs
```diff
@@ -224,4 +225,60 @@ public record CaptureMetric<T>(string Name, string? Unit, string? Description)
 
         public void Set(T value, object tags) => Values.Enqueue((value, (Dictionary<string, object>)tags));
     }
-}
\ No newline at end of file
+
+    public static async Task<Temporalio.Api.WorkflowService.V1.SetWorkerDeploymentRampingVersionResponse>
```
cretz, line 229: Arguably this utility and that last one doesn't provide much value vs inlining the call heh

tests/Temporalio.Tests/TestUtils.cs
```diff
@@ -224,4 +225,60 @@ public record CaptureMetric<T>(string Name, string? Unit, string? Description)
...
+            {
+                Namespace = client.Options.Namespace,
+                DeploymentName = version.DeploymentName,
+                Version = version.ToCanonicalString(),
+                ConflictToken = conflictToken,
+                Percentage = rampPercentage,
+            });
+    }
+
+    public static async Task<Temporalio.Api.WorkflowService.V1.DescribeWorkerDeploymentResponse>
```
cretz, line 249: I'd personally toss this in `AssertMore` and call it `EventuallyWorkerDeploymentPresentAsync` or something, but no biggie

src/Temporalio.Extensions.Hosting/TemporalWorkerServiceId.cs
Sushisource: Eh, they're not really mutually exclusive though. It's no safer.

src/Temporalio/Worker/TemporalWorkerOptions.cs
Sushisource: We're going to move people towards using a deployment, even if versioning is off.

src/Temporalio/Worker/TemporalWorkerOptions.cs
Sushisource: I think it makes more sense here, because of the use versioning flag inside. Deployment implies just the name part

src/Temporalio/Bridge/src/worker.rs
Sushisource: Nothing that will be guaranteed to be caught at compile time. I would need linear types.

src/Temporalio.Extensions.Hosting/TemporalWorkerServiceId.cs
cretz: Just a bit more confusing that `Version` isn't `Version`, it's sometimes `BuildId`. Maybe we can just do the boolean property name change.

src/Temporalio/Worker/TemporalWorkerOptions.cs
cretz: I see. Then I think the `AddHostedTemporalWorker(this IServiceCollection services, string taskQueue, string? buildId = null)` overload of the DI creator should be delegating to the `WorkerDeploymentOptions` overload instead of setting this option.

src/Temporalio/Bridge/src/worker.rs
cretz: Not following exactly. A value of `3` here isn't caught at compile time either. An invalid integer here I think is a runtime error.

src/Temporalio/Bridge/OptionsExtensions.cs
Sushisource: It never gets autodetermined here because they have to provide version, which means they have to provide build id. The autodetermined thing is used for the `UseWorkerVersioning`/fallback cases. 

I have added the stuff to enforce mutual exclusion which makes it more clear

cretz: I see. I wonder if we should check that `options.BuildId` is null then if it is ignored in these cases

src/Temporalio.Extensions.Hosting/TemporalHostingServiceCollectionExtensions.cs
```diff
@@ -62,23 +63,96 @@ public static ITemporalWorkerServiceOptionsBuilder AddHostedTemporalWorker(
         public static ITemporalWorkerServiceOptionsBuilder AddHostedTemporalWorker(
             this IServiceCollection services, string taskQueue, string? buildId = null)
         {
+            // TODO: Review - We probably need to deprecate buildId param & add deployment version?
```
cretz: If the `BuildId` worker option is deprecated, this should be too IMO

src/Temporalio.Extensions.Hosting/TemporalHostingServiceCollectionExtensions.cs
```diff
@@ -62,23 +63,96 @@ public static ITemporalWorkerServiceOptionsBuilder AddHostedTemporalWorker(
...
                 ActivatorUtilities.CreateInstance<TemporalWorkerService>(
-                    provider, (TaskQueue: taskQueue, BuildId: buildId)));
+                    provider, new TemporalWorkerServiceIdentifier(taskQueue, buildId, true)));
             return new TemporalWorkerServiceOptionsBuilder(taskQueue, buildId, services).ConfigureOptions(
                 options =>
                 {
                     options.TaskQueue = taskQueue;
+#pragma warning disable 0618
                     options.BuildId = buildId;
+#pragma warning restore 0618
```
cretz, line 80: If this is deprecated, does that mean that there's a way to use `WorkerDeploymentOptions` to represent "just build ID"? Otherwise, how do I represent just build ID in a non-deprecated way if I don't care about deployment name and won't ever enable versioning? And if I can do that, why does this not just call that overload?

src/Temporalio.Extensions.Hosting/TemporalWorkerServiceOptions.cs
```diff
@@ -53,15 +53,15 @@ public override object Clone()
         /// Get an options name for the given task queue and build ID.
         /// </summary>
         /// <param name="taskQueue">Task queue.</param>
-        /// <param name="buildId">Build ID.</param>
+        /// <param name="version">Worker Depoyment version or Build ID.</param>
```
cretz:
```suggestion
        /// <param name="version">Worker Deployment version or Build ID.</param>
```

src/Temporalio.Extensions.Hosting/TemporalWorkerService.cs
```diff
@@ -113,27 +114,72 @@ public TemporalWorkerService(
         /// WARNING: Do not rely on the signature of this constructor, it is for DI container use
         /// only and may change in incompatible ways.
         /// </remarks>
-        [ActivatorUtilitiesConstructor]
         public TemporalWorkerService(
             (string TaskQueue, string? BuildId) taskQueueAndBuildId,
             IOptionsMonitor<TemporalWorkerServiceOptions> optionsMonitor,
             ITemporalClient? existingClient = null,
             ILoggerFactory? loggerFactory = null)
+        : this(
```
cretz, line 122: If the `BuildId` worker option is deprecated, this constructor should be too IMO

src/Temporalio.Extensions.Hosting/TemporalWorkerService.cs
```diff
@@ -113,27 +114,72 @@ public TemporalWorkerService(
...
+                if (options.BuildId != serviceId.Version)
+                {
+                    throw new InvalidOperationException(
+                        $"BuildID '{serviceId.Version ?? "<unset>"}' on constructor different than '{options.BuildId ?? "<unset>"}' on options");
+                }
+#pragma warning restore 0618
+            }
+            else
+            {
+                if (options.DeploymentOptions?.Version?.ToCanonicalString() != serviceId.Version)
```
cretz, line 178: Should there be a check to ensure `options.BuildId` is `null` and `options.UseVersioning` is `false` here? Meaning, if they are using new versioning, shouldn't they not be setting the old values away from their default?

src/Temporalio.Extensions.Hosting/TemporalWorkerServiceOptionsBuilder.cs
```diff
@@ -32,13 +33,30 @@ public TemporalWorkerServiceOptionsBuilder(string taskQueue, string? buildId, IS
...
+            TaskQueue = taskQueue;
+            DeploymentOptions = deploymentOptions;
+            Services = services;
+        }
+
         /// <inheritdoc />
         public string TaskQueue { get; private init; }
 
         /// <inheritdoc />
         public string? BuildId { get; private init; }
```
cretz, line 58: I think we should deprecate all uses of `BuildId` in the DI stuff if it's deprecated for the non-DI stuff

src/Temporalio/Workflows/IWorkflowContext.cs
```diff
@@ -78,6 +83,12 @@ internal interface IWorkflowContext
         /// </summary>
         object Instance { get; }
 
+        /// <summary>
+        /// Gets the options defined by the <see cref="WorkflowDynamicOptionsAttribute" /> annotated
+        /// method, if any.
+        /// </summary>
+        WorkflowDefinitionOptions? DynamicDefinitionOptions { get; }
```
cretz: I don't think we need to expose this to users

src/Temporalio/Worker/WorkflowInstance.cs
```diff
@@ -651,6 +682,20 @@ public WorkflowActivationCompletion Activate(WorkflowActivation act)
                             var checkConditions = jobs.Any(j => j.VariantCase != WorkflowActivationJob.VariantOneofCase.QueryWorkflow);
                             RunOnce(checkConditions);
                         }
+
+                        var establishedVersioningBehavior = Definition.VersioningBehavior;
+                        Console.WriteLine("Established versioning behavior: " + establishedVersioningBehavior);
+                        Console.WriteLine("Dynamic versioning behavior: " + dynamicDefinitionOptions?.VersioningBehavior);
```
cretz: Leftover stuff

src/Temporalio/Workflows/WorkflowDefinitionOptions.cs
```diff
@@ -0,0 +1,42 @@
...
+        /// <remarks>
+        /// If set, overrides any value set on <see cref="WorkflowAttribute.FailureExceptionTypes"/>.
+        /// </remarks>
+        /// <remarks>
+        /// WARNING: This property is experimental and may change in the future. If unset
+        /// (i.e. left null), currently the default is to only fail the workflow/update on
+        /// <see cref="Exceptions.FailureException" /> + cancellation and suspend via task failure
+        /// all others. But this default may change in the future.
+        /// </remarks>
+        public Type[]? FailureExceptionTypes { get; init; }
```
cretz: I think options should be `get`/`set` for those using builder patterns (otherwise `record`s are better because at least you can `with` them)

src/Temporalio/Worker/WorkflowInstance.cs
```diff
@@ -304,10 +312,30 @@ public object Instance
...
                 instance ??= Definition.CreateWorkflowInstance(startArgs!.Value);
+                // Dynamic options method needs to be invoked at this point, after initting the
+                // workflow instance but before performing any activations.
+                _ = DynamicDefinitionOptions;
                 return instance;
             }
         }
 
+        /// <inheritdoc />
+        public WorkflowDefinitionOptions? DynamicDefinitionOptions
```
cretz: I was thinking this would be more like:

```csharp
internal WorkflowDefinitionOptions DefinitionOptions
{
    get
    {
        if (definitionOptions == null)
        {
            definitionOptions = new()
            {
                FailureExceptionTypes = Definition.FailureExceptionTypes,
                VersioningBehavior = Definition.VersioningBehavior,
            };
            if (Definition.DynamicOptionsMethod != null)
            {
                var dynOptions = Definition.DynamicOptionsMethod.Invoke(Instance);
                if (dynOptions.FailureExceptionTypes != null)
                {
                    definitionOptions.FailureExceptionTypes = dynOptions.FailureExceptionTypes;
                }
                if (dynOptions.VersioningBehavior != VersioningBehavior.Unspecified)
                {
                    definitionOptions.VersioningBehavior = dynOptions.VersioningBehavior;
                }
            }
        }
        return definitionOptions;
    }
}
```

and then `DefinitionOptions` can be used everywhere in code here and elsewhere. Though I could understand concern of it being accessed before instance made available. I wasn't expecting it to be user-facing.

And if you're always going to be initializing this in one place, no need to make it lazy.

src/Temporalio/Worker/WorkflowInstance.cs
```diff
@@ -304,10 +312,30 @@ public object Instance
             {
                 // We create this lazily because we want the constructor in a workflow context
                 instance ??= Definition.CreateWorkflowInstance(startArgs!.Value);
+                // Dynamic options method needs to be invoked at this point, after initting the
+                // workflow instance but before performing any activations.
+                _ = DynamicDefinitionOptions;
```
cretz: If you're always running this anyways, then you don't need to make it lazy IMO, can just set the definition options right here. Before I didn't think you cared about when you called the dynamic definition method, but I can see now it is better to be predictable. My concern was what happens if the dynamic options method accesses `Workflow.Instance` per https://github.com/temporalio/sdk-dotnet/pull/466#discussion_r2070621271?

cretz (commented):
Added a few more comments based on recent updates, but older unaddressed comments may still apply

src/Temporalio/Workflows/WorkflowDefinition.cs
Sushisource: It's only dynamic workflows. Added enforcement.

src/Temporalio/Bridge/src/worker.rs
Sushisource: Oh, you mean just always convert blindly by integer. Yeah, we can do that.

src/Temporalio/Bridge/OptionsExtensions.cs
Sushisource: Git or github did something really weird with this review. I had a bunch of changes that just didn't get published even though I have them locally and everything is pushed.

I had all this in here after my first commit with review comments but it's just magically not there. I'll figure out how to get it actually be in sync

(they appear to be there now)

Sushisource: It's checked now

src/Temporalio/Worker/TemporalWorkerOptions.cs
Sushisource: It can't - it would have to make up a deployment name. I'm deprecating it per the other comment (let's try to keep related ones like that together on one thread).

src/Temporalio.Extensions.Hosting/TemporalHostingServiceCollectionExtensions.cs
Sushisource: Yes, it means that, just Build ID is deprecated. You use a fully-formed deployment name now

src/Temporalio/Worker/WorkflowInstance.cs
Sushisource: Yeah, that was exactly the point of doing it inline there. It needs to happen at the right time

src/Temporalio/Worker/WorkflowInstance.cs
Sushisource: Ah, ok, I get what you mean now, use that as a bag for the things set on definition too, sure, makes sense.

It can't _exactly_ work this way because the failure types on the definition can be used _before_ initialization ever even happens, as in `ExecuteWorkflowAsync_BadInput_CanFailWorkflow`, which was fun to discover. The idea is pretty much the same still though.

src/Temporalio.Extensions.Hosting/TemporalWorkerService.cs
Sushisource: I don't think there needs to be, because if the versions match, then they couldn't possibly have set the other options on `options`, because they're exclusive.

src/Temporalio.Extensions.Hosting/TemporalHostingServiceCollectionExtensions.cs
```diff
@@ -34,12 +34,13 @@ public static class TemporalHostingServiceCollectionExtensions
...
         /// </param>
         /// <returns>Options builder to configure the service.</returns>
+        [Obsolete("Use the AddHostedTemporalWorker overload with WorkerDeploymentOptions instead")]
         public static ITemporalWorkerServiceOptionsBuilder AddHostedTemporalWorker(
             this IServiceCollection services,
             string clientTargetHost,
             string clientNamespace,
             string taskQueue,
-            string? buildId = null) =>
+            string buildId) =>
```
Sushisource, line 43: In order to make these obsolete, I had to make them not take null for the last param, and instead have that happen on the new one. I believe that should be a compatible change still

src/Temporalio/Worker/TemporalWorkerOptions.cs
cretz: That was my concern with deprecating `BuildId`, we're deprecating `BuildId`-only without a replacement. People use this field and now they're going to have to set their code to ignore deprecation because they don't have a dot in their build ID where they can convert. I'm not sure it's fair to make people make up a deployment name for no reason or benefit to them just to use build ID.

src/Temporalio.Extensions.Hosting/TemporalHostingServiceCollectionExtensions.cs
cretz: I think this is a bit rough on users who shouldn't have to care about deployment name to just have their existing build ID functionality remain

src/Temporalio.Extensions.Hosting/TemporalWorkerService.cs
cretz: What if they used the overload that only sets the legacy/deprecated build ID and does not set deployment options, won't this conditional fail for them now?

cretz: I think my concern was if I was using new deployment options _and_ have manually set old build ID, but I guess worker creation throws for that, no worries

src/Temporalio.Extensions.Hosting/TemporalHostingServiceCollectionExtensions.cs
cretz: With C# named parameters, it's technically not compatible since a user could have had `buildId: null`. Can you leave the `string?` type here (can still make the parameter required I believe).

src/Temporalio/Worker/WorkflowInstance.cs
```diff
@@ -87,6 +87,8 @@ internal class WorkflowInstance : TaskScheduler, IWorkflowInstance, IWorkflowCon
         private WorkflowUpdateDefinition? dynamicUpdate;
         private bool workflowInitialized;
         private bool applyModernEventLoopLogic;
+        private WorkflowDefinitionOptions definitionOptions;
```
cretz:
```suggestion
        private readonly WorkflowDefinitionOptions definitionOptions;
```

(and move to the readonly set of fields, unless I'm missing something)

src/Temporalio/Workflows/WorkflowDefinitionOptions.cs
```diff
@@ -0,0 +1,42 @@
+using System;
+using Temporalio.Common;
+
+namespace Temporalio.Workflows
+{
+    /// <summary>
+    /// Additional options for a workflow definition that may be returned by
+    /// a function annotated with <see cref="WorkflowDynamicOptionsAttribute"/>.
+    /// </summary>
+    public class WorkflowDefinitionOptions
```
cretz: Can this be cloneable? Not really important, just a pattern we've been doing for "options"

src/Temporalio/Worker/WorkflowInstance.cs
```diff
@@ -651,6 +682,12 @@ public WorkflowActivationCompletion Activate(WorkflowActivation act)
                             var checkConditions = jobs.Any(j => j.VariantCase != WorkflowActivationJob.VariantOneofCase.QueryWorkflow);
                             RunOnce(checkConditions);
                         }
+
+                        if (definitionOptions?.VersioningBehavior != null)
```
cretz: I don't think this condition could ever evaluate to false now

src/Temporalio/Worker/WorkflowInstance.cs
```diff
@@ -983,7 +1020,7 @@ private bool IsWorkflowFailureException(Exception e) =>
             // reusing cancellation tokens if the workflow fails.
             e is FailureException ||
                 e is OperationCanceledException ||
-                Definition.FailureExceptionTypes?.Any(t => t.IsAssignableFrom(e.GetType())) == true ||
+                definitionOptions?.FailureExceptionTypes?.Any(t => t.IsAssignableFrom(e.GetType())) == true ||
```
cretz:
```suggestion
                definitionOptions.FailureExceptionTypes?.Any(t => t.IsAssignableFrom(e.GetType())) == true ||
```

src/Temporalio/Worker/WorkflowWorker.cs
```diff
@@ -53,6 +54,17 @@ public WorkflowWorker(
                         "Note, dependency injection is not supported in workflows. " +
                         "Workflows must be deterministic and self-contained with a lifetime controlled by Temporal.");
                 }
+
+                if (options.DeploymentOptions?.UseWorkerVersioning == true)
+                {
+                    if (options.DeploymentOptions?.DefaultVersioningBehavior is null or VersioningBehavior.Unspecified
```
cretz, line 60: Interesting, I have not used `or` in this project before, but it is C# 9, so works

src/Temporalio/Workflows/WorkflowDefinition.cs
```diff
@@ -104,6 +109,16 @@ private WorkflowDefinition(
...
 
+        /// <summary>
+        /// Gets the versioning behavior.
+        /// </summary>
+        public VersioningBehavior? VersioningBehavior { get; private init; }
+
+        /// <summary>
+        /// Gets the dynamic options method.
+        /// </summary>
+        public Func<object?, WorkflowDefinitionOptions>? DynamicOptionsMethod { get; private init; }
```
cretz: Arguably not a "method" like "runMethod", more like a `DynamicOptionsGetter` like "creator", but doesn't really matter

cretz (commented):
Only piddly things remain I think then we're good

src/Temporalio/Worker/TemporalWorkerOptions.cs
Sushisource: That's a product decision, I'm just implementing it. At the end of the day I agree with it though. Defining an application/deployment name isn't exactly a huge burden. You do it one time and you're done.

cretz: It's not whether it's a burden, it's whether it's necessary for us to retain the existing Temporal behavior they are used to that doesn't relate to this. I can bring this up internally, not blocking by any means.

cretz (approved):
LGTM, thanks! The concern over deprecating build ID for those users that don't care about deployment name is not .NET specific, so doesn't really affect this PR specifically.

Sushisource:
Like TS, want to get https://github.com/temporalio/api/pull/579 in so I can use it in this before merging, so I don't have to go back immediately after

cretz:
No prob, can leave this hanging open as long as needed.

src/Temporalio/Common/Priority.cs
```diff
@@ -0,0 +1,60 @@
...
+        /// <param name="priority">The proto to initialize from.</param>
+        internal Priority(Temporalio.Api.Common.V1.Priority priority)
+            : this(priority?.PriorityKey != 0 ? priority?.PriorityKey : null)
+        {
+        }
+
+        /// <summary>
+        /// Gets the default Priority instance.
+        /// </summary>
+        public static Priority Default => new();
```
cretz:
```suggestion
        public static Priority Default { get; } = new();
```

Forgot to mention this on other PR. We shouldn't reinstantiate this every time the property is called.

src/Temporalio/Common/VersioningOverride.cs
```diff
@@ -0,0 +1,78 @@
...
+        /// <param name="Behavior">The behavior of the pinned override.</param>
+        public sealed record PinnedVersioningOverride(
+            WorkerDeploymentVersion Version,
+            PinnedOverrideBehavior Behavior = PinnedOverrideBehavior.Pinned) : VersioningOverride;
+
+        /// <summary>
+        /// The workflow will auto-upgrade to the current deployment version on the next workflow
+        /// task.
+        /// </summary>
+        public sealed record AutoUpgradeVersioningOverride() : VersioningOverride;
```
cretz:
```suggestion
        public sealed record Pinned(
            WorkerDeploymentVersion Version,
            PinnedOverrideBehavior Behavior = PinnedOverrideBehavior.Pinned) : VersioningOverride;

        /// <summary>
        /// The workflow will auto-upgrade to the current deployment version on the next workflow
        /// task.
        /// </summary>
        public sealed record AutoUpgrade() : VersioningOverride;
```

Probably don't need the `VersioningOverride` suffix on these

src/Temporalio/Common/VersioningOverride.cs
```diff
@@ -0,0 +1,78 @@
...
+    public abstract record VersioningOverride
+    {
+        private VersioningOverride()
+        {
+        }
+
+        /// <summary>
+        /// Specifies different sub‐types of pinned override.
+        /// </summary>
+        public enum PinnedOverrideBehavior
```
cretz, line 18: Could also move this to be an inner class of `Pinned` called `Behavior`, leaving the only direct inner types of `VersioningOverride` to be implementations of it. Usually .NET doesn't have so many nested types, but in this case it makes sense I think

cretz (approved):
Semantic things only

PR #468, opened 2025-04-28
Handle errors with update result serialization

### LINKED ISSUE #465: [Bug] Update result serialization failure silently swallowed
### Describe the bug

Due to how update chose to use `task.ContinueWith` (maybe not the best idea in #142), when an exception occurs in update result serialization, it is swallowed. We need to make sure to surface this as a failure likely following the same rules as regular update failure (so for most it'll be a task failure).

cretz:
## What was changed

Previously we swallowed failures to serialize update results. Now we will fail the update or the task depending on the error.

## Checklist

1. Closes #465

### DIFF
--- a/src/Temporalio/Worker/WorkflowInstance.cs
+++ b/src/Temporalio/Worker/WorkflowInstance.cs
@@ -1185,49 +1185,73 @@ namespace Temporalio.Worker
                 return task.ContinueWith(
                     _ =>
                     {
-                        inProgressHandlers.Remove(inProgress);
-                        // If workflow failure exception, it's an update failure. If it's some
-                        // other exception, it's a task failure. Otherwise it's a success.
-                        var exc = task.Exception?.InnerExceptions?.SingleOrDefault();
-                        // There are .NET cases where cancellation occurs but is not considered
-                        // an exception. We are going to make it an exception. Unfortunately
-                        // there is no easy way to make it include the outer stack trace at this
-                        // time.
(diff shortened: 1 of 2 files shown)


src/Temporalio/Worker/WorkflowInstance.cs
```diff
@@ -1185,49 +1185,73 @@ private Task ApplyDoUpdateAsync(DoUpdate update)
                 return task.ContinueWith(
                     _ =>
                     {
-                        inProgressHandlers.Remove(inProgress);
```
cretz, line 1188: The PR makes this a bit ugly, I recommend "hide whitespace" in the this diff view (i.e. https://github.com/temporalio/sdk-dotnet/pull/468/files?diff=unified&w=1)

tests/Temporalio.Tests/Worker/WorkflowWorkerTests.cs
```diff
@@ -3081,7 +3081,7 @@ public class DynamicHandlersWorkflow
         [WorkflowRun]
         public Task RunAsync() => Workflow.WaitConditionAsync(() => false);
 
-        [WorkflowSignal]
+        [WorkflowUpdate]
```
cretz, line 3084: This is to fix a mostly-unrelated flake where slow GH runners were not guaranteeing signal order as the test expected

Sushisource (approved):

PR #470, opened 2025-05-02
Add Priorities

Sushisource:
Depends on https://github.com/temporalio/sdk-dotnet/pull/466 merging first

## What was changed
Add priority settings for workflows/activities

## Why?
Feature parity

## Checklist

1. Closes https://github.com/temporalio/sdk-dotnet/issues/403

2. How was this tested:
Added tests

3. Any docs updates needed?

### DIFF
--- a/src/Temporalio/Common/Priority.cs
+++ b/src/Temporalio/Common/Priority.cs
@@ -14,14 +14,27 @@ namespace Temporalio.Common
     /// 1. First, consider "priority_key": lower number goes first.
     /// (more will be added here later).
     /// </summary>
-    public sealed record Priority()
+    public sealed record Priority
     {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Priority"/> class.
+        /// </summary>
+        /// <param name="priorityKey">The priority key.</param>
+        public Priority(int? priorityKey = null) => PriorityKey = priorityKey;
+
         /// <summary>
         /// Initializes a new instance of the <see cref="Priority"/> class from a proto.
         /// </summary>
         /// <param name="priority">The proto to initialize from.</param>
         internal Priority(Temporalio.Api.Common.V1.Priority priority)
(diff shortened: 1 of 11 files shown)


src/Temporalio/Common/Priority.cs
```diff
@@ -0,0 +1,47 @@
...
+        /// Gets the priority key, which is a positive integer from 1 to n, where smaller integers
+        /// correspond to higher priorities (tasks run sooner). In general, tasks in a queue should
+        /// be processed in close to priority order, although small deviations are possible. The
+        /// maximum priority value (minimum priority) is determined by server configuration, and
+        /// defaults to 5.
+        ///
+        /// The default priority is (min+max)/2. With the default max of 5 and min of 1, that comes
+        /// out to 3.
+        /// </summary>
+        public int? PriorityKey { get; init; }
```
cretz, line 49: I think this should be in the primary constructor (can be default null if wanted)

src/Temporalio/Worker/ActivityWorker.cs
```diff
@@ -183,6 +183,7 @@ private void StartActivity(Bridge.Api.ActivityTask.ActivityTask tsk)
                 HeartbeatDetails: start.HeartbeatDetails,
                 HeartbeatTimeout: OptionalTimeSpan(start.HeartbeatTimeout),
                 IsLocal: start.IsLocal,
+                Priority: new(start.Priority),
```
cretz: Not a super big deal, but a GC/instance-count conscious .NET dev (there are a lot of them) would have something like `start.Priority is { } p ? new(p) : Priority.Default` since it is immutable, meaning a `public static Priority Default { get; } = new()` property on the `Priority` class I think is valuable to have for everyone (including us).

Granted some may prefer `record struct` for such a thing, though it wouldn't work well with being nullable in some of the options (you'd use `default` instead). But `record struct` came in C# 10 anyways so we can't use it.

cretz (commented):
Only two minor suggestions

cretz (approved):

PR #472, opened 2025-05-07
Fail the workflow task if a requested local activity is not registered

tconley1428:
## What was changed
Executing an unregistered local activity will fail the workflow task rather than just the activity.

## Why?
Reduces unhelpful retries.

## Checklist

1. Closes 463

2. How was this tested:
New unit test added to WorkflowWorkerTests

3. Any docs updates needed?
None

### DIFF
--- a/src/Temporalio/Worker/TemporalWorker.cs
+++ b/src/Temporalio/Worker/TemporalWorker.cs
@@ -93,7 +93,7 @@ namespace Temporalio.Worker
                     RuntimeMetricMeter: MetricMeter,
                     WorkerLevelFailureExceptionTypes: options.WorkflowFailureExceptionTypes,
                     DisableEagerActivityExecution: options.DisableEagerActivityExecution,
-                    ActivityLookup: (activityType) => activityWorker?.ActivityLookup(activityType) ?? null));
+                    AssertValidActivity: (activityType) => activityWorker?.AssertValidActivity(activityType)));
             }
         }
 
--- a/src/Temporalio/Worker/TemporalWorker.cs
+++ b/src/Temporalio/Worker/TemporalWorker.cs
@@ -92,7 +92,8 @@ namespace Temporalio.Worker
                     OnTaskCompleted: options.OnTaskCompleted,
                     RuntimeMetricMeter: MetricMeter,
                     WorkerLevelFailureExceptionTypes: options.WorkflowFailureExceptionTypes,
-                    DisableEagerActivityExecution: options.DisableEagerActivityExecution));
+                    DisableEagerActivityExecution: options.DisableEagerActivityExecution,
+                    ActivityLookup: (activityType) => activityWorker?.ActivityLookup(activityType) ?? null));
             }
         }
 
--- a/src/Temporalio/Worker/TemporalWorker.cs
+++ b/src/Temporalio/Worker/TemporalWorker.cs
@@ -93,7 +93,11 @@ namespace Temporalio.Worker
                     RuntimeMetricMeter: MetricMeter,
                     WorkerLevelFailureExceptionTypes: options.WorkflowFailureExceptionTypes,
                     DisableEagerActivityExecution: options.DisableEagerActivityExecution,
-                    AssertValidActivity: (activityType) => activityWorker?.AssertValidActivity(activityType)));
+                    AssertValidLocalActivity: (activityType) => (activityWorker ??
+                                throw new InvalidOperationException(
+                                    $"Activity {activityType} is not registered on this worker," +
+                                    $" no available activities."))
+                                .AssertValidActivity(activityType)));
             }
         }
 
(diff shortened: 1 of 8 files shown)


src/Temporalio/Worker/WorkflowWorkerOptions.cs
```diff
@@ -22,5 +22,6 @@ internal record WorkflowWorkerOptions(
         Action<WorkflowInstance, Exception?> OnTaskCompleted,
         Lazy<MetricMeter> RuntimeMetricMeter,
         IReadOnlyCollection<Type>? WorkerLevelFailureExceptionTypes,
-        bool DisableEagerActivityExecution);
+        bool DisableEagerActivityExecution,
+        Action<string> AssertValidActivity);
```
cretz: Pedantic, but I'd call this `AssertValidLocalActivity` (don't need to change the source method name in activity worker, since that makes sense). The concern is that non-local activities don't use this so the name may be a bit confusing, because this only asserts local activity validity.

src/Temporalio/Worker/TemporalWorker.cs
```diff
@@ -92,7 +92,8 @@ public TemporalWorker(IWorkerClient client, TemporalWorkerOptions options)
                     OnTaskCompleted: options.OnTaskCompleted,
                     RuntimeMetricMeter: MetricMeter,
                     WorkerLevelFailureExceptionTypes: options.WorkflowFailureExceptionTypes,
-                    DisableEagerActivityExecution: options.DisableEagerActivityExecution));
+                    DisableEagerActivityExecution: options.DisableEagerActivityExecution,
+                    AssertValidActivity: (activityType) => activityWorker?.AssertValidActivity(activityType)));
```
cretz: I wonder if we should fail if there is no activity worker? Technically it's not a valid local activity if there is no activity worker. Unsure how to best handle this, open to just about any approach.

cretz (commented):
LGTM, very minor things. Only blocking thing is what to do when there is no activity worker.

src/Temporalio/Worker/TemporalWorker.cs
```diff
@@ -92,7 +92,12 @@ public TemporalWorker(IWorkerClient client, TemporalWorkerOptions options)
                     OnTaskCompleted: options.OnTaskCompleted,
                     RuntimeMetricMeter: MetricMeter,
                     WorkerLevelFailureExceptionTypes: options.WorkflowFailureExceptionTypes,
-                    DisableEagerActivityExecution: options.DisableEagerActivityExecution));
+                    DisableEagerActivityExecution: options.DisableEagerActivityExecution,
+                    AssertValidLocalActivity: (activityType) => (activityWorker ??
+                                throw new InvalidOperationException(
+                                    $"Activity {activityType} is not registered on this worker," +
+                                    $" no available activities."))
+                                .AssertValidActivity(activityType)));
```
cretz, line 100: A bit convoluted for a single-expression lambda, but meh. Feel free to have a private method in this class do this and just pass that. But not required.

tests/Temporalio.Tests/Worker/WorkflowWorkerTests.cs
```diff
@@ -7210,6 +7210,86 @@ public async Task ExecuteWorkflowAsync_PayloadMissing_StillWorks()
...
+        var err = await ExecuteWorkerAsync<LocalActivityMissingWorkflow, WorkflowHandle>(
+            async worker =>
+            {
+                var handle = await Client.StartWorkflowAsync(
+                    (LocalActivityMissingWorkflow wf) => wf.RunAsync(),
+                    new($"workflow-{Guid.NewGuid()}", worker.Options.TaskQueue!));
+                await AssertTaskFailureContainsEventuallyAsync(handle, "Activity invalid_activity is not registered on this worker, available activities: DoSomething");
+                return handle;
+            },
+            new TemporalWorkerOptions().AddAllActivities<LocalActivityMissingWorkflow>(null));
```
cretz:
```suggestion
        await ExecuteWorkerAsync<LocalActivityMissingWorkflow>(
            async worker =>
            {
                var handle = await Client.StartWorkflowAsync(
                    (LocalActivityMissingWorkflow wf) => wf.RunAsync(),
                    new($"workflow-{Guid.NewGuid()}", worker.Options.TaskQueue!));
                await AssertTaskFailureContainsEventuallyAsync(handle, "Activity invalid_activity is not registered on this worker, available activities: DoSomething");
            },
            new TemporalWorkerOptions().AddAllActivities<LocalActivityMissingWorkflow>(null));
```

Pedantic, but no need for `err` or `return` of the handle or any of that here (or in other tests below)

tests/Temporalio.Tests/Worker/WorkflowWorkerTests.cs
```diff
@@ -7210,6 +7210,86 @@ public async Task ExecuteWorkflowAsync_PayloadMissing_StillWorks()
...
+        var err = await ExecuteWorkerAsync<LocalActivityMissingNoActivitiesWorkflow, WorkflowHandle>(
+            async worker =>
+            {
+                var handle = await Client.StartWorkflowAsync(
+                    (LocalActivityMissingNoActivitiesWorkflow wf) => wf.RunAsync(),
+                    new($"workflow-{Guid.NewGuid()}", worker.Options.TaskQueue!));
+                await AssertTaskFailureContainsEventuallyAsync(handle, "Activity invalid_activity is not registered on this worker, no available activities.");
+                return handle;
+            },
+            new TemporalWorkerOptions());
```
cretz:
```suggestion
            });
```

cretz (approved):
Only non-blocking things. Update them if you'd like or not, then feel free to merge.

PR #474, opened 2025-05-09
Expose poller automation

yuandrew:
Currently targeted for `deployment-versioning` branch, will switch to main once Spencer's PR goes in.

## What was changed

added opt-in poller autoscaling 

## Checklist

1. Closes 

2. How was this tested:

 Added simple test. More meaningful tests are in Core, we just need to make sure option is passed properly here.

3. Any docs updates needed?


src/Temporalio/Bridge/src/worker.rs
```diff
@@ -832,6 +869,7 @@ impl TryFrom<&WorkerOptions> for temporal_sdk_core::WorkerConfig {
 
     fn try_from(opt: &WorkerOptions) -> anyhow::Result<Self> {
         let converted_tuner: temporal_sdk_core::TunerHolder = (&opt.tuner).try_into()?;
+        println!("opt.max_concurrent_workflow_task_polls {:?} {:?}", opt.max_concurrent_activity_task_polls.autoscaling, opt.max_concurrent_activity_task_polls.simple_maximum);
```
cretz: Leftover debugging?

src/Temporalio/Worker/TemporalWorkerOptions.cs
```diff
@@ -207,9 +207,9 @@ public TemporalWorkerOptions()
 
         /// <summary>
         /// Gets or sets the maximum number of concurrent poll workflow task requests we will
-        /// perform at a time on this worker's task queue. Default is 5.
+        /// perform at a time on this worker's task queue. Default is SimpleMaximum(5).
         /// </summary>
-        public int MaxConcurrentWorkflowTaskPolls { get; set; } = 5;
+        public PollerBehavior MaxConcurrentWorkflowTaskPolls { get; set; } = new PollerBehavior.SimpleMaximum(5);
```
cretz: We can't make backwards incompatible changes to user-facing options like this. We should make a new properties that override these if set. Not even sure we can make these nullable `int`s safely, we may just have to say the other properties override these blindly if set.

src/Temporalio/Worker/Tuning/PollerBehavior.cs
```diff
@@ -0,0 +1,73 @@
...
+        /// A poller behavior that will attempt to poll as long as a slot is available, up to the
+        /// provided maximum. Cannot be less than two for workflow tasks, or one for other tasks.
+        /// </summary>
+        public sealed class SimpleMaximum : PollerBehavior
+        {
+            /// <summary>
+            /// Initializes a new instance of the <see cref="PollerBehavior.SimpleMaximum"/> class.
+            /// </summary>
+            /// <param name="maximum">The maximum number of pollers at a time.</param>
+            public SimpleMaximum(uint maximum = 5)
```
cretz: I think this should be an `int` instead of a `uint` even though it can never be negative. We can check at runtime it's not negative and can document as such, but like Go and Java, usually general integers in .NET are `int`s unless there's a really good reason for them not to be.

cretz: I don't think the constructor parameters here or in autoscaling one should have default values, same as they don't in https://github.com/temporalio/sdk-python/pull/830

tests/Temporalio.Tests/Worker/WorkflowWorkerTests.cs
```diff
@@ -7395,6 +7396,95 @@ await ExecuteWorkerAsync<WorkflowUsingPriorities>(
...
+
+        [WorkflowSignal]
+        public async Task CompleteAsync() => complete = true;
+
+        [Activity]
+        public static string SomeActivity() => "activity-result";
+    }
+
+    [Fact]
+    public async Task AutoscallingPollingWorker()
```
cretz: Can we name this test similar to other test names?

tests/Temporalio.Tests/Worker/WorkflowWorkerTests.cs
```diff
@@ -7395,6 +7396,95 @@ await ExecuteWorkerAsync<WorkflowUsingPriorities>(
...
+            {
+                TargetHost = Client.Connection.Options.TargetHost,
+                Namespace = Client.Options.Namespace,
+                Runtime = runtime,
+            });
+
+        await ExecuteWorkerAsync<WaitOnSignalWorkflowThenActivity>(
+        async worker =>
+        {
+            await client.WorkflowService.GetSystemInfoAsync(new());
```
cretz:
```suggestion
```

This is done implicitly by the system

tests/Temporalio.Tests/Worker/WorkflowWorkerTests.cs
```diff
@@ -7395,6 +7396,95 @@ await ExecuteWorkerAsync<WorkflowUsingPriorities>(
...
+
+            // Give pollers a beat to start
+            await Task.Delay(TimeSpan.FromMilliseconds(300));
```
cretz: Instead of sleeping up front for a fixed time, can you wrap what you want to assert into an `AssertMore.EventuallyAsync`?

tests/Temporalio.Tests/Worker/WorkflowWorkerTests.cs
```diff
@@ -7395,6 +7396,95 @@ await ExecuteWorkerAsync<WorkflowUsingPriorities>(
             new TemporalWorkerOptions().AddAllActivities<WorkflowUsingPriorities>(null));
     }
 
+    [Workflow]
+    public class WaitOnSignalWorkflowThenActivity
```
cretz:
```suggestion
    public class WaitOnSignalWorkflowThenActivityWorkflow
```

Pedantic but suffixing with `Workflow` is common

tests/Temporalio.Tests/Worker/WorkflowWorkerTests.cs
```diff
@@ -7395,6 +7396,95 @@ await ExecuteWorkerAsync<WorkflowUsingPriorities>(
...
+                "First workflow poller count should be 1 or 2");
+            Assert.True(
+                workflowPollers[1].EndsWith('2') || workflowPollers[1].EndsWith('1'),
+                "Second workflow poller count should be 1 or 2");
+
+            // Run multiple workflow instances concurrently
+            var workflowTasks = Enumerable.Range(0, 20).Select(async _ =>
+            {
+                var handle = await client.StartWorkflowAsync(
+                    (WaitOnSignalWorkflow wf) => wf.RunAsync(),
```
cretz: This seems to be starting a workflow that is not on this worker. After the signal below, you should wait on the workflow result if you expect it to finish

tests/Temporalio.Tests/Worker/WorkflowWorkerTests.cs
```diff
@@ -7395,6 +7396,95 @@ await ExecuteWorkerAsync<WorkflowUsingPriorities>(
...
+                    new($"workflow-{Guid.NewGuid()}", worker.Options.TaskQueue!));
+                await handle.SignalAsync(wf => wf.CompleteAsync());
+            });
+
+            await Task.WhenAll(workflowTasks);
```
cretz, line 7478: What is the purpose of running the workflows/activities if you aren't asserting anything about poller behavior?

src/Temporalio/Worker/TemporalWorkerOptions.cs
yuandrew: Thanks for callout, didn't realize we used a new property in the other PRs 😅

tests/Temporalio.Tests/Worker/WorkflowWorkerTests.cs
yuandrew: Looking at [python](https://github.com/temporalio/sdk-python/pull/830/files#diff-9b903f5884cdfe70b0e3d6c9bc456d016a33fcc5bb72d6e7c3393a535867e289) and [TS](https://github.com/temporalio/sdk-typescript/pull/1704/files#diff-7c6ec2f93f1091a3efce1cb940bcf183c992b94b0961d881a03b527767bc9de0), I think the goal is to ensure all 20 workflows/activities complete successfully, but as you mentioned above I was missing the `GetResultAsync` to ensure all workflows completed

src/Temporalio/Worker/TemporalWorkerOptions.cs
```diff
@@ -207,9 +207,11 @@ public TemporalWorkerOptions()
 
         /// <summary>
         /// Gets or sets the maximum number of concurrent poll workflow task requests we will
-        /// perform at a time on this worker's task queue. Default is 5.
+        /// perform at a time on this worker's task queue.
         /// </summary>
-        public int MaxConcurrentWorkflowTaskPolls { get; set; } = 5;
+        /// <remarks>If set, will override any value set in <see cref="WorkflowTaskPollerBehavior"/>.</remarks>
+        [Obsolete("Use WorkflowTaskPollerBehavior instead")]
+        public int? MaxConcurrentWorkflowTaskPolls { get; set; }
```
cretz: Hrmm, not sure it is deprecated and would have to think through whether changing this from `int` to `int?` is a backwards compatible change we're ok with (it's a bit different to change a data type on an options set than it is in a parameter list). Will bring up internally.

One option is to leave this alone and make `PollerBehavior` the optional item and say if _that_ is set it overrides _this_. That is usually a safer way to introduce something new and opt-in. Can even have a check if you want that confirms if they changed the int field from the default, then the poller behavior should be null.

src/Temporalio/Bridge/src/worker.rs
```diff
@@ -47,13 +46,51 @@ pub struct WorkerOptions {
...
     nondeterminism_as_workflow_fail_for_types: ByteArrayRefArray,
 }
 
+#[repr(C)]
+pub struct PollerSimpleMaximum {
+    pub simple_maximum: usize,
+}
+
+#[repr(C)]
+pub struct PollerAutoscaling {
```
cretz:
```suggestion
pub struct PollerBehaviorSimpleMaximum {
    pub simple_maximum: usize,
}

#[repr(C)]
pub struct PollerBehaviorAutoscaling {
```

Pedantic, but figured the full `PollerBehavior` prefix makes more sense in the C header when using the C API

tests/Temporalio.Tests/Worker/WorkflowWorkerTests.cs
cretz: :+1: Now that `await handle.GetResultAsync()` is in there, it makes sense.

src/Temporalio/Worker/TemporalWorkerOptions.cs
cretz: Ok, confirmed with @Sushisource this doesn't break expectations. So lets leave these properties alone (no deprecation, no data type change) and just have poller behavior presence _override_ these if set (can note this in docs above both). We don't have to go the extra step of throwing an exception if these are changed from default _and_ poller behavior present, but can if you think it's worth it.

cretz (approved):

Issue #458, opened 2025-04-17
[Bug] Issues with ReserveCtxFromBridge


cretz:
### Describe the bug

There was a bug in older CLI where persistence may fail and the worker was getting something like this warning:

    2025-04-17T18:56:22.257529Z  WARN temporal_sdk_core::worker::workflow: Task not found when completing error=status: NotFound, message: "Namespace 0196451a-5a35-7546-9b8d-84a40f52f73e is not found."

Notice the namespace not found. This may be causing a panic/segfault somewhere in the code. Try to replicate this situation, maybe by deleting the namespace while running.

cretz:
Here's a segfault on Windows after this came from the server:

```
Fatal error. System.AccessViolationException: Attempted to read or write protected memory. This is often an indication that other memory is corrupt.
   at Temporalio.Bridge.CustomSlotSupplier.ReserveCtxFromBridge(Temporalio.Bridge.Interop.SlotReserveCtx*)
   at Temporalio.Bridge.CustomSlotSupplier+<>c__DisplayClass8_0+<<SafeReserve>b__0>d.MoveNext()
   at System.Threading.ExecutionContext.RunInternal(System.Threading.ExecutionContext, System.Threading.ContextCallback, System.Object)
   at System.Runtime.CompilerServices.AsyncTaskMethodBuilder`1+AsyncStateMachineBox`1[[System.Threading.Tasks.VoidTaskResult, System.Private.CoreLib, Version=6.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e],[System.__Canon, System.Private.CoreLib, Version=6.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]].MoveNext(System.Threading.Thread)
   at System.Runtime.CompilerServices.TaskAwaiter+<>c.<OutputWaitEtwEvents>b__12_0(System.Action, System.Threading.Tasks.Task)
   at System.Threading.Tasks.AwaitTaskContinuation.RunOrScheduleAction(System.Action, Boolean)
   at System.Threading.Tasks.Task.RunContinuations(System.Object)
   at System.Threading.Tasks.Task.TrySetResult()
   at System.Threading.Tasks.Task+DelayPromise.CompleteTimedOut()
   at System.Threading.TimerQueueTimer.Fire(Boolean)
   at System.Threading.TimerQueue.FireNextTimers()
[...]

PR #477, opened 2025-05-14
Expose poller automation

yuandrew:
remake of #474, now targeted at main

## What was changed

added opt-in poller autoscaling 

## Checklist

1. Closes 

2. How was this tested:

 Added simple test. More meaningful tests are in Core, we just need to make sure option is passed properly here.

3. Any docs updates needed?


cretz (approved):
Looks great. Seems CI is having two flakes. One I have not seen before and opened #478, another is #458. Rerunning CI now...

PR #481, opened 2025-05-16
Improve/verbosify error messages in various ToInteropOptions methods

adeal:
## What was changed

This a small change to make (most of) the ArgumentException messages in OptionsExtensions more verbose. Specifically, this change focuses on errors caused by the absence of a property being set, and includes the name of the options class that it is missing from.

## Why?

The various `ToInteropOptions(...)` methods are the first place where options classes get checked for correctness. Since these are "deeper" in the SDK than what most users are digging into, it can make the error challenging to interpret.

For example, we can look at an error thrown when `TargetHost` is not set for `HttpConnectProxyOptions`:

image

1. The call stack is unrelated to the location where a user might configure these client options, such as `Program.cs` or `Startup.cs`, so it can be unclear that this is a user error
2. "TargetHost" is an overloaded property name throughout the SDK, so disambiguating it to `HttpConnectProxyOptions` would be helpful

## Checklist

1. Closes 

4. How was this tested:

Not necessary; no functional changes.

5. Any docs updates needed?

None known.

### DIFF
--- a/src/Temporalio/Bridge/OptionsExtensions.cs
+++ b/src/Temporalio/Bridge/OptionsExtensions.cs
@@ -150,7 +150,7 @@ namespace Temporalio.Bridge
         {
             if (string.IsNullOrEmpty(options.Filter.FilterString))
             {
-                throw new ArgumentException($"FilterString is required when {nameof(Temporalio.Runtime.LoggingOptions)} is configured");
+                throw new ArgumentException($"FilterString is required when {nameof(Temporalio.Runtime.TelemetryFilterOptions)} is configured");
             }
             return new Interop.LoggingOptions()
             {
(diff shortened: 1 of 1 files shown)


src/Temporalio/Bridge/OptionsExtensions.cs
```diff
@@ -150,7 +150,7 @@ public static unsafe Interop.LoggingOptions ToInteropOptions(
         {
             if (string.IsNullOrEmpty(options.Filter.FilterString))
             {
-                throw new ArgumentException("Logging filter string is required");
+                throw new ArgumentException($"FilterString is required when {nameof(Temporalio.Runtime.LoggingOptions)} is configured");
```
adeal:
```suggestion
                throw new ArgumentException($"FilterString is required when {nameof(Temporalio.Runtime.TelemetryFilterOptions)} is configured");
```

cretz (approved):
Thanks!

PR #482, opened 2025-05-16
Activity pause support

### LINKED ISSUE #441: Heartbeating activities should be interrupted when the activities are paused.
See https://github.com/temporalio/features/issues/602

cretz:
## What was changed

* Added `ActivityCancellationDetails` with cancellation information
* Added `ActivityExecutionContext.CancellationDetails` which returns an instance of cancellation details upon cancel
* Added `ActviityCancelReason.Paused`
* Added support for setting cancellation details in activity testing environment

## Checklist

1. Closes #441


src/Temporalio/Activities/ActivityCancellationDetails.cs
```diff
@@ -0,0 +1,42 @@
...
+        public bool IsGoneFromServer { get; init; }
+
+        /// <summary>
+        /// Gets a value indicating whether the activity was explicitly cancelled.
+        /// </summary>
+        public bool IsCancelRequested { get; init; }
+
+        /// <summary>
+        /// Gets a value indicating whether the activity timeout caused activity to be marked
+        /// cancelled.
```
Sushisource, line 22:
```suggestion
        /// Gets a value indicating whether the activity is timed out.
```

Technically this can be set true without it being the cause I think?

Sushisource (approved):

src/Temporalio/Activities/ActivityCancellationDetails.cs
cretz: Without it being the singular cause, true, but it is _one_ of the causes. I guess I was using the word "caused" here to clarify that this is not true if it is timed out, it is true if timeout was one of the causes that tripped cancellation. You can trip cancellation for something else and _then_ timeout.

PR #484, opened 2025-05-28
Improve code analyze of not null member on ActivityDefinition

JobaDiniz:
This pull request introduces a small enhancement to the `ActivityDefinition` class in `src/Temporalio/Activities/ActivityDefinition.cs`. The change adds a `[MemberNotNullWhen]` attribute to the `Dynamic` property, improving nullability annotations and ensuring better static analysis.

### DIFF
--- a/src/Temporalio/Activities/ActivityDefinition.cs
+++ b/src/Temporalio/Activities/ActivityDefinition.cs
@@ -57,6 +57,7 @@ namespace Temporalio.Activities
         /// <summary>
         /// Gets a value indicating whether the activity is dynamic.
         /// </summary>
+        [MemberNotNullWhen(true, nameof(Name))]
         public bool Dynamic => Name == null;
 
         /// <summary>


cretz:
This would also apply to workflow definitions too. We of course could just get rid of the `Dynamic` property, it has no real value, it's just a helper, not sure how it helps analysis to say that a `X == null` being `false` means that `X` is not null (also I think the code would change to `[MemberNotNullWhen(false, nameof(Name))]`).

Regardless, unfortunately this attribute is only in newer versions of .NET (5+) and we must support older versions (see CI build). We can consider dropping support for, say, .NET framework, but that would be a wider discussion and not sure we want to do it for this one attribute. We could put compiler directives around it, though arguably it doesn't have enough value to even do that.

JobaDiniz:
The value is just _developer experience_.

When I did:

```csharp
if (!input.Dynamic)
return;

var activityName = input.Name;
```

The compiler complained that `input.Name` can be null.

Of course, I then did:

```
if (string.IsNullOrEmpty(intput.Name))
return;

var activityName = input.Name;
```

But `Dynamic` property would be better to **convey** clearly why - _because the activity is dynamic and we won't handle that_.

cretz:
:+1: Makes sense. The current concern is that this attribute is not in all .NET versions we support.

PR #485, opened 2025-06-05
Updated sdk-core to latest

maciejdudko:
## What was changed
Updated `sdk-core` submodule to latest upstream commit https://github.com/temporalio/sdk-core/commit/8913cdb27d18f2297fa3f76cc10911874bf89df2

The update broke Workflow Replayer due to bad poller behavior config, the config was fixed.


cretz (approved):
LGTM, but looks like one test at `ExecuteWorkflowAsync_PollingBehavior_Autoscaling` may be flaky.

You also may need to regenerate protos (that fails a CI check we have only in the Windows run) since new ones came over in the submodule. See https://github.com/temporalio/sdk-dotnet?tab=readme-ov-file#regenerating-protos. Unsure if there are other aspects that need to be added (e.g. any new client calls).

Sushisource (approved):
May be worth checking out that flake per Chad's suggestion. I think that might be a matter of just adjusting the test to use an eventually assert, because these counts are timing sensitive.

PR #486, opened 2025-06-06
💥 BREAKING - Make static summary/details lazy on workflow description

### LINKED ISSUE #460: [Feature Request] Change get static summary/details on workflow in client to be lazy

Today we eagerly decode these, change to be lazy upon request

cretz:
## What was changed

Changed `WorkflowExecutionDescription` properties of `StaticSummary` and `StaticDetails` to now be methods of `GetStaticSummaryAsync` and `GetStaticDetailsAsync`. This is a breaking change to experimental API.

## Checklist

1. Closes #460

### DIFF
--- a/src/Temporalio/Client/WorkflowExecutionDescription.cs
+++ b/src/Temporalio/Client/WorkflowExecutionDescription.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Threading.Tasks;
 using Temporalio.Api.WorkflowService.V1;
 using Temporalio.Converters;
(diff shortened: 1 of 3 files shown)


src/Temporalio/Client/WorkflowExecutionDescription.cs
```diff
@@ -9,57 +10,46 @@ namespace Temporalio.Client
...
-            string? staticSummary,
-            string? staticDetails,
-            DataConverter dataConverter)
+            DescribeWorkflowExecutionResponse rawDescription, DataConverter dataConverter)
             : base(rawDescription.WorkflowExecutionInfo, dataConverter, null)
         {
             RawDescription = rawDescription;
-            StaticSummary = staticSummary;
-            StaticDetails = staticDetails;
+#pragma warning disable VSTHRD011 // This should not be able to deadlock
```
maciejdudko, line 26: This is almost certainly safe but I'd still explore changing it to `AsyncLazy` sometime in the future.

maciejdudko (approved):
Looks good, LGTM!

src/Temporalio/Client/WorkflowExecutionDescription.cs
cretz: The concern there is https://learn.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.threading.asynclazy-1?view=visualstudiosdk-2019 requires a dependency we don't really want to take. But we could make our own if we had to.

JobaDiniz: take Stephen implementation: https://devblogs.microsoft.com/dotnet/asynclazyt/

PR #492, opened 2025-06-24
Fix issue where child workflow starts and completes in same activation

cretz:
## What was changed

There was an issue where, when a child start and a child resolve came in the same activation, the resolve had no pending child to work with because the start code doesn't run to transfer its state to pending resolve.

Changed code to only have one container of pending children and not differentiate waiting for start vs waiting for result. This should have no backwards incompatibilities that I can see. Reviewers may want to "hide whitespace" to see what changed more clearly.


jaroslaw-korecki (approved):

maciejdudko (approved):

grzegorz-pasternak:
Hello, is this fixing the following problem? 

image

cretz:
It _may_ fix that problem in rare cases if child workflow initiated and child workflow completed events are in the same task.

PR #495, opened 2025-07-10
Very rough outline of plugins

tconley1428:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/src/Temporalio/Client/Interceptors/IClientPlugin.cs
+++ b/src/Temporalio/Client/Interceptors/IClientPlugin.cs
@@ -1,0 +1,18 @@
+using System.Threading.Tasks;
+
+namespace Temporalio.Client.Interceptors
+{
+    /// <summary>
+    /// Plugin for client calls.
+    /// </summary>
+    public interface IClientPlugin
+    {
+        public void InitClientPlugin(IClientPlugin nextPlugin);
+
+        public TemporalClientOptions OnCreateClient(TemporalClientOptions options);
+
+        public Task<TemporalConnection> TemporalConnectAsync(TemporalClientConnectOptions options);
+
+        public TemporalConnection TemporalConnect(TemporalClientConnectOptions options);
+    }
+}
--- a/src/Temporalio/Client/TemporalClient.cs
+++ b/src/Temporalio/Client/TemporalClient.cs
@@ -1,6 +1,8 @@
 ﻿using System;
 using System.Linq;
 using System.Threading.Tasks;
+using Temporalio.Client.Interceptors;
+using Temporalio.Common;
 
 namespace Temporalio.Client
 {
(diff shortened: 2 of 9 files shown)


src/Temporalio/Client/TemporalClient.cs
```diff
@@ -112,5 +143,33 @@ internal partial class Impl : Interceptors.ClientOutboundInterceptor
             /// </summary>
             internal TemporalClient Client { get; init; }
         }
+
+#pragma warning disable CA1822 // We don't want to force plugin methods to be static
+        /// <summary>
+        /// Placeholder.
+        /// </summary>
+        internal class RootPlugin : Plugin
```
cretz, line 151: Would recommend making this implement the interface w/ a throw `NotImplementedException` for init

src/Temporalio/Client/Interceptors/IClientPlugin.cs
```diff
@@ -0,0 +1,18 @@
...
+    /// <summary>
+    /// Plugin for client calls.
+    /// </summary>
+    public interface IClientPlugin
+    {
+        public void InitClientPlugin(IClientPlugin nextPlugin);
+
+        public TemporalClientOptions OnCreateClient(TemporalClientOptions options);
+
+        public Task<TemporalConnection> TemporalConnectAsync(TemporalClientConnectOptions options);
```
cretz, line 14: Would call this `ClientConnectAsync` and other `ClientCreateLazy`

src/Temporalio/Common/Plugin.cs
```diff
@@ -0,0 +1,35 @@
...
+using Temporalio.Client.Interceptors;
+using Temporalio.Worker;
+using Temporalio.Worker.Interceptors;
+
+namespace Temporalio.Common
+{
+    public class Plugin : IClientPlugin, IWorkerPlugin
+    {
+        private IWorkerPlugin? nextWorkerPlugin;
+        private IClientPlugin? nextClientPlugin;
```
cretz, line 14: Would recommend these be protected properties, and that they return non-null values and fail if the backing fields are null

src/Temporalio/Client/Interceptors/IClientPlugin.cs
cretz, line 1: Would not recommend putting this in the `Interceptors` namespace, and would recommend calling this `ITemporalClientPlugin` (same two things for worker side)

src/Temporalio/Common/Plugin.cs
```diff
@@ -0,0 +1,35 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Temporalio.Client;
+using Temporalio.Client.Interceptors;
+using Temporalio.Worker;
+using Temporalio.Worker.Interceptors;
+
+namespace Temporalio.Common
+{
+    public class Plugin : IClientPlugin, IWorkerPlugin
```
cretz, line 11: Mark all methods here as virtual

src/Temporalio/Client/TemporalClient.cs
```diff
@@ -112,5 +143,33 @@ internal partial class Impl : Interceptors.ClientOutboundInterceptor
...
+        {
+            /// <summary>
+            /// Place.
+            /// </summary>
+            /// <param name="options">holder.</param>
+            /// <returns>Connection.</returns>
+            public new async Task<TemporalConnection> TemporalConnectAsync(TemporalClientConnectOptions options)
+            {
+                return await TemporalConnection.ConnectAsync(options).ConfigureAwait(false);
+            }
```
cretz, line 161:
```suggestion
            public Task<TemporalConnection> TemporalConnectAsync(TemporalClientConnectOptions options) =>
                TemporalConnection.ConnectAsync(options);
```

Pedantic shortcut

PR #504, opened 2025-08-07
Do not require heartbeating to be done in context

### LINKED ISSUE #487: [Bug] Improperly accessing current activity context in outbound heartbeat
### Describe the bug

First found in Ruby at https://github.com/temporalio/sdk-ruby/issues/279, we are accessing the current context in the outbound activity heartbeater to get the task token. We should not, because we allow heartbeating outside of context. So we should give the interceptor the task token another way.

maciejdudko:
## What was changed

Fixed issue where we were accessing current context to get token for heartbeating.

## Checklist

1. Closes #487

2. How was this tested:
Added test case `ActivityWorkerTests.ExecuteActivityAsync_BackgroundThreadHeartbeat_Received`


Sushisource (approved):

cretz (approved):
Thanks!

PR #508, opened 2025-08-20
Fairness Keys & Weights

Sushisource:
## What was changed
Added fairness keys/weights to priority

## Why?
New feature

## Checklist

1. Closes 

2. How was this tested:
Added test

3. Any docs updates needed?


.github/workflows/ci.yml
```diff
@@ -69,8 +69,9 @@ jobs:
           ClangSharpPInvokeGenerator @src/Temporalio/Bridge/GenerateInterop.rsp
           dotnet run --project src/Temporalio.Api.Generator
           npx doctoc README.md
+          git add src/Temporalio/Api
```
cretz: Is this leftover code from using CI to generate some aspect?

src/Temporalio/Common/Priority.cs
```diff
@@ -2,32 +2,42 @@ namespace Temporalio.Common
...
     public sealed record Priority
     {
         /// <summary>
         /// Initializes a new instance of the <see cref="Priority"/> class.
         /// </summary>
         /// <param name="priorityKey">The priority key.</param>
-        public Priority(int? priorityKey = null) => PriorityKey = priorityKey;
+        /// <param name="fairnessKey">The fairness key.</param>
+        /// <param name="fairnessWeight">The fairness weight.</param>
+        public Priority(int? priorityKey = null, string? fairnessKey = null, float? fairnessWeight = null)
```
cretz, line 25: We are inconsistent across languages on whether we require a priority key in the constructor or not

tests/Temporalio.Tests/AssertMore.cs
```diff
@@ -127,11 +128,21 @@ public static void Every<T>(IEnumerable<T> items, params Action<T>[] actions)
...
+                if (!found)
+                {
+                    Console.WriteLine($"Failed to match item: {item}");
+                    Console.WriteLine("Attempted matches and their failures:");
+                    for (int i = 0; i < exceptions.Count; i++)
                     {
+                        Console.WriteLine($"  Match {i + 1}: {exceptions[i].Message}");
                     }
+                    Assert.True(found, $"Item {item} had no match");
                 }
```
cretz: Would rather not write to stdout in this utility, if you really need to give back the exceptions, can probably change to:

```suggestion
                if (!found)
                {
                    throw new AggregateException($"Item {item} had no match", exceptions);
                }
```

.github/workflows/ci.yml
Sushisource: It's not leftover, it's genuinely useful for getting protos from CI that are net-new files

src/Temporalio/Common/Priority.cs
Sushisource: Yeah it's actually Ruby that was the odd one out here. Everywhere else is optional. Adding a fix here: https://github.com/temporalio/sdk-ruby/pull/323

Sushisource:
Hmm, looks like there are some additional changes to the span expectations that are causing issues

.github/workflows/ci.yml
cretz: I forget the `git diff` details, but do we also need to change `git diff --exit-code` to be `git diff --cached --exit-code`? The goal is to detect anything that doesn't match a re-run of the generator, including new files. Also, what about the other generated proto place, `src/Temporalio/Bridge/Api`?

src/Temporalio/Common/Priority.cs
cretz: Python also requires priority key be set when constructing, note how https://github.com/temporalio/sdk-python/blob/3c6fae6606428bb895c4a0df0c0a9c20c6e4bf82/temporalio/common.py#L985 doesn't have a default of `None`

.github/workflows/ci.yml
Sushisource: Added those 👍

tests/Temporalio.Tests/AssertMore.cs
```diff
@@ -119,6 +119,7 @@ public static void Every<T>(IEnumerable<T> items, params Action<T>[] actions)
             foreach (var item in items)
             {
                 var found = false;
+                var exceptions = new List<Exception>();
```
cretz: Technically no need for this collection anymore, but no big deal

cretz (approved):

tests/Temporalio.Tests/AssertMore.cs
Sushisource: Ah, derp

PR #509, opened 2025-08-23
Env config

### LINKED ISSUE #490: [Feature Request] Environment Configuration
**Describe the solution you'd like**

See https://github.com/temporalio/features/issues/441 and https://github.com/temporalio/proposals/blob/master/all-sdk/external-client-configuration.md

THardy98:
## What was changed
Add environment configuration feature

1. Closes #490

2. How was this tested:
Integration test suite `ClientConfigTests.cs`

3. Any docs updates needed?
Yes


THardy98:
Relies on https://github.com/temporalio/sdk-core/pull/986

src/Temporalio/Client/Configuration/ClientConfig.cs
```diff
@@ -0,0 +1,124 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text.Json;
+
+namespace Temporalio.Client.Configuration
```
cretz:
```suggestion
namespace Temporalio.Client.EnvConfig
```

I think `EnvConfig` is better because 1) that's what we're calling it everywhere else, and 2) there's a lot of configuration not in here, so it's not really general purpose "configuration"

src/Temporalio/Client/Configuration/ClientConfig.cs
```diff
@@ -0,0 +1,124 @@
...
+        /// <param name="profiles">The configuration profiles.</param>
+        public ClientConfig(IReadOnlyDictionary<string, ClientConfigProfile> profiles)
+        {
+            Profiles = profiles ?? new Dictionary<string, ClientConfigProfile>();
+        }
+
+        /// <summary>
+        /// Gets the configuration profiles.
+        /// </summary>
+        public IReadOnlyDictionary<string, ClientConfigProfile> Profiles { get; private set; }
```
cretz:
```suggestion
    public sealed record ClientConfig(IReadOnlyDictionary<string, ClientConfigProfile> profiles)
    {
```
I think this is better as a record instead of a simple class like our "options" classes are. And can remove cloneable since records have their own way to kinda clone. Same for the other options.

src/Temporalio/Client/Configuration/ClientConfig.cs
```diff
@@ -0,0 +1,124 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text.Json;
+
+namespace Temporalio.Client.Configuration
+{
+    /// <summary>
+    /// Represents the overall client configuration.
+    /// </summary>
+    public sealed class ClientConfig : ICloneable
```
cretz: For many languages, this is generic "client config" name is fine because the namespace is often qualifying it at the callsite. But for .NET, I would recommend calling this `ClientEnvConfig`. Also, same for `ClientEnvConfigProfile` and `ClientEnvConfigTls`, though I wonder if we should consider just making those two, `Profile` and `Tls`, as inner classes here. Not sure, thoughts?

src/Temporalio/Client/Configuration/ClientConfig.cs
```diff
@@ -0,0 +1,124 @@
...
+        /// <param name="configSource">The data source to load from.</param>
+        /// <param name="disableFile">If true, do not load from file (only from environment).</param>
+        /// <param name="configFileStrict">If true, fail if configuration file is invalid.</param>
+        /// <param name="overrideEnvVars">Environment variables to use, or null to use system environment.</param>
+        /// <returns>Loaded configuration data.</returns>
+        public static ClientConfig Load(
+            DataSource? configSource = null,
+            bool disableFile = false,
+            bool configFileStrict = false,
+            Dictionary<string, string>? overrideEnvVars = null)
```
cretz:
```suggestion
            IReadOnlyDictionary<string, string>? overrideEnvVars = null)
```

Here and elsewhere

src/Temporalio/Client/Configuration/ClientConfig.cs
```diff
@@ -0,0 +1,124 @@
...
+            return clientProfile.ToConnectionOptions();
+        }
+
+        /// <summary>
+        /// Deserialize client configuration from JSON.
+        /// </summary>
+        /// <param name="json">JSON string to deserialize.</param>
+        /// <param name="options">Optional JSON serializer options.</param>
+        /// <returns>Client configuration instance.</returns>
+        public static ClientConfig FromJson(string json, JsonSerializerOptions? options = null)
```
cretz: I don't think we need to publicly support JSON serializability of env config in any SDK

src/Temporalio/Client/Configuration/ClientConfig.cs
```diff
@@ -0,0 +1,124 @@
...
+            var clonedProfiles = new Dictionary<string, ClientConfigProfile>();
+            foreach (var kvp in Profiles)
+            {
+                clonedProfiles[kvp.Key] = (ClientConfigProfile)kvp.Value.Clone();
+            }
+            newConfig.Profiles = clonedProfiles;
+            return newConfig;
+        }
+
+        private static JsonSerializerOptions GetDefaultJsonOptions()
```
cretz: Make this a private static readonly field

src/Temporalio/Client/Configuration/ClientConfig.cs
```diff
@@ -0,0 +1,124 @@
...
+
+        /// <summary>
+        /// Serialize this client configuration to JSON.
+        /// </summary>
+        /// <param name="options">Optional JSON serializer options.</param>
+        /// <returns>JSON string representation.</returns>
+        public string ToJson(JsonSerializerOptions? options = null)
+        {
+            return JsonSerializer.Serialize(this, options ?? GetDefaultJsonOptions());
+        }
```
cretz: Just a note, you can shorten these one-line methods:

```suggestion
        public string ToJson(JsonSerializerOptions? options = null) =>
            JsonSerializer.Serialize(this, options ?? GetDefaultJsonOptions());
```

src/Temporalio/Client/Configuration/ClientConfigProfile.cs
```diff
@@ -0,0 +1,188 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text.Json;
+
+namespace Temporalio.Client.Configuration
+{
+    /// <summary>
+    /// Represents a client configuration profile.
+    /// </summary>
+    public sealed class ClientConfigProfile : ICloneable
```
cretz: Mentioned in other place, but this should be a `record` IMO

src/Temporalio/Client/Configuration/ClientConfigProfile.cs
```diff
@@ -0,0 +1,188 @@
...
+            string? profile = null,
+            DataSource? configSource = null,
+            bool disableFile = false,
+            bool disableEnv = false,
+            bool configFileStrict = false,
+            Dictionary<string, string>? overrideEnvVars = null)
+        {
+            if (disableFile && disableEnv)
+            {
+                throw new InvalidOperationException("Cannot disable both file and environment sources - at least one must be enabled");
```
cretz: Pedantic, but this is more of an `ArgumentError`

cretz: General question, what happens if I provide a `configSource` with some data and set `disableFile` as `true`? Does it disregard my `configSource` parameter? Or if not, then it is reasonable to have `disableFile` and `disableEnv` when there is a `configSource` right?

src/Temporalio/Client/Configuration/ClientConfigProfile.cs
```diff
@@ -0,0 +1,188 @@
...
+        {
+            return JsonSerializer.Deserialize<ClientConfigProfile>(json, options ?? GetDefaultJsonOptions())
+                ?? throw new JsonException("Failed to deserialize ClientConfigProfile from JSON.");
+        }
+
+        /// <summary>
+        /// Create a <see cref="TemporalClientConnectOptions"/> from this profile.
+        /// </summary>
+        /// <returns>Connection options for a client.</returns>
+        public TemporalClientConnectOptions ToConnectionOptions()
```
cretz:
```suggestion
        public TemporalClientConnectOptions ToClientConnectOptions()
```

src/Temporalio/Client/Configuration/ClientConfigTls.cs
```diff
@@ -0,0 +1,112 @@
+using System;
+using System.Text.Json;
+
+namespace Temporalio.Client.Configuration
+{
+    /// <summary>
+    /// TLS configuration as specified as part of client configuration.
+    /// </summary>
+    public sealed class ClientConfigTls : ICloneable
```
cretz: Mentioned elsewhere, but let's make this a `record`

src/Temporalio/Client/Configuration/DataSource.cs
```diff
@@ -0,0 +1,151 @@
...
+
+        /// <summary>
+        /// Gets the file path for this data source, if applicable.
+        /// </summary>
+        public string? Path { get; private set; }
+
+        /// <summary>
+        /// Gets the raw data for this data source, if applicable.
+        /// </summary>
+        public byte[]? Data { get; private set; }
```
cretz:
```suggestion
        public string? Path { get; private init; }

        /// <summary>
        /// Gets the raw data for this data source, if applicable.
        /// </summary>
        public byte[]? Data { get; private init; }
```

These should only be mutable on init

src/Temporalio/Client/Configuration/DataSource.cs
```diff
@@ -0,0 +1,151 @@
...
+                throw new ArgumentException("Path cannot be null or empty", nameof(path));
+            }
+
+            return new DataSource { Path = path };
+        }
+
+        /// <summary>
+        /// Create a data source from string content.
+        /// </summary>
+        /// <param name="content">Configuration data as a string.</param>
```
cretz:
```suggestion
        /// <param name="content">Configuration data as a UTF-8 string.</param>
```

src/Temporalio/Client/Configuration/DataSource.cs
```diff
@@ -0,0 +1,151 @@
...
+            }
+
+            return new DataSource { Data = data };
+        }
+
+        /// <summary>
+        /// Create a data source using default configuration discovery.
+        /// </summary>
+        /// <returns>A new data source using default paths.</returns>
+        public static DataSource FromDefault() => new DataSource();
```
cretz: I don't understand this one exactly. What is the use case?

src/Temporalio/Client/Configuration/DataSource.cs
```diff
@@ -0,0 +1,151 @@
...
+
+        /// <summary>
+        /// JSON converter for DataSource objects.
+        /// </summary>
+        internal class JsonConverter : JsonConverter<DataSource>
```
cretz: Mentioned before, but I don't think JSON serializability makes sense for env config

src/Temporalio/Bridge/EnvConfig.cs
```diff
@@ -0,0 +1,416 @@
...
+        /// <param name="overrideEnvVars">Environment variables to use, or null to use system environment.</param>
+        /// <returns>Dictionary of profile name to client configuration profile.</returns>
+        public static Dictionary<string, ClientConfigProfile> LoadClientConfig(
+            Runtime runtime,
+            DataSource source,
+            bool disableFile = false,
+            bool configFileStrict = false,
+            Dictionary<string, string>? overrideEnvVars = null)
+        {
+            var task = LoadClientConfigAsync(
```
cretz: Mentioned in Core review, the C bridge calls here don't need to be async IMO. That may change the structure enough where I'll hold off reviewing this file until that part changes.

.claude/settings.local.json
cretz, line 1: Can this file be removed?

CLAUDE.md
cretz, line 1: Can this file be removed?

cretz (commented):
Is this ready for re-review?

.claude/settings.local.json
THardy98: removed

CLAUDE.md
THardy98: removed

THardy98:
> Is this ready for re-review?

Yes, RFR. Will resolve the above comments

src/Temporalio/Client/Configuration/ClientConfig.cs
THardy98: Renamed

src/Temporalio/Client/Configuration/ClientConfig.cs
THardy98: Yes - so this was intended to be the .NET equivalent of `to_dict/from_dict` in Python. Realized later this makes more sense as `toRecord/fromRecord` than JSON

src/Temporalio/Client/Configuration/ClientConfig.cs
THardy98: removed

src/Temporalio/Client/Configuration/ClientConfig.cs
THardy98: Removed

src/Temporalio/Client/Configuration/ClientConfigProfile.cs
THardy98: Removed - we handle this case lower in core

THardy98: If `disableFile` is `true`, then the `configSource` will be ignored. We return an empty profile object.

THardy98: Here's the code snippet from core:
```
let mut profile = if options.disable_file {
        ClientConfigProfile::default()
    } else {
        // Load the full config
        let config = load_client_config(
            LoadClientConfigOptions {
                config_source: options.config_source,
                config_file_strict: options.config_file_strict,
            },
            env_vars,
        )?;
        ...
    }
```

src/Temporalio/Client/Configuration/DataSource.cs
THardy98: Removed - we want to propagate null data sources to core

src/Temporalio/Client/Configuration/DataSource.cs
THardy98: Removed (in favor of serializing to/from dictionary)

src/Temporalio/Bridge/EnvConfig.cs
THardy98: Yup - refactored

src/Temporalio/Client/EnvConfig/ClientEnvConfig.cs
```diff
@@ -0,0 +1,489 @@
...
+
+            /// <summary>
+            /// Gets the TLS configuration.
+            /// </summary>
+            public Tls? Tls { get; private init; }
+
+            /// <summary>
+            /// Gets the gRPC metadata.
+            /// </summary>
+            public IReadOnlyDictionary<string, string>? GrpcMeta { get; private init; }
```
cretz: Can probably just move all of this to the primary class constructor where you make the `record`. Same for other objects here.

src/Temporalio/Client/EnvConfig/ClientEnvConfig.cs
```diff
@@ -0,0 +1,489 @@
...
+                configFileStrict,
+                overrideEnvVars);
+            return clientProfile.ToClientConnectionOptions();
+        }
+
+        /// <summary>
+        /// Convert to a record structure that can be used for TOML serialization.
+        /// </summary>
+        /// <returns>Dictionary mapping profile names to their record representations.</returns>
+        public IReadOnlyDictionary<string, ProfileRecord> ToRecord()
```
cretz: I wouldn't necessarily call this a "record", maybe `ToDictionary`? Arguably across SDKs this would be a dictionary with a single "profiles" entry if we wanted to properly match TOML, but I see we didn't really do that anywhere.

src/Temporalio/Client/EnvConfig/ClientEnvConfig.cs
```diff
@@ -0,0 +1,489 @@
...
+            return new DataSource { Path = path };
+        }
+
+        /// <summary>
+        /// Create a data source from string content.
+        /// </summary>
+        /// <param name="content">Configuration data as a UTF-8 string.</param>
+        /// <returns>A new data source for the specified data.</returns>
+        /// <exception cref="ArgumentException">Thrown when content is null.</exception>
+        public static DataSource FromString(string content)
```
cretz:
```suggestion
        public static DataSource FromUTF8String(string content)
```

This would be a more .NET like name IMO

src/Temporalio/Client/EnvConfig/ClientEnvConfig.cs
```diff
@@ -0,0 +1,489 @@
...
+    /// <summary>
+    /// Represents the overall client configuration.
+    /// </summary>
+    /// <param name="Profiles">The configuration profiles.</param>
+    public sealed record ClientEnvConfig(IReadOnlyDictionary<string, ClientEnvConfig.ConfigProfile> Profiles)
+    {
+        /// <summary>
+        /// Load client configuration from environment variables and configuration files.
+        /// </summary>
+        /// <param name="configSource">The data source to load from.</param>
```
cretz: Should mention it defaults to a default disk location IMO (don't have to get into detail on where/how). Was a bit confusing at first.

src/Temporalio/Client/EnvConfig/ClientEnvConfig.cs
```diff
@@ -0,0 +1,489 @@
...
+            return clientProfile.ToClientConnectionOptions();
+        }
+
+        /// <summary>
+        /// Convert to a record structure that can be used for TOML serialization.
+        /// </summary>
+        /// <returns>Dictionary mapping profile names to their record representations.</returns>
+        public IReadOnlyDictionary<string, ProfileRecord> ToRecord()
+        {
+            return Profiles.ToDictionary(
```
cretz:
```suggestion
        public IReadOnlyDictionary<string, ProfileRecord> ToRecord() => Profiles.ToDictionary(
```

Pedantic, but remember there's a shortcut for single-line methods

src/Temporalio/Client/EnvConfig/ClientEnvConfig.cs
```diff
@@ -0,0 +1,489 @@
...
+                    {
+                        var domain = Address.Split(':')[0];
+                        options.Tls.Domain = domain;
+                    }
+                }
+                else if (ApiKey != null && (Tls == null || Tls.Disabled != true) && Address != null)
+                {
+                    // If API key is provided but no TLS config (or TLS is not disabled), set up default TLS with domain from address
+                    var domain = Address.Split(':')[0];
+                    options.Tls = new TlsOptions { Domain = domain };
```
cretz: We shouldn't be setting the `Domain` on TLS options, that is for SNI override. If we are doing that in other SDKs, we should fix it IMO. Address check does not need to be in the if, and this same logic should be removed from the TLS section.

src/Temporalio/Client/EnvConfig/ClientEnvConfig.cs
```diff
@@ -0,0 +1,489 @@
...
+                    ApiKey: ApiKey,
+                    Tls: Tls?.ToRecord(),
+                    GrpcMeta: GrpcMeta);
+            }
+        }
+
+        /// <summary>
+        /// TLS configuration as specified as part of client configuration.
+        /// </summary>
+        public sealed record Tls
```
cretz: We qualified "profile" as `ClientProfile` even though it is nested in `ClientEnvConfig`, I think we should either not qualify `ClientProfile` or qualify this as `ClientTls` (I think the former may make more sense).

src/Temporalio/Client/EnvConfig/ClientEnvConfig.cs
```diff
@@ -0,0 +1,489 @@
...
+
+            return null;
+        }
+
+        /// <summary>
+        /// Convert a data source to a dictionary for TOML serialization.
+        /// </summary>
+        /// <param name="source">The data source to convert.</param>
+        /// <returns>A dictionary with either "path" or "data" key, or null if source is null.</returns>
+        public static Dictionary<string, string>? ToDictionary(DataSource? source)
```
cretz: I think this can be a method on the class instead of a static helper

src/Temporalio/Client/EnvConfig/ClientEnvConfig.cs
```diff
@@ -0,0 +1,489 @@
...
+
+            return new DataSource { Data = data };
+        }
+
+        /// <summary>
+        /// Create a data source from a dictionary containing either "data" or "path" keys.
+        /// </summary>
+        /// <param name="dictionary">Dictionary containing configuration data or path.</param>
+        /// <returns>A new data source, or null if the dictionary is null or contains neither "data" nor "path".</returns>
+        public static DataSource? FromDictionary(IReadOnlyDictionary<string, string>? dictionary)
```
cretz, line 408: Pedantic, but IMO the caller can do the null check and this can always accept and always return non-null

src/Temporalio/Client/EnvConfig/ClientEnvConfig.cs
```diff
@@ -0,0 +1,489 @@
...
+        /// <returns>Client connection options.</returns>
+        public static TemporalClientConnectOptions LoadClientConnectOptions(
+            string? profile = null,
+            DataSource? configSource = null,
+            bool disableFile = false,
+            bool disableEnv = false,
+            bool configFileStrict = false,
+            IReadOnlyDictionary<string, string>? overrideEnvVars = null)
+        {
+            var profileName = profile ?? "default";
```
cretz: Hrmm, this doesn't seem quite right. So this is "profile name override" right and if left null, Core will use the `TEMPORAL_PROFILE` env var correct? But if you _always_ pass in this value to Core, how does it know to fall back to the env var? (may need to check other SDKs if we're doing this there too)

src/Temporalio/Bridge/EnvConfig.cs
```diff
@@ -0,0 +1,353 @@
...
+{
+    /// <summary>
+    /// Bridge for environment configuration loading.
+    /// </summary>
+    internal static class EnvConfig
+    {
+        private static readonly JsonSerializerOptions JsonOptions = new()
+        {
+            PropertyNamingPolicy = JsonNamingPolicy.CamelCase,
+            PropertyNameCaseInsensitive = true,
```
cretz: Not sure you have to be forgiving with your own internal JSON format

src/Temporalio/Client/EnvConfig/ClientEnvConfig.cs
```diff
@@ -0,0 +1,489 @@
...
+            /// <param name="configFileStrict">If true, fail if configuration file is invalid.</param>
+            /// <param name="overrideEnvVars">Environment variables to use, or null to use system environment.</param>
+            /// <returns>The loaded profile.</returns>
+            public static ConfigProfile Load(
+                string? profile = null,
+                DataSource? configSource = null,
+                bool disableFile = false,
+                bool disableEnv = false,
+                bool configFileStrict = false,
+                IReadOnlyDictionary<string, string>? overrideEnvVars = null)
```
cretz: Gonna sound a bit annoying, but lots of parameters like this is not that common in .NET. I think we should consider having a `ConfigProfileLoadOptions` (which can also be used in `LoadClientConnectOptions`), and to be consistent we should have a `ClientEnvConfigLoadOptions` for use with `ClientEnvConfig`. Can see other options classes to see how we've done those in .NET (basically cloneable, no required constructor params, all options optional).

src/Temporalio/Bridge/EnvConfig.cs
```diff
@@ -0,0 +1,353 @@
...
+                var envVarsRef = overrideEnvVars?.Count > 0
+                    ? scope.ByteArray(JsonSerializer.Serialize(overrideEnvVars, JsonOptions))
+                    : ByteArrayRef.Empty.Ref;
+
+                unsafe
+                {
+                    var options = new Interop.TemporalCoreClientConfigProfileLoadOptions
+                    {
+                        profile = (sbyte*)profilePtr,
+                        path = (sbyte*)pathPtr,
```
cretz: Mentioned in Core PR, but I think we should use the same string-passing approach we have elsewhere, namely `scope.ByteArray(`. Same for other places in this file.

src/Temporalio/Bridge/EnvConfig.cs
```diff
@@ -0,0 +1,353 @@
...
+                        data = source?.Data != null ? scope.ByteArray(source.Data) : ByteArrayRef.Empty.Ref,
```
cretz:
```suggestion
                        data = scope.ByteArray(source?.Data),
```

It's already built to handle nulls

src/Temporalio/Bridge/EnvConfig.cs
```diff
@@ -0,0 +1,353 @@
...
+            data.TryGetValue(key, out var value) ? value?.ToString() : null;
+
+        private static ClientEnvConfig.Tls? CreateTlsConfig(Dictionary<string, object?> profileData)
+        {
+            if (!profileData.TryGetValue("tls", out var tlsObj))
+            {
+                return null;
+            }
+
+            var tlsData = tlsObj switch
```
cretz: Not sure switch expressions are supported in all versions of C# we support, haven't checked though

src/Temporalio/Bridge/EnvConfig.cs
```diff
@@ -0,0 +1,353 @@
...
+            {
+                Dictionary<string, object?> dict => dict,
+                JsonElement el => JsonSerializer.Deserialize<Dictionary<string, object?>>(el.GetRawText(), JsonOptions),
```
cretz: Which situations is this one vs the other? Doesn't make sense IMO to have a forgiving parser for our own internal JSON format (not _completely_ against the use of JSON to pass across Core, just a bit abnormal).

src/Temporalio/Client/EnvConfig/ClientEnvConfig.cs
```diff
@@ -0,0 +1,489 @@
...
+                    configFileStrict,
+                    overrideEnvVars);
+            }
+
+            /// <summary>
+            /// Create a Profile from a record structure.
+            /// </summary>
+            /// <param name="record">The record to convert from.</param>
+            /// <returns>Profile configuration instance.</returns>
+            public static ConfigProfile FromRecord(ProfileRecord record)
```
cretz: Why a "record" instead of a dictionary? That record won't necessarily serialize to TOML correctly which is the purpose of this IIRC

src/Temporalio/Client/EnvConfig/ClientEnvConfig.cs
THardy98: Oh this is wrong - I had removed this but I might have lost the change in a rebase. Fixing this

src/Temporalio/Client/EnvConfig/ClientEnvConfig.cs
THardy98: I had it as `Profile` originally but .NET complained that it had the same name as `System.something.Profile`, you couldn't name this as `Profile` as well. I opted for `ConfigProfile` because of that, open to `EnvProfile` or `ClientProfile` as well. I think they are all look the same amount of weird after `ClientEnvConfig`

THardy98: Unless there's a way to get around that, I'll rename `Tls` -> `ConfigTls` (or `Client<x>` or `Env<x>` whatever we prefer, I have no preference)

src/Temporalio/Client/EnvConfig/ClientEnvConfig.cs
THardy98: Can't always return non-null, in the case where no `data` or `path` keys exist in the dictionary. It shouldn't happen in principle, but would rather guard this case than not. Given that, I don't really see much value in making it such that callers have to null check

src/Temporalio/Client/EnvConfig/ClientEnvConfig.cs
cretz: > I had it as Profile originally but .NET complained that it had the same name as System.something.Profile, you couldn't name this as Profile as well. [...] Unless there's a way to get around that,

IMO you can silence this warning here since it's a nested class. Everyone will access this as `ClientEnvConfig.Profile` I think. Can you show the error here so I can check the MS page on it? Let me know if not obvious how to use pragmas to disable that check for this line (assuming it is disable-able).

src/Temporalio/Bridge/EnvConfig.cs
THardy98: Yeah - I was having trouble deserializing the incoming JSON properly (nested JSON) so had this as a fallback which worked. Not very nice though.

I've removed this in favor of actual JSON DTO objects, much cleaner.

src/Temporalio/Bridge/EnvConfig.cs
THardy98: Removed

src/Temporalio/Bridge/EnvConfig.cs
THardy98: removed

src/Temporalio/Client/EnvConfig/ClientEnvConfig.cs
THardy98: Removed

src/Temporalio/Client/EnvConfig/ClientEnvConfig.cs
THardy98: Not done in other SDKs, this is a code-gen artifact I missed. Removed all `Domain` setting logic. Thanks for the catch.

src/Temporalio/Client/EnvConfig/ClientEnvConfig.cs
THardy98: Why is it the case that a record won't serialize to TOML correctly? I haven't seen anything on this.

src/Temporalio/Client/EnvConfig/ClientEnvConfig.cs
THardy98: This is noted in the summaries of `ConfigLoadOptions.DataSource` and `ProfileLoadOptions.DataSource`

src/Temporalio/Client/EnvConfig/ClientEnvConfig.cs
cretz: Depends on the TOML library in use on how it handles PascalCased properties and such, but even if it can, users will have to know to set the setting to convert to underscore-based names and such. If you give/accept a dictionary (of dictionaries/lists/primitives), you can control the key names. And as a bonus you don't have to have new classes.

src/Temporalio/Client/EnvConfig/ClientEnvConfig.cs
THardy98: Changed to `ToDictionary`.

wrt changing to a single "profiles" entry, if it's something we care about I can open some PRs once this is in

src/Temporalio/Client/EnvConfig/ClientEnvConfig.cs
THardy98: Sweet - suppressed the warning. Much nicer.

src/Temporalio/Client/EnvConfig/ClientEnvConfig.cs
THardy98: Have changed this to use dictionaries instead of records

src/Temporalio/Client/EnvConfig/ClientEnvConfig.cs
```diff
@@ -0,0 +1,414 @@
...
+        /// <summary>
+        /// Represents a client configuration profile.
+        /// </summary>
+        /// <param name="Address">Client address.</param>
+        /// <param name="Namespace">Client namespace.</param>
+        /// <param name="ApiKey">Client API key.</param>
+        /// <param name="Tls">TLS configuration.</param>
+        /// <param name="GrpcMeta">gRPC metadata.</param>
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Naming", "CA1724:Type names should not match namespaces", Justification = "Profile is appropriate name in this context and unlikely to cause confusion")]
+        public sealed record Profile(
```
cretz:
```suggestion
#pragma warning disable CA1724 // We're ok with the Profile name since it's a nested class
        public sealed record Profile(
#pragma warning restore CA1724
```

I think this is the better/clearer way

src/Temporalio/Client/EnvConfig/ClientEnvConfig.cs
```diff
@@ -0,0 +1,414 @@
...
+            var profiles = Bridge.EnvConfig.LoadClientConfig(runtime, options);
+            return new ClientEnvConfig(profiles);
+        }
+
+        /// <summary>
+        /// Load client connection options directly from configuration.
+        /// </summary>
+        /// <param name="options">Options for loading the configuration profile.</param>
+        /// <returns>Client connection options.</returns>
+        public static TemporalClientConnectOptions LoadClientConnectOptions(ProfileLoadOptions options)
```
cretz:
```suggestion
        public static TemporalClientConnectOptions LoadClientConnectOptions(ProfileLoadOptions? options = null)
```

We should make the options parameter optional on all of these if the default `new()` is acceptable. Can do `options ??= new()` in the method just before use if needed to be non-null in code (or something like that). But I think we want our samples code to just be able to `TemporalClient.ConnectAsync(ClientEnvConfig.LoadClientConnectOptions())`

src/Temporalio/Client/EnvConfig/ClientEnvConfig.cs
```diff
@@ -0,0 +1,414 @@
...
+    public sealed record ClientEnvConfig(IReadOnlyDictionary<string, ClientEnvConfig.Profile> Profiles)
+    {
+        /// <summary>
+        /// Load client configuration from environment variables and configuration files.
+        /// </summary>
+        /// <param name="options">Options for loading the configuration.</param>
+        /// <returns>Loaded configuration data.</returns>
+        public static ClientEnvConfig Load(ConfigLoadOptions options)
+        {
+            var runtime = Runtime.TemporalRuntime.Default.Runtime;
```
cretz: Ok, sorry I am just noticing this. So we should not make this `.Default` call to runtime. It lazily creates a thread pool and does other things. The only reason runtime seems like it is needed is for freeing byte arrays. But that is just an added optional optimization I put in there if I wanted one day to come back and reuse byte array buffers. Not only is the `return_buf` operation not even implemented, but `temporal_core_byte_array_free` actually accepts a null pointer for runtime, which we should use in this case.

I only required runtimes for freeing byte arrays because I always had a runtime available. But this is the first use case where it doesn't make. I think we should:

1. Update `ByteArray` constructor/field to accept `Runtime?` instead of `Runtime`
2. Remove `Runtime.FreeByteArray` method altogether
3. Update `ByteArray.ReleaseHandle` to explicitly invoke `Interop.Methods.temporal_core_byte_array_free` with the first param as `Runtime?.Ptr` so it can be null
4. Update your instantiation of `ByteArray` to pass in `null` for the runtime

Sorry about just realizing this

src/Temporalio/Client/EnvConfig/ClientEnvConfig.cs
```diff
@@ -0,0 +1,414 @@
...
+                }
+
+                return options;
+            }
+
+            /// <summary>
+            /// Convert to a dictionary structure that can be used for TOML serialization.
+            /// </summary>
+            /// <returns>Dictionary representation of this profile.</returns>
+            public Dictionary<string, object?> ToDictionary() =>
```
cretz:
```suggestion
            public Dictionary<string, object> ToDictionary() =>
```

Not sure any of these values in these dictionaries should be nullable. The key should be absent of the value is null IMO.

src/Temporalio/Client/EnvConfig/ClientEnvConfig.cs
```diff
@@ -0,0 +1,414 @@
...
+        /// Options for loading a specific client configuration profile.
+        /// </summary>
+        public class ProfileLoadOptions : ICloneable
+        {
+            /// <summary>
+            /// Initializes a new instance of the <see cref="ProfileLoadOptions"/> class.
+            /// </summary>
+            public ProfileLoadOptions()
+            {
+            }
```
cretz:
```suggestion
```

Default constructor is implied when there are no other constructors (same for other options class)

cretz (commented):
Few more things, but I think we're basically there

src/Temporalio/Client/EnvConfig/ClientEnvConfig.cs
```diff
@@ -0,0 +1,414 @@
...
+            {
+                return new Dictionary<string, string> { ["path"] = this.Path! };
+            }
+
+            if (this.Data != null)
+            {
+                return new Dictionary<string, string> { ["data"] = Encoding.UTF8.GetString(this.Data) };
+            }
+
+            return null;
```
cretz, line 444: It is impossible to reach this line, so arguably this method should never return null (and therefore to my other comment, I wonder if we should never accept a null dictionary or return null data source from dictionary, but can leave that if you really want).

src/Temporalio/Client/EnvConfig/ClientEnvConfig.cs
```diff
@@ -0,0 +1,414 @@
...
+        }
+
+        /// <summary>
+        /// Convert a data source to a dictionary for TOML serialization.
+        /// </summary>
+        /// <returns>A dictionary with either "path" or "data" key, or null if neither key exists.</returns>
+        public Dictionary<string, string>? ToDictionary()
+        {
+            if (!string.IsNullOrEmpty(this.Path))
```
cretz: Shouldn't have to use `this.` in these cases

src/Temporalio/Client/EnvConfig/ClientEnvConfig.cs
THardy98: Removed `runtime` usage

src/Temporalio/Client/EnvConfig/ClientEnvConfig.cs
THardy98: options are nullable and use fallback defaults

cretz (approved):

PR #511, opened 2025-08-27
Added retry policy to activity info

### LINKED ISSUE #457: [Feature Request] Show retry policy on activity info
**Describe the solution you'd like**

See https://github.com/temporalio/features/issues/615

maciejdudko:
## What was changed
Added retry policy to activity info.

## Why?

Feature request: https://github.com/temporalio/features/issues/615

## Checklist

1. Closes #457

2. How was this tested:

Added new assertion to `Temporalio.Tests.Worker.ActivityWorkerTests.ExecuteActivityAsync_CheckInfo_IsAccurate`

### DIFF
--- a/src/Temporalio/Activities/ActivityInfo.cs
+++ b/src/Temporalio/Activities/ActivityInfo.cs
@@ -20,6 +20,10 @@ namespace Temporalio.Activities
     /// <param name="HeartbeatTimeout">Heartbeat timeout set by the caller.</param>
     /// <param name="IsLocal">Whether the activity is a local activity or not.</param>
     /// <param name="Priority">The Priority of this activity.</param>
+    /// <param name="RetryPolicy">
+    /// The retry policy of this activity.
+    /// The server may set a different policy than the one provided when scheduling the activity.
+    /// </param>
     /// <param name="ScheduleToCloseTimeout">Schedule to close timeout set by the caller.</param>
     /// <param name="ScheduledTime">When the activity was scheduled.</param>
     /// <param name="StartToCloseTimeout">Start to close timeout set by the caller.</param>
@@ -44,6 +48,7 @@ namespace Temporalio.Activities
         TimeSpan? HeartbeatTimeout,
         bool IsLocal,
         Temporalio.Common.Priority Priority,
+        Temporalio.Common.RetryPolicy? RetryPolicy,
         TimeSpan? ScheduleToCloseTimeout,
         DateTime ScheduledTime,
         TimeSpan? StartToCloseTimeout,
--- a/src/Temporalio/Activities/ActivityInfo.cs
+++ b/src/Temporalio/Activities/ActivityInfo.cs
@@ -22,7 +22,11 @@ namespace Temporalio.Activities
     /// <param name="Priority">The Priority of this activity.</param>
     /// <param name="RetryPolicy">
     /// The retry policy of this activity.
-    /// The server may set a different policy than the one provided when scheduling the activity.
+    /// <para>
+    /// Note that the server may have set a different policy than the one provided when scheduling the activity.
+    /// If the value is null, it means the server didn't send information about retry policy (e.g. due to old server
+    /// version), but it may still be defined server-side.
+    /// </para>
     /// </param>
     /// <param name="ScheduleToCloseTimeout">Schedule to close timeout set by the caller.</param>
     /// <param name="ScheduledTime">When the activity was scheduled.</param>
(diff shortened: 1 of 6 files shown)


src/Temporalio/Common/RetryPolicy.cs
```diff
@@ -65,7 +65,7 @@ internal static RetryPolicy FromProto(Api.Common.V1.RetryPolicy proto)
         {
             return new()
             {
-                InitialInterval = proto.InitialInterval.ToTimeSpan(),
+                InitialInterval = (proto.InitialInterval ?? new Duration()).ToTimeSpan(),
```
cretz: Can you explain this change a bit? Is this sometimes null from the server? What does the `InitialInterval` in this PR end up being when that happens?

maciejdudko: Without this change, the test `Temporalio.Tests.Worker.WorkflowWorkerTests.ExecuteWorkflowAsync_SimpleActivity_ExecutesProperly(local=true)` hangs up indefinitely due to NullReferenceException on `InitialInterval`. The value is null because default-constructed `Api.Common.V1.RetryPolicy` has it as null (`Duration` is a message type). I am still trying to find the root cause why Core SDK uses default-constructed `RetryPolicy` here - we may have a bug in local activity start message somewhere. But even if we fixed that part, I think we still don't want `RetryPolicy.FromProto` to crash when `InitialInterval` is null for whatever reason.

In this case, `InitialInterval` is initialized to zero. Alternatively we could use the fields default value (1 second) but I'm not a fan of it - it would make it impossible to differentiate missing/malformed RetryPolicy and a properly initialized RetryPolicy with default interval. Third alternative is to make `FromProto` return null.

cretz: > I am still trying to find the root cause why Core SDK uses default-constructed RetryPolicy here - we may have a bug in local activity start message somewhere

I think it's worth understanding the situation where `initial_interval` is _ever_ null from Core because I think we have several SDKs that assume it is _never_ null from Core. I don't think a zero initial interval is even considered valid for a retry policy (not sure .NET validates, but I see such a validation in Python) and I wouldn't want to set it some default value that doesn't relate to anything real.

maciejdudko: Regardless of Core behavior, should we validate and return null on failed validation here, or should we throw?

cretz: IMO we should throw if we get unexpected values (same as we have until now). I have never seen this line throw before and IMO we have to be able to make assumptions in dozens of places about nullability of proto fields (or lack thereof) without manually checking each time.

maciejdudko: Restored original behavior. After https://github.com/temporalio/sdk-core/pull/1012, Core guarantees `initial_interval` is not null.

maciejdudko:
The code in this PR is finalized, but it requires updating Core to a version that includes https://github.com/temporalio/sdk-core/pull/1012. There's an incompatibility in C bridge lib related to Nexus, so I'll postpone this PR until after we merge Nexus support in #517.

src/Temporalio/Activities/ActivityInfo.cs
```diff
@@ -20,6 +20,14 @@ namespace Temporalio.Activities
     /// <param name="HeartbeatTimeout">Heartbeat timeout set by the caller.</param>
     /// <param name="IsLocal">Whether the activity is a local activity or not.</param>
     /// <param name="Priority">The Priority of this activity.</param>
+    /// <param name="RetryPolicy">
+    /// The retry policy of this activity.
+    /// <para>
+    /// Note that the server may have set a different policy than the one provided when scheduling the activity.
+    /// If the value is null, it means the server didn't send information about retry policy (e.g. due to old server
+    /// version), but it may still be defined server-side.
```
cretz, line 28:
```suggestion
    /// version), but it may still be defined server-side. It may also be null if testing with an activity testing environment.
```

Not that important, just something I noticed reading code

cretz (approved):

src/Temporalio/Activities/ActivityInfo.cs
maciejdudko: Or maybe I'll just fill in default retry policy.

cretz: I think null is ok here. Users can of course customize this if they need to for testing their activity.

maciejdudko: Changed testing environment to initialize retry policy, see comment at https://github.com/temporalio/sdk-ruby/pull/341#discussion_r2363114625

Issue #515, opened 2025-08-29
[Feature Request] Nexus OpenTelemetry support


cretz:
### Describe the solution you'd like

Like tracing work done at https://github.com/temporalio/sdk-java/pull/2278, but for OTel .NET

Issue #514, opened 2025-08-29
[Feature Request] Nexus cancellation type support


cretz:
### Describe the solution you'd like

Need to do https://github.com/temporalio/sdk-python/pull/981, but for .NET

cretz:
Note, there is an issue with Core this found and is waiting on fix for: https://github.com/temporalio/sdk-core/issues/1026

PR #517, opened 2025-08-29
Nexus support

### LINKED ISSUE #426: [Feature Request] Initial Nexus implementation

Alongside Nexus .NET SDK, need Temporal-specific implementation in this SDK. See https://github.com/temporalio/features/issues/609

cretz:
## What was changed

Added Nexus support. Notes:

* Nexus PR that must be merged/published first at https://github.com/nexus-rpc/sdk-dotnet/pull/4
* 💥 Slight compatibility break when manually instantiating a worker tuner where we require a 4th parameter to constructor now for Nexus
* Issue opened to do samples at https://github.com/temporalio/samples-dotnet/issues/104
* Issue opened for cancellation types at #514
* Issue opened for OTel support at #515
* Issue opened for Core to dedupe headers at https://github.com/temporalio/sdk-core/issues/993
* Issue that will encompass Nexus slot supplier tests at https://github.com/temporalio/sdk-dotnet/issues/519

## Checklist

1. Closes #426


src/Temporalio/Temporalio.csproj
```diff
@@ -15,12 +15,13 @@
   <ItemGroup>
     <PackageReference Include="Google.Protobuf" Version="3.26.1" />
     <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="2.2.0" />
-    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.1.1" PrivateAssets="All" />
+    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="8.0.0" PrivateAssets="All" />
+    <ProjectReference Include="..\..\..\nexus-sdk-dotnet\src\NexusRpc\NexusRpc.csproj" />
```
cretz: Won't pass CI until https://github.com/nexus-rpc/sdk-dotnet/pull/4 merged/published

src/Temporalio/Nexus/NexusOperationInfo.cs
```diff
@@ -0,0 +1,12 @@
+namespace Temporalio.Nexus
+{
+    /// <summary>
+    /// Temporal-specific Nexus operation information.
+    /// </summary>
+    /// <param name="Namespace">Current namespace.</param>
+    /// <param name="TaskQueue">Current task queue.</param>
+    /// <remarks>WARNING: Nexus support is experimental.</remarks>
+    public record NexusOperationInfo(
```
Quinn-With-Two-Ns, line 9: Is this exposed to users? I am not sure we exposed anything similar in Go or Java? Not saying we shouldn't just if we do we should open issue in Go and Java for parity.

src/Temporalio/Testing/WorkflowEnvironment.cs
```diff
@@ -209,6 +210,29 @@ public Task WithAutoTimeSkippingDisabledAsync(Func<Task> func) =>
...
         public virtual Task<T> WithAutoTimeSkippingDisabledAsync<T>(Func<Task<T>> func) => func();
 
+        /// <summary>
+        /// Create Nexus endpoint on this test environment.
+        /// </summary>
+        /// <param name="name">Endpoint name.</param>
+        /// <param name="taskQueue">Task queue.</param>
+        /// <returns>Created endpoint.</returns>
+        /// <remarks>WARNING: Nexus support is experimental.</remarks>
+        public async Task<Endpoint> CreateNexusEndpointAsync(string name, string taskQueue)
```
Quinn-With-Two-Ns, line 220: nit: may want to add a delete helper as well if a user is using a test server for multiple tests and wants to clean up in between runs. We have one in Java

src/Temporalio/Nexus/NexusOperationInfo.cs
cretz: This info was exposed in Python by @dandavison. We can make this one and that one internal only or we can open Go/Java/TypeScript issues. Do you have a preference/vote?

src/Temporalio/Testing/WorkflowEnvironment.cs
cretz: 👍 Will add, though I'd hope anyone using ephemeral test servers doesn't mind leaving Nexus endpoints around (or namespaces, task queues with tasks, etc).

src/Temporalio/Worker/WorkflowInstance.cs
```diff
@@ -2452,6 +2490,106 @@ public override Task<ChildWorkflowHandle<TWorkflow, TResult>> StartChildWorkflow
                 return handleSource.Task;
             }
 
+            /// <inheritdoc/>
+            public override Task<NexusOperationHandle<TResult>> StartNexusOperationAsync<TResult>(
+                StartNexusOperationInput input)
+            {
+                var token = input.Options.CancellationToken ?? instance.CancellationToken;
+                // We do not even want to schedule if the cancellation token is already cancelled.
+                // We choose to use cancelled failure instead of wrapping in child failure which is
```
Quinn-With-Two-Ns:
```suggestion
                // We choose to use cancelled failure instead of wrapping in nexus operation failure which is
```

I think this is what you mean?

src/Temporalio/Worker/WorkflowInstance.cs
```diff
@@ -2722,6 +2860,118 @@ public override Task CancelAsync() =>
...
+
+            public override Task<NexusOperationHandle<TResult>> StartNexusOperationAsync<TResult>(
+                string operationName, object? arg, NexusOperationOptions? options = null) =>
+                instance.outbound.Value.StartNexusOperationAsync<TResult>(new(
+                    Service: Service,
+                    ClientOptions: Options,
+                    OperationName: operationName,
+                    Arg: arg,
+                    Options: options ?? new(),
+                    Headers: null));
```
Quinn-With-Two-Ns, line 2887: Is there a reason you pass `null` headers vs an empty map? If that is what we did in other places in the dotnet SDK not suggesting we change, just curious.

src/Temporalio/Worker/WorkflowInstance.cs
cretz: Yes, this is a typo via copy pasta from child method above

src/Temporalio/Worker/WorkflowInstance.cs
cretz: Yes, it is what we did in other places. .NET devs are self-conscious about unnessesarily creating new objects if they don't have to and they aren't used often. For headers, to create an empty mutable dictionary would be extra unnecessary object creation in most cases (and immutable dictionary w/ copy-on-write semantics is not a common .NET approach here). Granted .NET has good short-lived-object GC like Java these days so it's not a big deal, but it's more proper to allow nullable here (which can be expressed via the type system so users are not surprised).

tests/Temporalio.Tests/Worker/NexusWorkerTests.cs
```diff
@@ -0,0 +1,823 @@
...
+        });
+    }
+
+    [Fact]
+    public async Task ExecuteNexusOperationAsync_ApplicationFailure_NonRetryable()
+    {
+        var workerOptions = new TemporalWorkerOptions($"tq-{Guid.NewGuid()}").
+            AddNexusService(new HandlerFactoryStringService(() =>
+                OperationHandler.Sync<string, string>(async (context, input) =>
+                // WorkflowRunOperationHandler.FromHandleFactory<string, string>(async (context, input) =>
```
Quinn-With-Two-Ns: What is this comment for?

src/Temporalio/Worker/NexusWorker.cs
```diff
@@ -0,0 +1,447 @@
...
+            }
+            else if (exc is ApplicationFailureException appExc && appExc.NonRetryable)
+            {
+                return new(
+                    HandlerErrorType.Internal,
+                    "Handler failed with non-retryable application error",
+                    appExc,
+                    HandlerErrorRetryBehavior.NonRetryable);
+            }
+            else if (exc is WorkflowAlreadyStartedException)
```
Quinn-With-Two-Ns, line 387: Is there any test for this behaviour? Would be good to have some tests for the general RPC exception to handler translation, but IMO this is the more important error to translate since most async nexus operations are just starting workflows

tests/Temporalio.Tests/Worker/NexusWorkerTests.cs
cretz: Copy/pasta from other tests, will remove

src/Temporalio/Worker/NexusWorker.cs
cretz: :+1: No, will add test. The tests I had done so far on dupe checking were more about checking conflict policy. I will add a test that confirms regular start will fail as non-retryable.

src/Temporalio/Nexus/NexusOperationInfo.cs
Quinn-With-Two-Ns: I am fine with adding it, might be worth having a way for a user to detect they are in a Nexus operation context like we have for activities

Quinn-With-Two-Ns (commented):
other then my previous comments LGTM

src/Temporalio/Nexus/NexusOperationInfo.cs
cretz: Ok, setting reminder to open issues in Go, Java, and TypeScript to ensure it is included there too

cretz: I have opened https://github.com/temporalio/features/issues/674 to track it in general

src/Temporalio/Worker/NexusWorker.cs
cretz: Actually, just realized we do have a test that confirms this, it is part of the `ExecuteNexusOperationAsync_SimpleWorkflow_ConflictPolicy` test (the test confirms already-started behavior before it confirms conflict-policy-use-existing behavior)

src/Temporalio/Testing/WorkflowEnvironment.cs
cretz: Added

src/Temporalio/Worker/WorkflowInstance.cs
cretz: Fixed

tests/Temporalio.Tests/Worker/NexusWorkerTests.cs
cretz: Removed

src/Temporalio/Worker/WorkflowInstance.cs
```diff
@@ -2453,6 +2491,106 @@ public override Task<ChildWorkflowHandle<TWorkflow, TResult>> StartChildWorkflow
...
+                }
+
+                var seq = ++instance.nexusOperationCounter;
+                var cmd = new ScheduleNexusOperation()
+                {
+                    Seq = seq,
+                    Endpoint = input.ClientOptions.Endpoint,
+                    Service = input.Service,
+                    Operation = input.OperationName,
+                    Input = input.Arg == null ? null : instance.PayloadConverter.ToPayload(input.Arg),
```
Quinn-With-Two-Ns, line 2517: to confirm this `PayloadConverter` does not have the workflow context attached correct?

src/Temporalio/Worker/WorkflowInstance.cs
```diff
@@ -2453,6 +2491,106 @@ public override Task<ChildWorkflowHandle<TWorkflow, TResult>> StartChildWorkflow
...
             }
 
+            /// <inheritdoc/>
+            public override Task<NexusOperationHandle<TResult>> StartNexusOperationAsync<TResult>(
+                StartNexusOperationInput input)
+            {
+                var token = input.Options.CancellationToken ?? instance.CancellationToken;
+                // We do not even want to schedule if the cancellation token is already cancelled.
+                // We choose to use cancelled failure instead of wrapping in Nexus failure which is
+                // similar to what Java and TypeScript do, with the accepted tradeoff that it makes
```
Quinn-With-Two-Ns, line 2501:
```suggestion
                // similar to what TypeScript do, with the accepted tradeoff that it makes
```
Java does wrap the cancelled failure in a Nexus failure. What does dotnet do in other immediate cancellation scenarios ?

src/Temporalio/Worker/WorkflowInstance.cs
cretz: It does as do all workflow calls. Based on off-PR discussions ,I have opened a separate issue at #523 to solve this generally (it is not a Nexus-specific concern).

src/Temporalio/Worker/WorkflowInstance.cs
cretz: To clarify, this is copied from the child workflow comment. I was under the impression for child/nexus cancel-before-start, that Java and TypeScript just throw (i.e. set promise result to) a `CanceledFailure`, not a wrapped cancel failure as would happen if the cancel happened after schedule/start.

src/Temporalio/Worker/WorkflowInstance.cs
Quinn-With-Two-Ns: I think we should fix this for Nexus to not make the problem worse

cretz: We can't just fix in Nexus, we have to fix everywhere. It is a general fix, not a Nexus specific fix. I do not think we should hold up experimental Nexus support for a general fix to this problem.

src/Temporalio/Worker/WorkflowInstance.cs
Quinn-With-Two-Ns: >I was under the impression for child/nexus cancel-before-start, that Java and TypeScript just throw (i.e. set promise result to) a CanceledFailure

That is not true for Java, cancel-before-start throws the same error chain as cancel happened after schedule/start.

Quinn-With-Two-Ns (approved):

PR #525, opened 2025-09-24
💥 Fix how serialization context is applied in workflows [MINOR COMPAT BREAK]

### LINKED ISSUE #519: [Feature Request] Ensure tests exist to confirm custom slot supplier slot info has proper fields

See https://github.com/temporalio/features/issues/672. Make sure Nexus is included.

### LINKED ISSUE #523: [Bug] Serialization context incorrectly applied to all workflow calls
### Describe the bug

As part of #446 we assumed every use of data conversion in a .NET workflow context needed a .NET workflow context. But to match Java, we actually need to only selectively apply contexts based on their situation. This applies to payload converter, failure converter, and payload codec. We will have to decide whether we want the context present on the payload/failure converter we expose to users.

This is technically a backwards incompatible change as we had not explicitly marked the .NET serialization context addition as experimental, so the implementer will have to decide how best to handle that.

cretz:
## What was changed

Previously, we applied context-based codecs/converters to everything a workflow does, and then if a workflow _also_ did anything needing more context (e.g. activity or child workflow calls), we applied a context on top of that. Per #523, this was an incorrect approach. We need to apply contexts only in serialization/deserialization situations where the exact same context can be applied on the other side with deserialization/serialization.

There are no test adjustments for this PR because the test coverage from #446 still applies, this just affects the edge case of stacking the codecs/converters.

This is a 💥 breaking change _technically_ because in situations where we used a specific context on top of outer-workflow context before (e.g. working with activity or child), we only apply the specific context now instead. This will only affect users with custom codecs/converters that support serialization contexts and built an expectation on the workflow-context one to be used for _everything_ in a workflow.

## Checklist

1. Closes #523
2. Closes #519

### DIFF
--- a/src/Temporalio/Worker/WorkflowCodecHelper.cs
+++ b/src/Temporalio/Worker/WorkflowCodecHelper.cs
@@ -20,25 +20,25 @@ namespace Temporalio.Worker
         /// <summary>
         /// Encode the completion.
         /// </summary>
-        /// <param name="codec">Codec to use.</param>
-        /// <param name="codecContext">Codec context.</param>
+        /// <param name="context">Codec context.</param>
         /// <param name="comp">Completion to encode.</param>
         /// <returns>Task for completion.</returns>
         internal static async Task EncodeAsync(
-            IPayloadCodec codec, WorkflowCodecContext codecContext, WorkflowActivationCompletion comp)
+            WorkflowCodecContext context, WorkflowActivationCompletion comp)
         {
             switch (comp.StatusCase)
             {
                 case WorkflowActivationCompletion.StatusOneofCase.Failed:
-                    if (comp.Failed.Failure_ != null)
+                    if (comp.Failed.Failure_ != null && context.CodecWorkflowContext != null)
                     {
-                        await codec.EncodeFailureAsync(comp.Failed.Failure_).ConfigureAwait(false);
+                        await context.CodecWorkflowContext.EncodeFailureAsync(
+                            comp.Failed.Failure_).ConfigureAwait(false);
                     }
                     break;
                 case WorkflowActivationCompletion.StatusOneofCase.Successful:
                     foreach (var cmd in comp.Successful.Commands)
                     {
-                        await EncodeAsync(codec, codecContext, cmd).ConfigureAwait(false);
+                        await EncodeAsync(context, cmd).ConfigureAwait(false);
                     }
                     break;
             }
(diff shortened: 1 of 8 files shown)


dandavison:
> There are no test adjustments for this PR because the test coverage from https://github.com/temporalio/sdk-dotnet/pull/446 still applies, this just affects the edge case of stacking the codecs/converters.

In Python I'm adding a test that demonstrates that every key in the context on the outbound side is available for decryption in the corresponding context on the inbound side, for a kitchen-sink-ish workflow. I'm not suggesting that's necessary to add here. But seeing as this PR does fix a genuine issue, it could be good to have a test that proves that an outbound Nexus payload now lacks any context and hence context cannot be used to encrypt in an un-decryptable way.

cretz:
👍 I adjusted the .NET kichen-sink-ish converter context test to check that Nexus is contextless (the first contextless use of a converter in workflows, so had to adjust a bit).

src/Temporalio/Worker/WorkflowCodecHelper.cs
```diff
@@ -47,91 +47,112 @@ internal static async Task EncodeAsync(
...
+                            if (childCodec2 != null)
+                            {
+                                await childCodec2.DecodeFailureAsync(
+                                    job.ResolveChildWorkflowExecutionStart.Cancelled.Failure).
+                                    ConfigureAwait(false);
+                            }
                         }
                         break;
                     case WorkflowActivationJob.VariantOneofCase.ResolveNexusOperation:
+                        // TODO(cretz): Support Nexus serialization context
```
maciejdudko, line 120: Nit: add issue link for Nexus serialization context

src/Temporalio/Worker/WorkflowCodecHelper.cs
```diff
@@ -47,91 +47,112 @@ internal static async Task EncodeAsync(
...
                         {
-                            await codec.DecodeFailureAsync(
+                            await context.CodecNoContext.DecodeFailureAsync(
                                 job.ResolveNexusOperation.Result.TimedOut).
                                 ConfigureAwait(false);
                         }
                         break;
                     case WorkflowActivationJob.VariantOneofCase.ResolveNexusOperationStart:
-                        if (job.ResolveNexusOperationStart.Failed != null)
+                        // TODO(cretz): Support Nexus serialization context
```
maciejdudko, line 151: Nit: add issue link for Nexus serialization context

maciejdudko (approved):
Looks good! We should add an issue link to TODO comments for Nexus serialization context (and create issue if we haven't yet) but even without that, it's ready to merge. LGTM!

src/Temporalio/Worker/WorkflowCodecHelper.cs
cretz: Yeah, arguably this shouldn't be a TODO, more of a "here is where you'd add it". We do not support Nexus serialization context at this time in SDKs and not sure when we will (it's not really a bug, we may never).

PR #529, opened 2025-09-25
Fix issue with search attribute keyword list type

### LINKED ISSUE #526: [Bug] KeywordList cannot be resolved as part of WorkflowExecutionDescription

I am trying to get a value of a custom search attribute of a `KeywordList` type once the workflow is completed.

### Describe the bug

When executing a code as following, the `TryGetValue` returns `false` even the value is there, but of a `List<object>` type.

```
var description = await workflowHandle.DescribeAsync();

if (description.TypedSearchAttributes.TryGetValue(MySearchAttributes.PetNames, out var petNames))
{}
```

### Minimal Reproduction

Can be found in this commit: https://github.com/temporalio/samples-dotnet/commit/9c72c609b02578209749b661535518c869118bff

### Environment/Versions

- OS and processor: M4 Mac
- Temporal Version: 1.7.0, 1.8.0
- Are you using Docker or Kubernetes or building Temporal from source? No

cretz:
## What was changed

* Made sure all representations of search attribute keyword list values (coming from user or deserialized from server) are `List<string>`
* Fixed tests to start including search attribute keyword lists (it was marked as a TODO because older dev server didn't support them on start)

## Checklist

1. Closes #526

### DIFF
--- a/src/Temporalio/Common/SearchAttributeCollection.cs
+++ b/src/Temporalio/Common/SearchAttributeCollection.cs
@@ -25,7 +25,7 @@ namespace Temporalio.Common
         /// <summary>
         /// An empty search attribute collection.
         /// </summary>
-        public static readonly SearchAttributeCollection Empty = new(new());
+        public static readonly SearchAttributeCollection Empty = new();
 
         private static readonly Dictionary<ByteString, IndexedValueType> NameToIndexType =
             Enum.GetValues(typeof(IndexedValueType)).Cast<IndexedValueType>().ToDictionary(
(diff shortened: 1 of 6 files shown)


src/Temporalio/Common/SearchAttributeCollection.cs
```diff
@@ -41,9 +41,12 @@ public class SearchAttributeCollection : IReadOnlyCollection<SearchAttributeKey>
...
-        /// <param name="values">The values that are actually used. Warning, this is not copied.
-        /// </param>
-        internal SearchAttributeCollection(SortedDictionary<SearchAttributeKey, object> values) =>
+        internal SearchAttributeCollection()
+            : this(new())
+        {
+        }
+
+        private SearchAttributeCollection(SortedDictionary<SearchAttributeKey, object> values) =>
             this.values = values;
```
maciejdudko, line 50: Nit: the private constructor is only used once, in the internal constructor. It can be removed and initialization can be done directly in the internal constructor (or even better, inline in the field declaration).

maciejdudko (approved):
Looks good! I left one minor suggestion. LGTM!

src/Temporalio/Common/SearchAttributeCollection.cs
cretz: This private constructor is also used in `FromProto` and `ToSearchAttributeCollection`

PR #531, opened 2025-09-30
Minor fix for Nexus worker validation

cretz:
## What was changed

Added minor fix to allow Nexus-only workers


Sushisource (approved):

cretz:
Waiting to merge this because we may not merge it since it is small and is incorporated into another PR coming concerning cancellation types

jmaeagle99:
Thanks for putting this together. The Nexus-only worker fix landed via #698, which made the same change (with an added test in GeneralWorkerTests). Closing this one in favor of that.

PR #532, opened 2025-10-01
Fixed memory corruption issues in CustomSlotSupplier

### LINKED ISSUE #458: [Bug] Issues with ReserveCtxFromBridge

maciejdudko:
## What was changed

https://github.com/temporalio/sdk-core/pull/1025 changes the design of custom slot supplier in C bridge to fix memory corruption issues. This PR adapts SDK code to these changes.

## Why?

Fixes memory corruption bug.

## Checklist

1. Closes #458

2. How was this tested:

**There is no test for the original bug.** Due to the nature of this bug, it's impossible to write a test that reliably triggers the problematic conditions. The fix can be verified over time by the lack of transient CI failures caused by this bug.

General functionality of the custom slot supplier is tested by pre-existing tests.

### DIFF
--- a/src/Temporalio/Bridge/CustomSlotSupplier.cs
+++ b/src/Temporalio/Bridge/CustomSlotSupplier.cs
@@ -38,6 +38,7 @@ namespace Temporalio.Bridge
                 try_reserve = FunctionPointer<Interop.TemporalCoreCustomSlotSupplierTryReserveCallback>(TryReserve),
                 mark_used = FunctionPointer<Interop.TemporalCoreCustomSlotSupplierMarkUsedCallback>(MarkUsed),
                 release = FunctionPointer<Interop.TemporalCoreCustomSlotSupplierReleaseCallback>(Release),
+                available_slots = FunctionPointer<Interop.TemporalCoreCustomSlotSupplierAvailableSlotsCallback>(AvailableSlots),
                 free = FunctionPointer<Interop.TemporalCoreCustomSlotSupplierFreeCallback>(Free),
             };
 
@@ -265,6 +266,26 @@ namespace Temporalio.Bridge
                     permits.Remove(permitId);
                 }
             }
+        }
+
+        private unsafe bool AvailableSlots(UIntPtr* availableSlots, void* userData)
+        {
+            try
+            {
+                if (userSupplier.AvailableSlots is { } slots)
+                {
+                    *availableSlots = new(slots);
+                    return true;
+                }
+            }
+#pragma warning disable CA1031 // We are ok catching all exceptions here
+            catch (Exception e)
+            {
+#pragma warning restore CA1031
+                logger.LogError(e, "Error querying available slots");
+            }
+
+            return false;
         }
 
         // this method must be called under a lock
--- a/src/Temporalio/Bridge/CustomSlotSupplier.cs
+++ b/src/Temporalio/Bridge/CustomSlotSupplier.cs
@@ -4,6 +4,7 @@ using System.Runtime.InteropServices;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Logging;
+using Temporalio.Worker.Tuning;
 
 namespace Temporalio.Bridge
 {
@@ -15,7 +16,7 @@ namespace Temporalio.Bridge
         private readonly ILogger logger;
         private readonly Temporalio.Worker.Tuning.CustomSlotSupplier userSupplier;
         private readonly ConcurrentDictionary<IntPtr, CancellationTokenSource> reservationCancelSources = new();
-        private readonly ConcurrentDictionary<IntPtr, Temporalio.Worker.Tuning.SlotPermit> permits = new();
+        private readonly ConcurrentDictionary<IntPtr, SlotPermit> permits = new();
 
         /// <summary>
         /// Initializes a new instance of the <see cref="CustomSlotSupplier" /> class.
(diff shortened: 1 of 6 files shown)


src/Temporalio/Bridge/CustomSlotSupplier.cs
```diff
@@ -14,8 +14,10 @@ internal class CustomSlotSupplier : NativeInvokeableClass<Interop.TemporalCoreCu
     {
         private readonly ILogger logger;
         private readonly Temporalio.Worker.Tuning.CustomSlotSupplier userSupplier;
+        private readonly Dictionary<IntPtr, CancellationTokenSource> pendingReservationTokenSources = new();
         private readonly Dictionary<uint, Temporalio.Worker.Tuning.SlotPermit> permits = new();
-        private uint permitId = 1;
+        private readonly object mutex = new();
```
cretz: Can you clarify in a comment here which fields are expected to be only used in this mutex? It was a bit unclear at first read

src/Temporalio/Bridge/CustomSlotSupplier.cs
```diff
@@ -59,76 +62,141 @@ private static Temporalio.Worker.Tuning.SlotInfo SlotInfoFromBridge(Interop.Temp
...
+            var cancelTokenSrc = new CancellationTokenSource();
+            lock (mutex)
+            {
+                pendingReservationTokenSources.Add(completionCtx, cancelTokenSrc);
+            }
+            var x = Task.Run(async () =>
+            {
+                while (true)
+                {
+                    if (cancelTokenSrc.Token.IsCancellationRequested)
```
cretz: Is there value in doing all of this here instead of the catch clause below? Arguably not much value in preventing a call on an already-canceled context, but meh, maybe there is since we sleep.

src/Temporalio/Bridge/CustomSlotSupplier.cs
```diff
@@ -59,76 +62,141 @@ private static Temporalio.Worker.Tuning.SlotInfo SlotInfoFromBridge(Interop.Temp
...
                     {
-                        var srcHandle = GCHandle.Alloc(cancelTokenSrc);
-                        Interop.Methods.temporal_core_set_reserve_cancel_target(
-                            (Interop.TemporalCoreSlotReserveCtx*)ctx.ToPointer(),
-                            GCHandle.ToIntPtr(srcHandle).ToPointer());
+                        lock (mutex)
+                        {
+                            pendingReservationTokenSources.Remove(completionCtx);
```
cretz: IMO this dictionary remove can be done in a `finally` for this whole task (I know it adds another 4-space-indent to the whole Task.Run lambda, but it's worth it IMO for clarity, though could technically do in a `ContinueWith`). Also, arguably you don't need a mutex here, can use a `ConcurrentDictionary` instead.

src/Temporalio/Bridge/CustomSlotSupplier.cs
```diff
@@ -59,76 +62,141 @@ private static Temporalio.Worker.Tuning.SlotInfo SlotInfoFromBridge(Interop.Temp
...
+                                permits.Remove(usedPermitId);
                             }
-                            return;
+                            CompleteCancelReserve(completionCtx);
+                        }
+
+                        return;
+                    }
+                    else
+                    {
```
cretz: Pedantic, but for clarity reasons, would move this to its own `if (permit == null)` above the current `if` and add a `continue` inside of it, so the rest of this permit-completion logic can be done outside of an if.

src/Temporalio/Bridge/CustomSlotSupplier.cs
```diff
@@ -190,56 +261,66 @@ private unsafe void Release(Interop.TemporalCoreSlotReleaseCtx* ctx)
...
-        private unsafe Temporalio.Worker.Tuning.SlotReleaseContext ReleaseCtxFromBridge(
-            Interop.TemporalCoreSlotReleaseCtx* ctx,
-            Temporalio.Worker.Tuning.SlotPermit permit)
+        // this method must be called under a lock
+        private uint UnsynchronizedAddPermitToMap(Temporalio.Worker.Tuning.SlotPermit permit)
         {
-            return new(
-                SlotInfo: (*ctx).slot_info is null ? null : SlotInfoFromBridge(*(*ctx).slot_info),
-                Permit: permit);
+            while (true)
```
cretz: This is a bit confusing. Firstly, the only clash could happen if you had a permit a second for 136 years or something. And we don't need a separate ID when you have an object ID anyways. And we don't need mutexes when you have concurrent dictionaries. This problem calls for a "concurrent identity hash set".

I recommend using a `ConcurrentDictionary<Temporalio.Worker.Tuning.SlotPermit, byte>` for permits, with it instantiated as:

```csharp
new ConcurrentDictionary<Temporalio.Worker.Tuning.SlotPermit, byte>(
    EqualityComparer<Temporalio.Worker.Tuning.SlotPermit>.Create(
        (x, y) => ReferenceEquals(x, y),
        obj => RuntimeHelpers.GetHashCode(obj)
    )
)
```

This is basically the best .NET can do for a concurrent identity hash set. You use concurrent dictionary key things to affect the map safely (the value does not matter).

(sorry, I should have suggested this when #372 was first created)

EDIT: Actually, `EqualityComparer.Create` does not exist in versions of .NET we need, you need to create a separate inner class at the very bottom here that looks like:

```csharp
private class PermitIdentityComparer : IEqualityComparer<Temporalio.Worker.Tuning.SlotPermit>
{
    public static PermitIdentityComparer Instance { get; } = new();

    public bool Equals(
        Temporalio.Worker.Tuning.SlotPermit x, Temporalio.Worker.Tuning.SlotPermit y) =>
            ReferenceEquals(x, y);

    public int GetHashCode(Temporalio.Worker.Tuning.SlotPermit obj) =>
        RuntimeHelpers.GetHashCode(obj);
}
```
And pass `PermitIdentityComparer.Instance` to the constructor of the concurrent dictionary. It's worth it over these mutexes and loops.

src/Temporalio/Worker/Tuning/CustomSlotSupplier.cs
```diff
@@ -12,6 +12,18 @@ namespace Temporalio.Worker.Tuning
...
+        /// Gets the number of available slots at the moment of the call if known. Null value indicates unknown number
+        /// of available slots. This call should be non-blocking. The default implementation always returns null.
+        /// </summary>
+        /// <remarks>
+        /// This property will be accessed concurrently from multiple threads, so it must be thread-safe.
+        /// </remarks>
+        /// <remarks>
+        /// Any exceptions thrown will be logged and ignored.
+        /// </remarks>
+        public virtual uint? AvailableSlots => null;
```
cretz: Not sure we're adding this to our other Core-based custom slot suppliers, not sure we should here either

maciejdudko: Removed for now, we can revisit it another time.

src/Temporalio/Bridge/CustomSlotSupplier.cs
```diff
@@ -170,76 +209,56 @@ private unsafe void MarkUsed(Interop.TemporalCoreSlotMarkUsedCtx* ctx)
...
-            lock (permits)
-            {
-                var usedPermitId = permitId;
-                permits.Add(permitId, permit);
-                permitId += 1;
-                return usedPermitId;
-            }
+            // We use an address of a newly created pinned object as a unique permit ID.
+            // We cannot pin the permit itself because the type of permit may be unpinnable.
+            var permitId = GCHandle.ToIntPtr(GCHandle.Alloc(new byte[1], GCHandleType.Pinned));
```
cretz: I was thinking the dictionary's key would be an `object` that you would derive from a pointer when needed. But if you just want a key as an `IntPtr`, can probably be cheaper/clearer with `Marshal.AllocHGlobal(1)` and a `Marshal.FreeHGlobal` later. Don't have to worry about GC or making an array or whatever.

src/Temporalio/Bridge/CustomSlotSupplier.cs
```diff
@@ -59,108 +61,145 @@ private static Temporalio.Worker.Tuning.SlotInfo SlotInfoFromBridge(Interop.Temp
...
-        // Note that this is always called by Rust, either because the call is cancelled or because
-        // it completed. Therefore the GCHandle is always freed.
-        private unsafe void CancelReserve(void* tokenSrc)
+        private unsafe void Reserve(Interop.TemporalCoreSlotReserveCtx* ctx, Interop.TemporalCoreSlotReserveCompletionCtx* completionCtx, void* userData)
         {
-            var handle = GCHandle.FromIntPtr(new IntPtr(tokenSrc));
-            var cancelTokenSrc = (CancellationTokenSource)handle.Target!;
-            cancelTokenSrc.Cancel();
-            handle.Free();
+            SafeReserve(ReserveCtxFromBridge(ctx), new IntPtr(completionCtx));
```
cretz, line 83: I assume we're not having `SafeReserve` accept `Interop.TemporalCoreSlotReserveCompletionCtx*` directly is because you don't want to mark the whole method as `unsafe`? I see where we're just unsafely casting it back anyways. Just want to confirm.

src/Temporalio/Bridge/CustomSlotSupplier.cs
```diff
@@ -59,108 +61,145 @@ private static Temporalio.Worker.Tuning.SlotInfo SlotInfoFromBridge(Interop.Temp
...
-                    }
                     while (true)
                     {
                         try
                         {
-                            Task<Temporalio.Worker.Tuning.SlotPermit> reserveTask;
+                            cancelTokenSrc.Token.ThrowIfCancellationRequested();
+                            var permit = await userSupplier.ReserveSlotAsync(ctx, cancelTokenSrc.Token)
+                                .ConfigureAwait(false);
+                            var permitId = StorePermit(permit);
```
cretz: It was a bit confusing to me at first that even though the below code is in this same try/catch block, we don't expect an exception after here (because if it could happen, you would leak permits never releasing them). I wonder if the try/catch should be around just the reserve slot call and not other areas just for reader clarity. At the least, a comment above this line that says "we do not expect exceptions after this so that's why we don't consider it a leak that we don't release this on exception" may help.

src/Temporalio/Bridge/CustomSlotSupplier.cs
```diff
@@ -59,108 +61,145 @@ private static Temporalio.Worker.Tuning.SlotInfo SlotInfoFromBridge(Interop.Temp
...
                 }
+#pragma warning disable CA1031 // The task is detached, logging here is the only way to observe exceptions.
+                catch (Exception e)
+                {
+#pragma warning restore CA1031
+                    logger.LogError(e, "Exception escaped reserve slot retry loop");
+                }
+                finally
+                {
+                    reservationCancelSources.TryRemove(completionCtx, out _);
```
cretz, line 190:
```suggestion
                    reservationCancelSources.Remove(completionCtx);
```

I think this is good enough IIUC

cretz (commented):
Mostly LGTM, just a couple of minor comments

src/Temporalio/Bridge/CustomSlotSupplier.cs
maciejdudko: Yes, that's the reason.

src/Temporalio/Bridge/CustomSlotSupplier.cs
maciejdudko: `ConcurrentDictionary` has explicit implementation of `IDictionary`, calling `Remove` would require an ugly type cast.

cretz (approved):

PR #535, opened 2025-10-13
Nexus release seemed to forget to AddNexusService in extensions

mnichols:
## What was changed
Support AddNexusService same as other worker `Add` calls for convenience.
## Why?
Because digging into `ConfigureOptions` isn't as obvious and we already expose AddX calls in other places. This makes things consistent.

3. Any docs updates needed?
The Extensions project README.md was updated to use `AddNexusService`. 

## Concerns

Ideally, we could use the service provider to register these services instead of forcing ctor while registering...something like:
```
builder.AddNexusService<IHasANexii>()
``` 

I couldn't see a clear way to do this without the terrible pattern:
```
var serviceProvider = services.BuildServiceProvider();
var nexusService = serviceProvider.GetService<IHasANexii>();
```

Any other ideas we can try to make this more container friendly?

### DIFF
--- a/src/Temporalio.Extensions.Hosting/TemporalWorkerServiceOptionsBuilderExtensions.cs
+++ b/src/Temporalio.Extensions.Hosting/TemporalWorkerServiceOptionsBuilderExtensions.cs
@@ -165,6 +165,16 @@ namespace Temporalio.Extensions.Hosting
             builder.ConfigureOptions(options => options.AddWorkflow(type));
 
         /// <summary>
+        /// Add the given Nexus service on this worker service.
+        /// </summary>
+        /// <param name="builder">Builder to use.</param>
+        /// <param name="serviceHandler">Service handler.</param>
+        /// <returns>Same builder instance.</returns>
+        public static ITemporalWorkerServiceOptionsBuilder AddNexusService(
+            this ITemporalWorkerServiceOptionsBuilder builder, object serviceHandler) =>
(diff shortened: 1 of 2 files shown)


src/Temporalio.Extensions.Hosting/TemporalWorkerServiceOptionsBuilderExtensions.cs
```diff
@@ -164,6 +164,16 @@ public static ITemporalWorkerServiceOptionsBuilder AddWorkflow(
...
 
+        /// <summary>
+        /// Add the given Nexus service on this worker service.
+        /// </summary>
+        /// <param name="builder">Builder to use.</param>
+        /// <param name="serviceHandler">Service handler.</param>
+        /// <returns>Same builder instance.</returns>
+        public static ITemporalWorkerServiceOptionsBuilder AddNexusService(
+            this ITemporalWorkerServiceOptionsBuilder builder, object serviceHandler) =>
+            builder.ConfigureOptions(options => options.AddNexusService(serviceHandler));
```
cretz, line 175: I am afraid to do this right, we'll need to support the different transient/scoped/static lifetimes and accept types and such so DI will work. I have opened #536 (not sure why I didn't have it before). Can tackle here or give us a bit to get around to it (Nexus .NET is missing a few things in its pre-release state that we are tracking).

cretz:
Fixed in #561

PR #546, opened 2025-10-23
Manual pagination of ListWorkflows operation

### LINKED ISSUE #221: [Feature Request] Allow manual pagination of workflows

Today we use async iterators that page internally, but we should surface paging to users (i.e. page size and next page token). Then the iterators can just leverage that. The outbound interceptor for list is `ListWorkflowsAsync`, so we may need another single-page interceptor or we need to tailor that.

maciejdudko:
## What was changed

- Added `GetListWorkflowsPageAsync` to `ITemporalClient` to allow listing workflows with manual paging.
- Added `FetchListWorkflowsPageAsync` to `ClientOutboundInterceptor` to be called by both the new `GetListWorkflowsPageAsync` and (indirectly) the existing `ListWorkflowsAsync`.
- Rewrote `TemporalClient.Impl.ListWorkflowsInternalAsync` to make use of `GetListWorkflowsPageAsync`.

Additionally, made `TemporalClient.Impl.FetchWorkflowHistoryEventPageAsync` return frozen collections for extra safety.

## Why?

Feature request: #221

## Checklist

1. Closes #221

2. How was this tested:

Added `TemporalClientWorkflowTests.ListWorkflowsAsync_ManualPaging_IsAccurate`


src/Temporalio/Client/ITemporalClient.Workflow.cs
```diff
@@ -183,5 +183,24 @@ public IAsyncEnumerable<WorkflowExecution> ListWorkflowsAsync(
...
+        /// Set to null for the initial call to retrieve the first page.
+        /// Set to <see cref="ListWorkflowsPage.NextPageToken"/> returned by a previous call to retrieve the next page.
+        /// </param>
+        /// <param name="options">Options for the list call.</param>
+        /// <returns>
+        /// A single page of a list of workflows.
+        /// Repeat the call using <see cref="ListWorkflowsPage.NextPageToken"/> to get more pages.
+        /// </returns>
+        /// <seealso href="https://docs.temporal.io/visibility">Visibility docs.</seealso>
+        Task<ListWorkflowsPage> GetListWorkflowsPageAsync(
```
cretz:
```suggestion
        Task<WorkflowListPage> FetchWorkflowListPageAsync(
```

would recommend this for a few reasons, namely because a "workflow list page" makes more sense than "list workflows page" IMO, and we don't really want to use getters for remote RPC calls. Could also just add `Paginated` to the existing call like so:

```suggestion
        Task<WorkflowListPage> ListWorkflowsPaginatedAsync(
```

That is also fairly clean I think

src/Temporalio/Client/ITemporalClient.Workflow.cs
```diff
@@ -183,5 +183,24 @@ public IAsyncEnumerable<WorkflowExecution> ListWorkflowsAsync(
...
+            string query, byte[]? nextPageToken, GetListWorkflowsPageOptions? options = null);
```
cretz: Arguably you could require options and put next page token inside of there. But this is fine too.

src/Temporalio/Client/ListWorkflowsPaginatedOptions.cs
```diff
@@ -0,0 +1,9 @@
+namespace Temporalio.Client
+{
+    /// <summary>
+    /// Options for <see cref="ITemporalClient.ListWorkflowsPaginatedAsync"/>.
+    /// </summary>
+    /// <param name="PageSize">Number of results per page. Zero means server default.</param>
+    /// <param name="Rpc">RPC options for listing workflows.</param>
+    public record ListWorkflowsPaginatedOptions(int PageSize = 0, RpcOptions? Rpc = null);
```
cretz: To match the existing `WorkflowListOptions`, can we change this class name to `WorkflowListPaginatedOptions`? We kinda have to switch adjective/verb/noun around on method vs option class name

cretz: All options classes need to not be records, have zero-arg constructor, and be cloneable. This helps with consistency and fits into some .NET options patterns that can support external configuration.

cretz (commented):
Only two minor things w/ the options class, then all good

cretz (approved):

PR #547, opened 2025-10-23
Fixed Temporalio.Bridge.Scope leaking cancellation tokens

### LINKED ISSUE #540: [Bug] Memory Leak in Temporalio.Bridge.CancellationToken

I am using the Temporal client to start workflows, and noticing that inside the process, there exists an ever growing amount of objects of type `Temporalio.Bridge.CancellationToken` and `CancellationTokenSource+CallbackNode` (which typically represents a callback to be invoked when the token is cancelled)

### Describe the bug

https://github.com/temporalio/sdk-dotnet/blob/main/src/Temporalio/Bridge/CancellationToken.cs#L40

`CancellationToken.Register` registers a callback on the token and returns a `CancellationTokenRegistration` that should be disposed of once the callback is no longer needed. It doesn't look like the `CancellationTokenRegistration` is getting disposed of

https://github.com/temporalio/sdk-dotnet/blob/main/src/Temporalio/Bridge/Scope.cs#L164

`Temporalio.Bridge.CancellationToken` is not evaluating to true in the conditional `if (v is GCHandle handle)` , and thus the `Temporalio.Bridge.CancellationToken.ReleaseHandle` method is never being invoked

I am not certain if there are any other types that should be getting disposed of that are not

Image

### Minimal Reproduction

In the root directory of this repository, create a new project like this

```
dotnet new console --output test/MemoryLeak  
```

Use the following csproj and Program.cs file

```
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net9.0</TargetFramework>
    <ImplicitUsings>enable</ImplicitUsings>
    <Nullable>enable</Nullable>
  </PropertyGroup>

  <ItemGroup>
    <ProjectReference Include="..\..\src\Temporalio\Temporalio.csproj" />
  </ItemGroup>

</Project>

```

```csharp
using Temporalio.Client;

using CancellationTokenSource cts = new();

TemporalClient client = await TemporalClient.ConnectAsync(new TemporalClientConnectOptions("localhost:7233")).ConfigureAwait(false);

await client.WorkflowService.GetSystemInfoAsync(new(), new RpcOptions() { CancellationToken = cts.Token }).ConfigureAwait(false);

```

Start a Temporal server: `temporal server start-dev`

Step through the program and validate that [ReleaseHandle](https://github.com/temporalio/sdk-dotnet/blob/main/src/Temporalio/Bridge/CancellationToken.cs#L56) is not called for the `GetSystemInfo` rpc

If you inspect the [registrations](https://github.com/dotnet/runtime/blob/1d1bf92fcf43aa6981804dc53c5174445069c9e4/src/libraries/System.Private.CoreLib/src/System/Threading/CancellationTokenSource.cs#L47) on the cancellation token source, you can see that there is still a callback attached as well

### Environment/Versions

- OS and processor: x64 Windows 11
- Temporal Version: 1.9.0
- Are you using Docker or Kubernetes or building Temporal from source? Running Temporal directly on host

### Additional context

I modified the SDK locally like below and validated that this seemed to address the memory leakage (although the implementation may be a bit naive)

```csharp
// src/Temporalio/Bridge/CancellationToken.cs
using System;
using System.Runtime.InteropServices;
using System.Threading;

namespace Temporalio.Bridge
{
    /// <summary>
    /// Core-owned cancellation token.
    /// </summary>
    internal class CancellationToken : SafeHandle
    {
        private CancellationTokenRegistration? cancellationCallback;

        /// <summary>
        /// Initializes a new instance of the <see cref="CancellationToken"/> class.
        /// </summary>
        public CancellationToken()
            : base(IntPtr.Zero, true)
        {
            unsafe
            {
                Ptr = Interop.Methods.temporal_core_cancellation_token_new();
                SetHandle((IntPtr)Ptr);
            }
        }

        /// <inheritdoc/>
        public override unsafe bool IsInvalid => Ptr == null;

        /// <summary>
        /// Gets internal token pointer.
        /// </summary>
        internal unsafe Interop.TemporalCoreCancellationToken* Ptr { get; private init; }

        /// <summary>
        /// Create a core cancellation token from the given cancellation token.
        /// </summary>
        /// <param name="token">Threading token.</param>
        /// <returns>Created cancellation token.</returns>
        public static CancellationToken FromThreading(System.Threading.CancellationToken token)
        {
            var ret = new CancellationToken();
            ret.cancellationCallback = token.Register(ret.Cancel);

            return ret;
        }

        /// <summary>
        /// Cancel this token.
        /// </summary>
        public void Cancel()
        {
            unsafe
            {
                Interop.Methods.temporal_core_cancellation_token_cancel(Ptr);
            }
        }

        /// <inheritdoc/>
        protected override unsafe bool ReleaseHandle()
        {
            // I suspect that the callback needs to be unregistered before freeing the core owned cancellation token to prevent any access issues
            cancellationCallback?.Dispose();
            Interop.Methods.temporal_core_cancellation_token_free(Ptr);
            return true;
        }
    }
}
```

```csharp
// src/Temporalio/Bridge/Scope.cs

/// <inheritdoc />
        public void Dispose()
        {
            foreach (var v in toKeepAlive)
            {
                if (v is GCHandle handle)
                {
                    handle.Free();
                }
                else if (v is CancellationToken cancellationToken)
                {
                    cancellationToken.Dispose();
                }
            }
            // This keep alive does nothing obviously, but it's good documentation to understand the
            // purpose of this separate dispose call
            GC.KeepAlive(toKeepAlive);
            GC.SuppressFinalize(this);
        }
```

maciejdudko:
## What was changed

- Implemented Dispose pattern in `Temporalio.Bridge.Scope` to correctly release both unmanaged (`GCHandle`) and managed (`CancellationToken`) resources.
- `Temporalio.Bridge.CancellationToken` now unregisters its callback from the associated `System.Threading.CancellationToken` when freeing the handle.

## Why?

Fixes memory leak due to undisposed core-owned cancellation tokens.

## Checklist

1. Closes #540 

2. How was this tested:

Automated testing is not possible for this bug. Manual testing was done following the "Minimal Reproduction" instructions in the linked issue.

### DIFF
--- a/src/Temporalio/Bridge/CancellationToken.cs
+++ b/src/Temporalio/Bridge/CancellationToken.cs
@@ -4,16 +4,17 @@ using System.Runtime.InteropServices;
 namespace Temporalio.Bridge
 {
     /// <summary>
-    /// Core-owned cancellation token.
+    /// Core-owned cancellation token that's bound to the given .NET cancellation token.
     /// </summary>
     internal class CancellationToken : SafeHandle
     {
-        private System.Threading.CancellationTokenRegistration? registration;
+        private readonly System.Threading.CancellationTokenRegistration registration;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="CancellationToken"/> class.
         /// </summary>
-        public CancellationToken()
+        /// <param name="token">.NET cancellation token to bind.</param>
+        public CancellationToken(System.Threading.CancellationToken token)
             : base(IntPtr.Zero, true)
         {
             unsafe
@@ -21,6 +22,7 @@ namespace Temporalio.Bridge
                 Ptr = Interop.Methods.temporal_core_cancellation_token_new();
                 SetHandle((IntPtr)Ptr);
             }
+            registration = token.Register(Cancel);
         }
 
         /// <inheritdoc/>
@@ -32,20 +34,12 @@ namespace Temporalio.Bridge
         internal unsafe Interop.TemporalCoreCancellationToken* Ptr { get; }
 
         /// <summary>
-        /// Create a core cancellation token from the given cancellation token.
+        /// Cancels the Core-owned token. Does not cancel the bound .NET token.
         /// </summary>
-        /// <param name="token">Threading token.</param>
-        /// <returns>Created cancellation token.</returns>
-        public static CancellationToken FromThreading(System.Threading.CancellationToken token)
-        {
-            var ret = new CancellationToken();
-            ret.registration = token.Register(ret.Cancel);
-            return ret;
-        }
-
-        /// <summary>
-        /// Cancel this token.
-        /// </summary>
+        /// <remarks>
+        /// Cancelling the bound .NET token automatically cancels the Core-owned token,
+        /// but not the other way around.
+        /// </remarks>
         public void Cancel()
         {
             if (!IsClosed)
@@ -60,7 +54,7 @@ namespace Temporalio.Bridge
         /// <inheritdoc/>
         protected override unsafe bool ReleaseHandle()
         {
-            registration?.Dispose();
+            registration.Dispose();
             Interop.Methods.temporal_core_cancellation_token_free(Ptr);
             return true;
         }
--- a/src/Temporalio/Bridge/CancellationToken.cs
+++ b/src/Temporalio/Bridge/CancellationToken.cs
@@ -8,6 +8,8 @@ namespace Temporalio.Bridge
     /// </summary>
     internal class CancellationToken : SafeHandle
     {
+        private System.Threading.CancellationTokenRegistration? registration;
+
         /// <summary>
         /// Initializes a new instance of the <see cref="CancellationToken"/> class.
         /// </summary>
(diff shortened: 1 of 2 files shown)


src/Temporalio/Bridge/CancellationToken.cs
```diff
@@ -8,6 +8,8 @@ namespace Temporalio.Bridge
     /// </summary>
     internal class CancellationToken : SafeHandle
     {
+        private System.Threading.CancellationTokenRegistration? registration;
```
cretz: Can this be readonly and set in the constructor? Also, do we only create this via `FromThreading`? If so, I wonder if the constructor can be made private and this not be nullable (or turn `FromThreading` into the public constructor).

cretz (commented):
Nothing really blocking, but wanted to discuss field nullabillity and such

src/Temporalio/Bridge/Scope.cs
```diff
@@ -152,24 +160,42 @@ public IntPtr FunctionPointer<T>(T func)
...
+        }
+
+        private void Dispose(bool disposing)
+        {
+            if (disposed)
+            {
+                return;
+            }
+
+            byteArrayRefs.Clear();
```
cretz, line 181: I don't think you need to clear these, but not a big deal

cretz (approved):

Issue #32, opened 2023-03-09
[Feature Request] Add CancellationToken parameter to TemporalClient.ConnectAsync and friends


cgillum:
### Is your feature request related to a problem? Please describe.
It's standard practice in .NET APIs that perform I/O to include a `CancellationToken` parameter so that callers can abort I/O requests that take too long.  `TemporalClient.ConnectAsync(...)` doesn't accept a cancellation token.

### Describe the solution you'd like
Add an optional `CancellationToken` to `TemporalClient.ConnectAsync(...)`, like so:

```csharp
public static async Task<TemporalClient> ConnectAsync(
    TemporalClientConnectOptions options,
    CancellationToken cancellationToken = default) { ... }
```

This feedback applies to other APIs as well, such as `TemporalConnection.ConnectAsync(...)`.

cretz:
On every async call that supports cancellation we do accept a cancellation token, though we intentionally deviate from .NET here and include it in the options instead of as trailing parameter to keep the overload count down. This includes every call on a connected client.

But we intentionally did not include a cancellation token on connect because we don't have that wired up with Rust. It's not as easy as just cancelling a .NET task, I have to weave the token through Tokio in Rust. I put in this effort for RPC calls but not yet the connect process itself. I will look into putting this time in before beta. In the meantime know that the connection process cannot be trivially cancelled with a token.

PR #548, opened 2025-10-24
Initial implementation of plugins

### LINKED ISSUE #507: [Feature Request] Plugin support

See https://github.com/temporalio/features/issues/652

tconley1428:
## What was changed
Added `ITemporalClientPlugin` and `ITemporalWorkerPlugin` to provide configurability to their respective objects. `SimplePlugin` was added to make it easier to implement plugins consistently across objects.

## Why?
It can be difficult to provide consistent and easy methods of configuring temporal objects.

## Checklist

1. Closes #507 #plu 

2. How was this tested:
New set of plugin tests.

3. Any docs updates needed?
Plugin documentation in general is pending.

### DIFF
--- a/src/Temporalio/Client/ITemporalClientPlugin.cs
+++ b/src/Temporalio/Client/ITemporalClientPlugin.cs
@@ -1,0 +1,14 @@
+using System;
+using System.Threading.Tasks;
+
+namespace Temporalio.Client
+{
+    public interface ITemporalClientPlugin
+    {
+        public string Name { get; }
+
+        public void ConfigureClient(TemporalClientOptions options);
+
+        public Task<TemporalConnection> TemporalConnectAsync(TemporalClientConnectOptions options, Func<TemporalClientConnectOptions, Task<TemporalConnection>> continuation);
+    }
+}
--- a/src/Temporalio/Client/ITemporalClientPlugin.cs
+++ b/src/Temporalio/Client/ITemporalClientPlugin.cs
@@ -3,12 +3,31 @@ using System.Threading.Tasks;
 
 namespace Temporalio.Client
 {
+    /// <summary>
+    /// Interface for temporal client plugins.
+    /// </summary>
+    /// <remarks>
+    /// WARNING: This API is experimental and may change in the future.
+    /// </remarks>
     public interface ITemporalClientPlugin
     {
+        /// <summary>
+        /// Gets the plugin name.
+        /// </summary>
         public string Name { get; }
 
+        /// <summary>
+        /// Configures the client options.
+        /// </summary>
+        /// <param name="options">The client options to configure.</param>
         public void ConfigureClient(TemporalClientOptions options);
 
+        /// <summary>
+        /// Handles temporal connection asynchronously.
+        /// </summary>
+        /// <param name="options">The connection options.</param>
+        /// <param name="continuation">The continuation function.</param>
+        /// <returns>A task representing the asynchronous operation.</returns>
         public Task<TemporalConnection> TemporalConnectAsync(TemporalClientConnectOptions options, Func<TemporalClientConnectOptions, Task<TemporalConnection>> continuation);
     }
 }
--- a/src/Temporalio/Client/ITemporalClientPlugin.cs
+++ b/src/Temporalio/Client/ITemporalClientPlugin.cs
@@ -14,13 +14,13 @@ namespace Temporalio.Client
         /// <summary>
         /// Gets the plugin name.
         /// </summary>
-        public string Name { get; }
+        string Name { get; }
 
         /// <summary>
         /// Configures the client options.
         /// </summary>
         /// <param name="options">The client options to configure.</param>
-        public void ConfigureClient(TemporalClientOptions options);
+        void ConfigureClient(TemporalClientOptions options);
 
         /// <summary>
         /// Handles temporal connection asynchronously.
@@ -28,6 +28,6 @@ namespace Temporalio.Client
         /// <param name="options">The connection options.</param>
         /// <param name="continuation">The continuation function.</param>
         /// <returns>A task representing the asynchronous operation.</returns>
-        public Task<TemporalConnection> TemporalConnectAsync(TemporalClientConnectOptions options, Func<TemporalClientConnectOptions, Task<TemporalConnection>> continuation);
+        Task<TemporalConnection> ConnectAsync(TemporalClientConnectOptions options, Func<TemporalClientConnectOptions, Task<TemporalConnection>> continuation);
     }
 }
--- a/src/Temporalio/Client/TemporalClient.cs
+++ b/src/Temporalio/Client/TemporalClient.cs
@@ -21,6 +21,11 @@ namespace Temporalio.Client
         /// <param name="options">Options for this client.</param>
         public TemporalClient(ITemporalConnection connection, TemporalClientOptions options)
         {
+            if (options.Plugins != null)
+            {
+                options = options.Plugins.Aggregate(
+                    options, (clientOptions, plugin) => plugin.ConfigureClient(clientOptions));
+            }
             Connection = connection;
             Options = options;
             OutboundInterceptor = new Impl(this);
@@ -61,10 +66,21 @@ namespace Temporalio.Client
         /// <param name="options">Options for connecting.</param>
         /// <returns>The connected client.</returns>
         public static async Task<TemporalClient> ConnectAsync(
-            TemporalClientConnectOptions options) =>
-            new(
-                await TemporalConnection.ConnectAsync(options).ConfigureAwait(false),
+            TemporalClientConnectOptions options)
+        {
+            Func<TemporalClientConnectOptions, Task<TemporalConnection>> connect = TemporalConnection.ConnectAsync;
+            if (options.Plugins != null)
+            {
+                foreach (var plugin in options.Plugins.Reverse())
+                {
+                    var localConnect = connect;
+                    connect = connectOptions => plugin.TemporalConnectAsync(connectOptions, localConnect);
+                }
+            }
+            return new(
+                await connect(options).ConfigureAwait(false),
                 options.ToClientOptions());
+        }
 
         /// <summary>
         /// Create a client to a Temporal namespace that does not connect until first call.
--- a/src/Temporalio/Client/TemporalClient.cs
+++ b/src/Temporalio/Client/TemporalClient.cs
@@ -21,11 +21,11 @@ namespace Temporalio.Client
         /// <param name="options">Options for this client.</param>
         public TemporalClient(ITemporalConnection connection, TemporalClientOptions options)
         {
-            if (options.Plugins != null)
+            foreach (var plugin in options.Plugins ?? Enumerable.Empty<ITemporalClientPlugin>())
             {
-                options = options.Plugins.Aggregate(
-                    options, (clientOptions, plugin) => plugin.ConfigureClient(clientOptions));
+                plugin.ConfigureClient(options);
             }
+
             Connection = connection;
             Options = options;
             OutboundInterceptor = new Impl(this);
--- a/src/Temporalio/Client/TemporalClient.cs
+++ b/src/Temporalio/Client/TemporalClient.cs
@@ -21,9 +21,12 @@ namespace Temporalio.Client
         /// <param name="options">Options for this client.</param>
         public TemporalClient(ITemporalConnection connection, TemporalClientOptions options)
         {
-            foreach (var plugin in options.Plugins ?? Enumerable.Empty<ITemporalClientPlugin>())
+            if (options.Plugins != null)
             {
-                plugin.ConfigureClient(options);
+                foreach (var plugin in options.Plugins)
+                {
+                    plugin.ConfigureClient(options);
+                }
             }
 
             Connection = connection;
--- a/src/Temporalio/Client/TemporalClient.cs
+++ b/src/Temporalio/Client/TemporalClient.cs
@@ -74,7 +74,7 @@ namespace Temporalio.Client
                 foreach (var plugin in options.Plugins.Reverse())
                 {
                     var localConnect = connect;
-                    connect = connectOptions => plugin.TemporalConnectAsync(connectOptions, localConnect);
+                    connect = connectOptions => plugin.ConnectAsync(connectOptions, localConnect);
                 }
             }
             return new(
--- a/src/Temporalio/Client/TemporalClientConnectOptions.cs
+++ b/src/Temporalio/Client/TemporalClientConnectOptions.cs
@@ -87,6 +87,7 @@ namespace Temporalio.Client
                 Interceptors = Interceptors,
                 LoggerFactory = LoggerFactory,
                 QueryRejectCondition = QueryRejectCondition,
+                Plugins = Plugins,
             };
     }
 }
--- a/src/Temporalio/Client/TemporalClientConnectOptions.cs
+++ b/src/Temporalio/Client/TemporalClientConnectOptions.cs
@@ -69,6 +69,9 @@ namespace Temporalio.Client
         /// <summary>
         /// Gets or sets the plugins.
         /// </summary>
+        /// <remarks>
+        /// WARNING: This API is experimental and may change in the future.
+        /// </remarks>
         public IReadOnlyCollection<ITemporalClientPlugin>? Plugins { get; set; }
 
         /// <summary>
--- a/src/Temporalio/Client/TemporalClientConnectOptions.cs
+++ b/src/Temporalio/Client/TemporalClientConnectOptions.cs
@@ -67,6 +67,11 @@ namespace Temporalio.Client
         public QueryRejectCondition? QueryRejectCondition { get; set; }
 
         /// <summary>
+        /// Gets or sets the plugins.
+        /// </summary>
+        public IReadOnlyCollection<ITemporalClientPlugin>? Plugins { get; set; }
+
+        /// <summary>
         /// Create client options from a subset of these options for use in
         /// <see cref="TemporalClient.TemporalClient" />.
         /// </summary>
--- a/src/Temporalio/Common/SimplePlugin.cs
+++ b/src/Temporalio/Common/SimplePlugin.cs
@@ -2,18 +2,29 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
-using Temporalio.Activities;
 using Temporalio.Client;
 using Temporalio.Worker;
 
 namespace Temporalio.Common
 {
-    public class SimplePlugin: ITemporalClientPlugin, ITemporalWorkerPlugin
+    /// <summary>
+    /// A simple plugin that implements both client and worker plugin interfaces.
+    /// </summary>
+    /// <remarks>
+    /// WARNING: This API is experimental and may change in the future.
+    /// </remarks>
+    public class SimplePlugin : ITemporalClientPlugin, ITemporalWorkerPlugin
     {
+        /// <summary>
+        /// The plugin options.
+        /// </summary>
         private readonly SimplePluginOptions pluginOptions;
 
-        public string Name { get; }
-
+        /// <summary>
+        /// Initializes a new instance of the <see cref="SimplePlugin"/> class.
+        /// </summary>
+        /// <param name="name">The plugin name.</param>
+        /// <param name="options">The plugin options.</param>
         public SimplePlugin(
             string name,
             SimplePluginOptions options)
@@ -22,15 +33,39 @@ namespace Temporalio.Common
             pluginOptions = options;
         }
 
-        public void ConfigureClient(TemporalClientOptions options)
+        /// <summary>
+        /// Gets the plugin name.
+        /// </summary>
+        public string Name { get; }
+
+        /// <summary>
+        /// Configures the client options.
+        /// </summary>
+        /// <param name="options">The client options to configure.</param>
+        public virtual void ConfigureClient(TemporalClientOptions options)
         {
             options.DataConverter = Resolve(options.DataConverter, pluginOptions.DataConverterOption);
             options.Interceptors = ResolveAppend(options.Interceptors, pluginOptions.ClientInterceptorsOption);
         }
 
-        public Task<TemporalConnection> TemporalConnectAsync(TemporalClientConnectOptions options, Func<TemporalClientConnectOptions, Task<TemporalConnection>> continuation) => throw new NotImplementedException();
+        /// <summary>
+        /// Handles temporal connection asynchronously.
+        /// </summary>
+        /// <param name="options">The connection options.</param>
+        /// <param name="continuation">The continuation function.</param>
+        /// <returns>A task representing the asynchronous operation.</returns>
+        public virtual Task<TemporalConnection> TemporalConnectAsync(
+            TemporalClientConnectOptions options,
+            Func<TemporalClientConnectOptions, Task<TemporalConnection>> continuation)
+        {
+            return continuation(options);
+        }
 
-        public void ConfigureWorker(TemporalWorkerOptions options)
+        /// <summary>
+        /// Configures the worker options.
+        /// </summary>
+        /// <param name="options">The worker options to configure.</param>
+        public virtual void ConfigureWorker(TemporalWorkerOptions options)
         {
             DoAppend(options.Activities, pluginOptions.Activities);
             DoAppend(options.Workflows, pluginOptions.Workflows);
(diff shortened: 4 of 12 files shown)


src/Temporalio/Client/ITemporalClientPlugin.cs
```diff
@@ -0,0 +1,33 @@
...
+        /// <param name="options">The client options to configure.</param>
+        public void ConfigureClient(TemporalClientOptions options);
+
+        /// <summary>
+        /// Handles temporal connection asynchronously.
+        /// </summary>
+        /// <param name="options">The connection options.</param>
+        /// <param name="continuation">The continuation function.</param>
+        /// <returns>A task representing the asynchronous operation.</returns>
+        public Task<TemporalConnection> TemporalConnectAsync(TemporalClientConnectOptions options, Func<TemporalClientConnectOptions, Task<TemporalConnection>> continuation);
```
cretz:
```suggestion
        public Task<TemporalConnection> ConnectClientAsync(TemporalClientConnectOptions options, Func<TemporalClientConnectOptions, Task<TemporalConnection>> continuation);
```

Pedantic, but I don't think the `Temporal` qualifier is necessary on the method name in this case (anymore than it is for the configure method)

src/Temporalio/Client/ITemporalClientPlugin.cs
```diff
@@ -0,0 +1,33 @@
...
+    /// </summary>
+    /// <remarks>
+    /// WARNING: This API is experimental and may change in the future.
+    /// </remarks>
+    public interface ITemporalClientPlugin
+    {
+        /// <summary>
+        /// Gets the plugin name.
+        /// </summary>
+        public string Name { get; }
```
cretz:
```suggestion
        string Name { get; }
```

`public` is implied for all interface things

src/Temporalio/Client/ITemporalClientPlugin.cs
```diff
@@ -0,0 +1,33 @@
...
+
+        /// <summary>
+        /// Configures the client options.
+        /// </summary>
+        /// <param name="options">The client options to configure.</param>
+        public void ConfigureClient(TemporalClientOptions options);
+
+        /// <summary>
+        /// Handles temporal connection asynchronously.
```
cretz, line 26: Pedantic - Was a bit confusing from outside to understand that this call is applied in reverse on plugins, but configure call was not in reverse. May be worth adding a note in both methods' documentation.

src/Temporalio/Client/TemporalClientConnectOptions.cs
```diff
@@ -66,6 +66,14 @@ public TemporalClientConnectOptions(string targetHost)
         /// </summary>
         public QueryRejectCondition? QueryRejectCondition { get; set; }
 
+        /// <summary>
+        /// Gets or sets the plugins.
+        /// </summary>
+        /// <remarks>
+        /// WARNING: This API is experimental and may change in the future.
+        /// </remarks>
+        public IReadOnlyCollection<ITemporalClientPlugin>? Plugins { get; set; }
```
cretz, line 75: Should also update `ToClientOptions` helper at the bottom of this file to include this

src/Temporalio/Client/ITemporalClientPlugin.cs
cretz: I am a bit concerned for an overload future if/when we implement #32. I wonder if that affects any API decisions here.

src/Temporalio/Worker/ITemporalWorkerPlugin.cs
```diff
@@ -0,0 +1,48 @@
...
+        /// <param name="options">The worker options to configure.</param>
+        public void ConfigureWorker(TemporalWorkerOptions options);
+
+        /// <summary>
+        /// Runs the worker asynchronously.
+        /// </summary>
+        /// <param name="worker">The worker to run.</param>
+        /// <param name="continuation">The continuation function.</param>
+        /// <returns>A task representing the asynchronous operation.</returns>
+        public Task RunWorkerAsync(TemporalWorker worker, Func<TemporalWorker, Task> continuation);
```
cretz:
```suggestion
        public Task<TResult> RunWorkerAsync<TResult>(TemporalWorker worker, Func<TemporalWorker, Task<TResult>> continuation, CancellationToken stoppingToken);
```

Arguably this is the full superset of the execute async call, even if our internals don't model it that way (we'd have to change the internal helper a bit maybe)

src/Temporalio/Worker/ITemporalWorkerPlugin.cs
```diff
@@ -0,0 +1,48 @@
...
+        /// <param name="options">The replayer options to configure.</param>
+        public void ConfigureReplayer(WorkflowReplayerOptions options);
+
+        /// <summary>
+        /// Runs the replayer asynchronously.
+        /// </summary>
+        /// <param name="replayer">The replayer to run.</param>
+        /// <param name="continuation">The continuation function.</param>
+        /// <returns>A task representing the asynchronous operation.</returns>
+        public Task<IEnumerable<WorkflowReplayResult>> RunReplayerAsync(WorkflowReplayer replayer, Func<WorkflowReplayer, Task<IEnumerable<WorkflowReplayResult>>> continuation);
```
cretz: Should I be allowed to access what is being sent to the replayer (e.g. the histories)?

src/Temporalio/Worker/WorkflowReplayer.cs
```diff
@@ -20,13 +20,21 @@ namespace Temporalio.Worker
...
+        private readonly IReadOnlyCollection<ITemporalWorkerPlugin> plugins;
+
         /// <summary>
         /// Initializes a new instance of the <see cref="WorkflowReplayer"/> class. Note, some
         /// options validation is deferred until replay is attempted.
         /// </summary>
         /// <param name="options">Replayer options.</param>
         public WorkflowReplayer(WorkflowReplayerOptions options)
         {
+            plugins = options.Plugins ?? new List<ITemporalWorkerPlugin>();
```
cretz:
```suggestion
            plugins = options.Plugins ?? Array.Empty<ITemporalWorkerPlugin>();
```

Pedantic, but a bit better option here IIUC

src/Temporalio/Worker/WorkflowReplayer.cs
```diff
@@ -133,6 +127,33 @@ await Task.WhenAny(
         }
 #endif
 
+        private async Task<IEnumerable<WorkflowReplayResult>> ReplayWorkflowsInternalAsync(
```
cretz, line 162: I am worried that the overload that accepts/returns `IAsyncEnumerable` doesn't go through plugins. Not sure if it's worth modelling both in the interface separately or somehow combined. Thoughts?

src/Temporalio/Common/SimplePlugin.cs
```diff
@@ -0,0 +1,167 @@
...
+    /// </summary>
+    /// <remarks>
+    /// WARNING: This API is experimental and may change in the future.
+    /// </remarks>
+    public class SimplePlugin : ITemporalClientPlugin, ITemporalWorkerPlugin
+    {
+        /// <summary>
+        /// The plugin options.
+        /// </summary>
+        private readonly SimplePluginOptions pluginOptions;
```
cretz: On most of these types of classes (e.g. client, worker, replayer, etc) we like to make `Options` a public init-only property they can access.

src/Temporalio/Common/SimplePluginOptions.cs
```diff
@@ -0,0 +1,275 @@
...
+        {
+            NexusServices.Add(ServiceHandlerInstance.FromInstance(serviceHandler));
+            return this;
+        }
+
+        /// <summary>
+        /// Represents a configurable plugin option.
+        /// </summary>
+        /// <typeparam name="T">The type of the option value.</typeparam>
+        public class SimplePluginOption<T>
```
cretz: In some languages it wouldn't make sense, but for me, I'd be ok if you dropped the `SimplePlugin` prefix to this class name since referencing nested classes in .NET from outside the class almost always use the qualified form with the parent class prefixed.

Though `SimplePluginOptions.Option` does sound weird (as does `SimplePluginOptions.SimplePluginOption`), feel free to use a different noun if you want, but I'm ok w/ just that.

src/Temporalio/Common/SimplePluginOptions.cs
```diff
@@ -0,0 +1,275 @@
...
+        {
+            /// <summary>
+            /// Initializes a new instance of the <see cref="SimplePluginOption{T}"/> class with a constant value.
+            /// </summary>
+            /// <param name="value">The constant value.</param>
+            public SimplePluginOption(T? value)
```
cretz: Pedantic, but a log of these single-line methods in this PR can use `=>` syntax, but by no means _have_ to

src/Temporalio/Common/SimplePluginOptions.cs
cretz: Can you help me understand the `SimplePluginOption` vs `SimplePluginRequiredOption` difference? Can they they be a single type and `Type?` not be the data type for non-required options? Also, am I not allowed to return a null from my plugin's configuration lambda for non-required options? E.g. maybe I only want to set a value if an env var is set.

src/Temporalio/Common/SimplePlugin.cs
```diff
@@ -0,0 +1,167 @@
...
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="SimplePlugin"/> class.
+        /// </summary>
+        /// <param name="name">The plugin name.</param>
+        /// <param name="options">The plugin options.</param>
+        public SimplePlugin(
+            string name,
+            SimplePluginOptions options)
```
cretz: I think we should consider making this parameter optional. People may want to use `SimplePlugin` as a "default base class" and still do everything via inheritance.

src/Temporalio/Common/SimplePluginOptions.cs
```diff
@@ -0,0 +1,275 @@
...
+
+namespace Temporalio.Common
+{
+    /// <summary>
+    /// Configuration options for simple plugins.
+    /// </summary>
+    /// <remarks>
+    /// WARNING: This API is experimental and may change in the future.
+    /// </remarks>
+    public class SimplePluginOptions
```
cretz: Can we make this implement `ICloneable` like other options? I don't expect `Clone` to do anything besides invoke/return `MemberwiseClone()`.

src/Temporalio/Common/SimplePluginOptions.cs
```diff
@@ -0,0 +1,275 @@
...
+        /// Gets or sets the workflow failure exception types option with configuration support.
+        /// </summary>
+        public SimplePluginOption<IReadOnlyCollection<Type>>? WorkflowFailureExceptionTypesOption
+        {
+            get;
+            set;
+        }
+
+        /// <summary>
+        /// Gets or sets the run context function for the plugin.
```
cretz: Pedantic, but may want to clarify this is for surrounding the running workers and replayers

src/Temporalio/Common/SimplePlugin.cs
```diff
@@ -0,0 +1,167 @@
...
+            options.WorkflowFailureExceptionTypes = ResolveAppend(options.WorkflowFailureExceptionTypes, pluginOptions.WorkflowFailureExceptionTypesOption);
+        }
+
+        /// <summary>
+        /// Runs the worker asynchronously.
+        /// </summary>
+        /// <param name="worker">The worker to run.</param>
+        /// <param name="continuation">The continuation function.</param>
+        /// <returns>A task representing the asynchronous operation.</returns>
+        public virtual Task RunWorkerAsync(TemporalWorker worker, Func<TemporalWorker, Task> continuation)
```
cretz: This and `RunReplayerAsync` don't seem to use `RunContext`

src/Temporalio/Worker/ITemporalWorkerPlugin.cs
cretz: I kinda like matching the replayer method name of `ReplayWorkflowsAsync`, but not a strong opinion

src/Temporalio/Client/ITemporalClientPlugin.cs
tconley1428: Sure

src/Temporalio/Client/ITemporalClientPlugin.cs
tconley1428: It isn't applied in reverse. The function is built in reverse, which I think isn't necessarily something the user needs to care about.

src/Temporalio/Common/SimplePlugin.cs
tconley1428: I haven't found a way to actually make `RunContext` work. For now we could either remove it an make it subclass only or find some typing workaround.

src/Temporalio/Common/SimplePluginOptions.cs
tconley1428: The original reason was because in many options I could make the lambda guarantee its input won't be null. I don't think I tried making it Option<T?> though, if that collapses T?? it may work.

src/Temporalio/Worker/ITemporalWorkerPlugin.cs
tconley1428: Mmm that might help with making run context work, but I'll have to check. From what I saw it is hard to coalesce `Task` and `Task<T>`

src/Temporalio/Worker/WorkflowReplayer.cs
tconley1428: Ah, I missed that one. I'll see what is possible.

src/Temporalio/Client/ITemporalClientPlugin.cs
tconley1428: It certainly would affect it, but I'm not sure what we would do here to mitigate that.

src/Temporalio/Client/ITemporalClientPlugin.cs
cretz: I guess what I mean is if I have plugin `Foo` and `Bar`, `ConfigureClient` is called on `Foo` _then_ `Bar` and `ConnectClientAsync` is called on `Bar` _then_ `Foo` which appears a bit inconsistent/surprising. I think that kind of order of operations may matter to plugin authors at least enough to document.

src/Temporalio/Worker/ITemporalWorkerPlugin.cs
cretz: It is indeed hard and something we have had to do in other places. What we've usually done is made the `Task<T>` the main impl and just use `Task<ValueTuple>` for `Task` equivalents, but it could add an extra layer of stack unnecessarily.

src/Temporalio/Worker/ITemporalWorkerPlugin.cs
tconley1428: You certainly could, but I'm not sure how necessary it is.

src/Temporalio/Client/ITemporalClientPlugin.cs
tconley1428: Right. I don't think that's the case. 
```
foreach (var plugin in options.Plugins.Reverse())
{
    var localConnect = connect;
    connect = connectOptions => plugin.ConnectAsync(connectOptions, localConnect);
}
```
It iterates in reverse to compose the functions, but the execution would then go back through in the forward order, just like interceptors.

cretz: Ah, my mistake, ignore me then

src/Temporalio/Client/ITemporalClientPlugin.cs
cretz: What we chose to do with interceptors was to make every method accept an single object we can expand. But with overloads, in .NET we can probably just make an overload with a default implementation of calling the older one.

cretz (approved):
All looks good to me, but I'm a bit biased because I was involved in the coding. @maciejdudko - can you take a look at this review?

src/Temporalio/Client/TemporalClient.cs
```diff
@@ -21,6 +21,11 @@ public partial class TemporalClient : ITemporalClient
         /// <param name="options">Options for this client.</param>
         public TemporalClient(ITemporalConnection connection, TemporalClientOptions options)
         {
+            foreach (var plugin in options.Plugins ?? Enumerable.Empty<ITemporalClientPlugin>())
```
jmaeagle99: nit: I don't know if this constructor is on a hot path or not. It would be theoretically slightly faster to null-check and iterate vs. constructing an empty enumerable and have the implicit cast down to `IEnumerable<ITemporalClientPlugin>` for the loop iteration.

src/Temporalio/Client/ITemporalClientPlugin.cs
jmaeagle99: Is the concern about being binary compatible (so if this library is update without rebuilding the consuming library) or source compatible (consumer of library updates temporal SDK version and doesn't get a build break)?

src/Temporalio/Client/ITemporalClientPlugin.cs
```diff
@@ -0,0 +1,33 @@
+using System;
+using System.Threading.Tasks;
+
+namespace Temporalio.Client
+{
+    /// <summary>
+    /// Interface for temporal client plugins.
+    /// </summary>
+    /// <remarks>
+    /// WARNING: This API is experimental and may change in the future.
```
jmaeagle99, line 10: For my learning, is this how experimental APIs are typically flagged in code and possibly via documentation? There might a more DevEx "friendlier" options for newer versions of .NET.

src/Temporalio/Common/SimplePlugin.cs
```diff
@@ -0,0 +1,243 @@
...
+        /// </summary>
+        public string Name { get; }
+
+        /// <summary>
+        /// Configures the client options.
+        /// </summary>
+        /// <param name="options">The client options to configure.</param>
+        public virtual void ConfigureClient(TemporalClientOptions options)
+        {
+            options.DataConverter = Resolve(options.DataConverter, Options.DataConverterOption);
```
jmaeagle99, line 50: Is there an assumption or guarantee that the same `TemporalClientOptions` instance is not used for multiple `ITemporalClientPlugin` implementations (if that is supported)? If not, this could be stomping on the configuration of other plugins.

src/Temporalio/Common/SimplePlugin.cs
```diff
@@ -0,0 +1,243 @@
...
+            return continuation(options);
+        }
+
+        /// <summary>
+        /// Configures the worker options.
+        /// </summary>
+        /// <param name="options">The worker options to configure.</param>
+        public virtual void ConfigureWorker(TemporalWorkerOptions options)
+        {
+            DoAppend(options.Activities, Options.Activities);
```
jmaeagle99, line 73: Similar concern about stomping on configuration of other plugin instances but for the same plugin type.

src/Temporalio/Common/SimplePluginOptions.cs
```diff
@@ -0,0 +1,245 @@
...
+
+            /// <summary>
+            /// Gets the configurable function for the required option.
+            /// </summary>
+            public Func<T, T>? Configurable { get; }
+
+            /// <summary>
+            /// Gets the constant value for the required option.
+            /// </summary>
+            public T? Constant { get; }
```
jmaeagle99, line 242: Why is this nullable? Would it not be sufficient to declare the usage of the `SimplyPluginOption<T>` as nullable itself? The only scenario I can think of where you might need this is if you want to say that this option doesn't have a non-null default value and you want to have the configurable callback; this feels like a simplified design of the [options pattern](https://learn.microsoft.com/en-us/dotnet/core/extensions/options#options-interfaces) in .NET where any number of consumers of the library could implement `IConfigureOptions<T>` and `IPostConfigureOptions<T>` for declaring and registering callbacks for mutating options before a consume of the options gets the final value.

src/Temporalio/Common/SimplePluginOptions.cs
```diff
@@ -0,0 +1,245 @@
...
+        public SimplePluginOption<IReadOnlyCollection<Type>?>? WorkflowFailureExceptionTypesOption
+        {
+            get;
+            set;
+        }
+
+        /// <summary>
+        /// Gets or sets a function to run before running worker/replayer.
+        /// </summary>
+        public Func<Task>? RunContextBefore { get; set; }
```
jmaeagle99, line 112: What's the intended functional use of the `RunContextBefore` and `RunContextAfter` callbacks? They don't have any context passed to them, so it feels like they would be there for generic logging/testing purpose.

jmaeagle99 (commented):
Wrote some thoughts down, but feel free to take or leave any of them.

src/Temporalio/Common/SimplePluginOptions.cs
tconley1428: More for lifetime management. Starting a DB and shutting it down or something. We wanted to give it more context but had generic type problems with the difference between the return types.

src/Temporalio/Client/TemporalClient.cs
tconley1428: Fair point, but not really. They're creating a whole client, this will be negligible.

src/Temporalio/Client/ITemporalClientPlugin.cs
tconley1428: Dunno. Dotnet owners can speak to that.

src/Temporalio/Common/SimplePlugin.cs
tconley1428: In other languages I returned a new one, @cretz wanted to do modification in place in dotnet. Thoughts?

src/Temporalio/Common/SimplePluginOptions.cs
tconley1428: That does sound similar. It's trying to mimic what I did in the other languages with either taking a value or a function that produces it. The options pattern seems to be doing the same thing.

src/Temporalio/Client/TemporalClient.cs
cretz: Yeah, in general the .NET SDK, while it takes care not to do unnecessary instantiations for the most part and does care about performance, it doesn't care about performance _that much_ because it's all lost in our C bridge and proto use and network use anyways. Having said that, yes I'd have probably done the conditional around the loop.

src/Temporalio/Client/ITemporalClientPlugin.cs
cretz: This is how we've traditionally flagged API as experimental, but totally open to other approaches. Though we do have to be compatible with several versions of .NET at compile time and can't really take a dependency, so can't really use `[Experimental]` I don't think, at least not easily. But if there is a way to make that easy (e.g. easy way to not include it with certain .NET targets) without too much heavy lifting, definitely am ok w/ that.

src/Temporalio/Common/SimplePlugin.cs
cretz: There is not that guarantee. Our options classes are meant to be instantiated/used only once, but are simple mutable POCOs due to options pattern requirements/simplicity. We intentionally didn't make options `record`s. So we just ask users to be careful. We also added (mostly) shallow cloning to help users a bit, but not too much.

We could consider cloning options just as they are passed in to our methods (e.g. the constructor of `TemporalClient` in this case) we just haven't thus far and kept it simple and kinda trust users. This is mostly because many options are not useful after connect/instantiation.

src/Temporalio/Common/SimplePluginOptions.cs
cretz: In this side of the SDK, we don't make any assumptions about `IConfigureOptions` and such, so we can only provide naive forms of callback-based options mutations. So this is effectively a "constant or function" mutually exclusive option value. Technically we could have made a whole abstract base class sum type thing, but this class should be rarely used anyways.

src/Temporalio/Common/SimplePluginOptions.cs
```diff
@@ -0,0 +1,245 @@
...
+        public virtual object Clone()
+        {
+            return (SimplePluginOptions)MemberwiseClone();
+        }
+
+        /// <summary>
+        /// Represents a required configurable plugin option.
+        /// </summary>
+        /// <typeparam name="T">The type of the option value.</typeparam>
+        public class SimplePluginOption<T>
```
cretz, line 220: Small bump of https://github.com/temporalio/sdk-dotnet/pull/548#discussion_r2466847847. I personally think `SimplePluginOptions.Option` is clearer than `SimplePluginOptions.SimplePluginOption`, but not a big deal. Just want to make sure you saw that comment/suggestion.

src/Temporalio/Client/ITemporalClientPlugin.cs
cretz: Source compatible is really the only concern. It is _technically_ a compatibility break for us to add an overload of an interface method because even though most .NET/C# versions support default interface methods allowing you to add interface methods with bodies, we technically have to surround that with `#if NETCOREAPP3_0_OR_GREATER`, so we try to think ahead of time to not have to extend the interface in the future.

However, we've been fine adding methods to interfaces in the past with that precompiler condition and breaking older .NET versions, it's just that we prefer to avoid it.

tests/Temporalio.Tests/Common/PluginTests.cs
```diff
@@ -0,0 +1,239 @@
...
+            throw new NotImplementedException();
+    }
+
+    private class WorkerPlugin : ITemporalWorkerPlugin
+    {
+        public string Name => "WorkerPlugin";
+
+        public void ConfigureWorker(TemporalWorkerOptions options)
+        {
+            options.TaskQueue = "NewTaskQueue";
```
yuandrew: Could you make sure task queues are unique (append a UUID) for each test? This is something we've always recommended (and I believe all existing tests all follow this), but only with my recent heartbeat changes do we actually fail tests for this. 

I've fixed up the python and TS tests for this when introducing heartbeat, if you can have them be unique for this and later SDK impl's, that would be super appreciated!

tconley1428: Sure.

src/Temporalio/Common/SimplePluginOptions.cs
tconley1428: I tried, `Option` is a reserved keyword so the linter complains about doing that. Didn't seem worth overriding.

PR #562, opened 2025-11-18
Update tests and generator to .NET 10

### LINKED ISSUE #522: [Feature Request] Consider updating tests to use .NET 10

We should probably update `tests/Temporalio.Tests/Temporalio.Tests.csproj` to use `<TargetFramework>net10.0</TargetFramework>` instead of `.net6.0` it does today. Also, while we had to downgrade the `dotnetVersionOverride` in `ci.yml` for Windows for #517 due to new dependencies requiring 6.x (we broke off this issue at that time), as part of this, we can probably just completely remove `dotNetVersionOverride` concept in `ci.yml`.

There should be a discussion on _whether_ we should stay on .NET 6 forever to allow our tests to make sure we run on at least this .NET version forever.

jmaeagle99:
## What was changed
- Updated test projects and generator project to .NET 10.
- Fixed code analysis violations and suppressed those that should not be fixed within workflows.
- Adjusted standard library call tests due to changes in .NET 8+ runtime.

## Why?

Lowest supported .NET version is 8 and both 8 and 9 will be out of support in less than a year at this point. Update to latest LTS (.NET 10) which will be supported for ~3 years.

## Checklist

1. Closes #522 

2. How was this tested: Local builds and test runs with only .NET 10 SDK installed. Built and ran all projects found to test the SDK tests, smoke test, benchmark, and docfx generation. Checked test counts before and after changes to check that tests were not dropped.

3. Any docs updates needed? None required. Can consider if the `Task.WhenAny -> Workflow.WhenAnyAsync` recommendation should be adjusted (as well as the SDK implementation of `Workflow.WhenAnyAsync`) for .NET 8+.

### DIFF
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -136,6 +136,8 @@ jobs:
 
       - name: Setup .NET
         uses: actions/setup-dotnet@v4
+        with:
+          dotnet-version: 10.x
 
       - name: Build package
         run: dotnet pack -c Release /p:BridgeLibraryRoot=${{ github.workspace }}/bridge-libraries
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -32,7 +32,7 @@ jobs:
             out-file: libtemporal_sdk_core_c_bridge.so
             out-prefix: linux-musl-x64
             # Need Alpine container since GH runner doesn't have one
-            container: mcr.microsoft.com/dotnet/sdk:8.0-alpine
+            container: mcr.microsoft.com/dotnet/sdk:10.0-alpine
             protobuf-url: https://github.com/protocolbuffers/protobuf/releases/download/v22.3/protoc-22.3-linux-x86_64.zip
             runsOn: ubuntu-latest
           - os: macos-intel
@@ -164,7 +164,7 @@ jobs:
           - os: ubuntu-arm
             runsOn: ubuntu-24.04-arm64-2-core
           - os: alpine-latest
-            container: mcr.microsoft.com/dotnet/sdk:6.0-alpine
+            container: mcr.microsoft.com/dotnet/sdk:10.0-alpine
             runsOn: ubuntu-latest
           - os: macos-intel
             runsOn: macos-13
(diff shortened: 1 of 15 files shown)


.github/workflows/package.yml
cretz, line 1: Can you do me a favor and push a commit that temporarily adds `pull_request` to `on` for this GH workflow so this PR will run this GH workflow and we can confirm it still works as expected? We can then remove it before merging.

tests/Temporalio.Tests/KitchenSinkWorkflow.cs
```diff
@@ -195,7 +195,12 @@ action with
                 tasks.Add(Task.Factory.StartNew(async () =>
                 {
                     await Workflow.DelayAsync((int)action.ExecuteActivity.CancelAfterMS);
+#pragma warning disable CA1849 // Call async methods when in an async method
+#pragma warning disable VSTHRD103 // Call async methods when in an async method
+                    // https://github.com/temporalio/sdk-dotnet/issues/327
                     cancelSource.Cancel();
```
cretz, line 201: I think we should consider updating the README "Workflow .editorconfig" section to exclude both of these rules in `.editorconfig` for `.workflow.cs` (what we tell people to suffix their workflow filenames with to differentiate the rules and what we also do in samples). I'd be ok excluding them in the test `.editorconfig` too, though this is fine.

cretz (commented):
Looks great, nothing blocking. But if possible, would like to see that "Build Package" workflow run inside this PR w/ a quick temporary update to confirm it works w/ the new containers/changes.

tests/Temporalio.Tests/KitchenSinkWorkflow.cs
jmaeagle99: I thought about excluding these rules with a global suppression in the tests but thought it would be better if we observed these rules in non-workflow test code. Maybe something that could be done is to refactor the tests to move all of the workflows to *.workflow.cs files and use the same workflow-specific rules that we recommend to customers and dogfood our own recommendations; this would be a separate change/effort from this PR though.

.github/workflows/package.yml
jmaeagle99: Found an issue where the correct .NET SDK was assumed to be on the runner. Fixed by adding the the dotnet_version parameter to the task and revalidated that it works. I've pushed the reversion commit to make the changes ready for another pass.

cretz: Yup, I see https://github.com/temporalio/sdk-dotnet/actions/runs/19485143464, thanks!

tests/Temporalio.Tests/KitchenSinkWorkflow.cs
cretz: Agreed, can just exclude inline for now. Refactoring tests to move to `.workflow.cs` files is too much for sure and I could see us not wanting to skip this rule for non-workflow code. Hopefully one day we can write a Roslyn analyzer that can do better at proper diagnostic config/reporting.

cretz (approved):
Looks great, merge at will

PR #567, opened 2025-12-02
💥  Enable TLS if api key specified

THardy98:
## What was changed
DWISOTT

1. Part of https://github.com/temporalio/features/issues/687

2. How was this tested:
Unit tests

3. Any docs updates needed?
Maybe

### DIFF
--- a/src/Temporalio/Client/TemporalConnectionOptions.cs
+++ b/src/Temporalio/Client/TemporalConnectionOptions.cs
@@ -10,8 +10,6 @@ namespace Temporalio.Client
     /// </summary>
     public class TemporalConnectionOptions : ICloneable
     {
-        private TlsOptions? tls;
-
         /// <summary>
         /// Initializes a new instance of the <see cref="TemporalConnectionOptions"/> class.
         /// Create default options.
@@ -42,16 +40,9 @@ namespace Temporalio.Client
         /// <remarks>
         /// This must be set, even to a default instance, to do any TLS connection. If not set and
         /// <see cref="ApiKey"/> is provided, TLS will be automatically enabled with default options.
+        /// To explicitly disable TLS, set instance with <see cref="TlsOptions.Disabled"/> set to true.
         /// </remarks>
-        public TlsOptions? Tls
-        {
-            get => tls;
-            set
-            {
-                tls = value;
-                TlsExplicitlySet = true;
-            }
-        }
+        public TlsOptions? Tls { get; set; }
 
         /// <summary>
         /// Gets or sets retry options for this connection.
@@ -115,11 +106,6 @@ namespace Temporalio.Client
         public TemporalRuntime? Runtime { get; set; }
 
         /// <summary>
-        /// Gets a value indicating whether TLS was explicitly set (even to null).
-        /// </summary>
-        internal bool TlsExplicitlySet { get; private set; }
-
-        /// <summary>
         /// Create a shallow copy of these options.
         /// </summary>
         /// <returns>A shallow copy of these options and any transitive options fields.</returns>
@@ -129,7 +115,7 @@ namespace Temporalio.Client
             var copy = (TemporalConnectionOptions)MemberwiseClone();
             if (Tls != null)
             {
-                copy.tls = (TlsOptions)Tls.Clone();
+                copy.Tls = (TlsOptions)Tls.Clone();
             }
             if (RpcRetry != null)
             {
--- a/src/Temporalio/Client/TemporalConnectionOptions.cs
+++ b/src/Temporalio/Client/TemporalConnectionOptions.cs
@@ -10,6 +10,8 @@ namespace Temporalio.Client
     /// </summary>
     public class TemporalConnectionOptions : ICloneable
     {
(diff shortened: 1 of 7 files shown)


tests/Temporalio.Tests/Client/TemporalConnectionOptionsTests.cs
```diff
@@ -0,0 +1,41 @@
...
+        {
+            ApiKey = "test-api-key",
+            Identity = "test-identity",
+        };
+
+        using var scope = new Scope();
+        var interopOptions = options.ToInteropOptions(scope);
+
+        // TLS should be auto-enabled when API key is provided and TLS not explicitly set
+        Assert.True(interopOptions.tls_options != null);
```
jmaeagle99, line 22: `Assert.NotNull` will give better error message when condition fails.

tests/Temporalio.Tests/Client/TemporalConnectionOptionsTests.cs
```diff
@@ -0,0 +1,41 @@
...
+            ApiKey = "test-api-key",
+            Identity = "test-identity",
+            Tls = null, // Explicitly disable TLS
+        };
+
+        using var scope = new Scope();
+        var interopOptions = options.ToInteropOptions(scope);
+
+        // TLS should remain disabled when explicitly set to null
+        Assert.True(interopOptions.tls_options == null);
```
jmaeagle99, line 39: `Assert.Null` will give better error message when condition fails.

tests/Temporalio.Tests/Client/TemporalConnectionOptionsTests.cs
```diff
@@ -0,0 +1,41 @@
+namespace Temporalio.Tests.Client;
+
+using Temporalio.Bridge;
+using Temporalio.Client;
+using Xunit;
+
+public unsafe class TemporalConnectionOptionsTests
```
jmaeagle99, line 7: Could you add tests for when:
1) `ApiKey` is not set, is null, or is empty?
1) `Tls` is set to a non-null value?

src/Temporalio/Client/TemporalConnectionOptions.cs
```diff
@@ -103,6 +114,11 @@ public TemporalConnectionOptions()
         /// </remarks>
         public TemporalRuntime? Runtime { get; set; }
 
+        /// <summary>
+        /// Gets a value indicating whether TLS was explicitly set (even to null).
+        /// </summary>
+        internal bool TlsExplicitlySet { get; private set; }
```
cretz: I don't think we should take this approach. The difference between a user manually doing `= null` and the default of `null` would be very confusing. I think for .NET we should add a `bool Disabled` property to `TlsOptions` for users that want to explicitly disable TLS. `null` should be considered unset/default (regardless of whether it's set that way)

tests/Temporalio.Tests/Client/TemporalConnectionOptionsTests.cs
cretz, line 1: Hrmm, these tests may be bit too detailed/low-level. You are testing internals and only relying on us making internals visible to this test project. Note we don't have any other tests confirming the bridge options conversions. Usually we test user-facing behavior.

How would a user write a test to confirm their setting of TLS disabled or API key or whatever does the right thing? Granted it's probably not very easy to do this. One approach that you may take is to alter the "Test cloud" CI test to use API key instead of mTLS to at least confirm the API key default.

Regardless, not a big deal if you keep these, just a bit strange for this repo (and why there never was anything like this in the past).

src/Temporalio/Client/TemporalConnectionOptions.cs
THardy98: Yeah, this is a bit nicer. Done

tests/Temporalio.Tests/Client/TemporalConnectionOptionsTests.cs
THardy98: Probably going to keep them. Changed the cloud CI workflow to use api key secret and remove the mTLS secrets

cretz (approved):
But would also like @jmaeagle99's approval

tests/Temporalio.Tests/Client/TemporalConnectionOptionsTests.cs
THardy98: `tls_options` is an object pointer, can't use `Null/NotNull` assertions, at least not to my knowledge

tests/Temporalio.Tests/Client/TemporalConnectionOptionsTests.cs
THardy98: `tls_options` is an object pointer, can't use `Null/NotNull` assertions, at least not to my knowledge

tests/Temporalio.Tests/Client/TemporalConnectionOptionsTests.cs
THardy98: Added

jmaeagle99 (approved):

PR #572, opened 2025-12-08
Fix logging message to clarify that the value is a run ID

johnricketts:
## What was changed
Fixed a logging message to clarify that the ID is a workflow run ID and not the workflow ID

## Why?
Message was slightly misleading before

## Checklist

1. Closes 

2. How was this tested:
Not required?

3. Any docs updates needed?
No?


cretz (approved):

PR #570, opened 2025-12-04
💥Enable Worker heartbeating, update Core to 44a6576

yuandrew:
## What was changed

- Enabled worker heartbeat
- plumb plugin names (both client and worker) to core
- update core to 44a6576
  - 💥temporal_sdk_core_c_bridge.dll got renamed to temporalio_sdk_core_c_bridge.dll

## Why?

New worker heartbeating feature

## Checklist

1. Closes #551

2. How was this tested:
Manually tested

3. Any docs updates needed?

---

> [!NOTE]
> Enables worker heartbeating, plumbs plugin names to Core, updates interop/bindings and build paths, and renames the native bridge library to temporalio_sdk_core_c_bridge across code, CI, and docs.
> 
> - **Runtime/Worker**:
>   - Add `TemporalRuntimeOptions.WorkerHeartbeatInterval` and propagate to Core (`worker_heartbeat_interval_millis`).
>   - Pass plugin names (client + worker) to Core via `TemporalCoreByteArrayRefArray` on client/worker options.
>   - Replace `no_remote_activities` with `TemporalCoreWorkerTaskTypes` in `TemporalCoreWorkerOptions` and populate based on registered work.
>   - `Bridge.Worker` constructor accepts `clientPlugins` and forwards to interop conversion.
> - **Interop/Core bindings**:
>   - Update header path to `sdk-core/crates/sdk-core-c-bridge` and adjust all `DllImport` names and library path to `temporalio_sdk_core_c_bridge`.
>   - Add new interop structs/fields: `TemporalCoreWorkerTaskTypes`, extra fields on client/runtime/worker options.
>   - Rename Rust crate log/targets in telemetry defaults from `temporal_*` to `temporalio_*`.
> - **Build/CI/Packaging**:
>   - Rename emitted artifacts to `temporalio_sdk_core_c_bridge.*`; update `.csproj`, `build/net462/Temporalio.targets`, and GitHub Actions workflow to new paths and Cargo manifest.
>   - Adjust CI matrices/includes and smoke-test config; use Manylinux Alpine/ARM containers.
> - **Docs/Tests**:
>   - Update README and test expectations for new library name and log targets; minor test fixes (unique task queues, metric assertions).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 2bbb1f484d14b8aca0d7812def95b15c639b4036. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #574, opened 2025-12-11
Fix api generator and sdk-core protos path

jmaeagle99:
## What was changed

- Suppress performance analysis rule in API generator program.
- Fix protos path within sdk-core submodule.
- Regenerate protos based on current sdk-core submodule version.

## Why?

I saw that the "Regen confirm unchanged" step of the CI workflow was failing silently. Example at https://github.com/temporalio/sdk-dotnet/actions/runs/20114313589/job/57719795349#step:7:21
- The update to .NET 10 likely caused the code analysis warning that fails the compilation of the Temporalio.Api.Generator project.
- After fixing the project compilation, the update to the latest sdk-core caused the project to not find the protos for regeneration since the crates were refactored.

## Checklist

- How was this tested: Will check CI workflow


src/Temporalio/Api/Cloud/Account/V1/Message.cs
cretz, line 1: We have a CI step that is only run on `windows-latest` called "Regen confirm unchanged" that is supposed to catch if these are mismatched, but checking #570, it passed with an error of:

> Error: D:\a\sdk-dotnet\sdk-dotnet\src\Temporalio.Api.Generator\Program.cs(248,37): error CA1870: Use a cached 'SearchValues' instance for improved searching performance (https://learn.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1870) [D:\a\sdk-dotnet\sdk-dotnet\src\Temporalio.Api.Generator\Temporal.Api.Generator.csproj]

I know that's already solved in this PR, but I wonder if we need to update the CI job to properly fail if that `dotnet run` in that inline script fails. AI is mentioning maybe we should add `$ErrorActionPreference = "Stop"` to the top of the inline script.

jmaeagle99: I thought the same thing but left it as-is initially to maintain existing behavior. But more than happy to add it to this PR.

cretz (approved):

PR #575, opened 2025-12-12
Support for gRPC binary metadata headers

### LINKED ISSUE #512: [Feature Request] Ensure gRPC binary metadata headers are supported

See https://github.com/temporalio/features/issues/671. Probably needs https://github.com/temporalio/sdk-core/issues/991 first, and then probably needs to support dictionary values as bytes somehow without breaking users (or maybe require `-bin` based values are base64'd by users).

jmaeagle99:
## What was changed

- Add APIs for specifying gRPC binary metadata headers and adjust binary conversions.
- Update sdk-core submodule to include gRPC binary metadata header support.
- Add small tests to validate conversion and setting of headers.

## Why?

Binary headers cannot be specified using the existing metadata APIs. These new APIs for client options and RPC options allow specifying the binary metadata headers.

## Checklist

1. Closes #512 

2. How was this tested: Add tests and validate it works with new sdk-core update.

3. Any docs updates needed? Unsure, will check with team


.gitmodules
```diff
@@ -1,3 +1,4 @@
 [submodule "src/Temporalio/Bridge/sdk-core"]
```
jmaeagle99, line 1: Thsi file will be reverted upon validation and completion of the outstanding changes in https://github.com/temporalio/sdk-core/pull/1076.

src/Temporalio/Client/ITemporalConnection.cs
```diff
@@ -24,6 +24,18 @@ public interface ITemporalConnection : IBridgeClientProvider
...
+        /// Gets or sets the current RPC binary metadata (i.e. the binary headers).
+        /// </summary>
+        /// <remarks>
+        /// This can be updated which will apply to all future calls the client makes including
+        /// inside a worker. Setting this value is thread safe. When setting, this will error if the
+        /// client is not already connected (e.g. a lazy client has not made a call).
+        /// </remarks>
+        /// <exception cref="System.InvalidOperationException">Client is not already
+        /// connected.</exception>
+        IReadOnlyCollection<KeyValuePair<string, byte[]>> RpcBinaryMetadata { get; set; }
```
jmaeagle99, line 37: I research ways to "upgrading" the `RpcMetadata` property to be able to accept both `string` and `byte[]` values for the key value pairs, using implicit conversions at various levels. It turns out that implicit conversions are not supported when the right-value is an interface type. I don't think we could update that property with a new type that supports all headers without making a breaking change, so I introduced separate `RpcBinaryMetadata`/`BinaryMetadata` properties where appropriate.

src/Temporalio/Bridge/Scope.cs
```diff
@@ -53,21 +60,65 @@ public Interop.TemporalCoreByteArrayRef ByteArray(string? str)
...
+            byteArrayRefs.Add(val);
+            return val.Ref;
+        }
+
+        /// <summary>
+        /// Create a metadata byte array ref array.
+        /// </summary>
+        /// <param name="metadata">Metadata to create from.</param>
+        /// <returns>Created byte array ref.</returns>
+        public Interop.TemporalCoreByteArrayRefArray Metadata(IEnumerable<KeyValuePair<string, string>>? metadata)
```
cretz: Looks to be a clone of `ByteArrayArray` except this one accounts for null/empty and that one doesn't. I wonder if we should just have the one and it accounts for null/empty. May not need these `Metadata` methods if the other ones exist.

src/Temporalio/Bridge/Scope.cs
```diff
@@ -53,21 +60,65 @@ public Interop.TemporalCoreByteArrayRef ByteArray(string? str)
...
         }
 
+        /// <summary>
+        /// Create a key-value pair byte array ref.
+        /// </summary>
+        /// <param name="pair">Key-value pair to create from.</param>
+        /// <returns>Created byte array ref.</returns>
+        public Interop.TemporalCoreByteArrayRef ByteArray(KeyValuePair<string, byte[]> pair)
+        {
+            if (pair.Key.Length == 0)
```
cretz: This conditional is a bit interesting. Is it just in case, or are there situations where we think it may happen? Same for the conditional in the other overload of this.

src/Temporalio/Bridge/Scope.cs
```diff
@@ -84,6 +135,22 @@ public Interop.TemporalCoreByteArrayRef NewlineDelimited(IEnumerable<string>? va
...
         }
 
+        /// <summary>
+        /// Create a newline-delimited byte array ref.
+        /// </summary>
+        /// <param name="values">Values to create from.</param>
+        /// <returns>Created byte array ref.</returns>
+        public Interop.TemporalCoreByteArrayRef NewlineDelimited(IEnumerable<KeyValuePair<string, string>>? values)
+        {
+            if (values == null)
```
cretz: In some functions you also check `.Any()` to return empty ref or not, but not here (doesn't really matter though IMO)

src/Temporalio/Client/ITemporalConnection.cs
cretz: I think that's acceptable (I am glad we added an `RpcOptions` class here unlike in some langs where we take the individual options per call). Alternative approach would have been to ask the (rare) binary users to base64 their `-bin` keys.

src/Temporalio/Common/EnvConfig/ClientEnvConfig.cs
```diff
@@ -124,6 +124,7 @@ public TemporalClientConnectOptions ToClientConnectionOptions()
                 }
 
                 // Add gRPC metadata if present
+                // Add support for gRPC binary metadata?
```
cretz: Nah, we can leave env config as string headers only for now IMO

cretz (approved):
Nothing blocking, just pedantic overload stuff on `internal` things

src/Temporalio/Client/ITemporalConnection.cs
jmaeagle99: I think if users base64 encoded their values, then they would come out and be visible as base64 instead of the raw binary data. My understanding is that tonic and lower libraries already base64 encode the binary headers for transport and will decode them before they become visible again.

src/Temporalio/Bridge/Scope.cs
jmaeagle99: I've removed it. From this function's perspective, it is value to have an empty key.

cretz (approved):
Looks great, thanks! (though as mentioned in comment, have to merge https://github.com/temporalio/sdk-core/pull/1076 first and update submodule stuff here)

PR #580, opened 2025-12-18
Allow disabling of workflow tracing event listener in custom meter

### LINKED ISSUE #573: [Feature Request] Allow custom meter implementation to disable tracing event listener

Today, a user providing a `CustomMetricMeter` may set an underlying .NET meter that uses threads/tasks in a non-deterministic manner, which means metrics emitted from user workflows will trip the tracing event listener. We should offer an option on the `CustomMetricMeter` that will put calls to the underlying `Meter` inside `Workflow.Unsafe.WithTracingEventListenerDisabled` if `Workflow.InWorkflow` is `true`. Whether this option is enabled by default or not is TBD, but probably will be on by default (and therefore would be ok if this wasn't a user-facing option at all).

cretz:
## What was changed

Added constructor option for `disableWorkflowTracingEventListener` for `Temporalio.Extensions.DiagnosticSource.CustomMetricMeter` that, when `true`, will make .NET meter calls with workflow tracing event listener disabled.

## Checklist

1. Closes #573

### DIFF
--- a/tests/Temporalio.Tests/Extensions/DiagnosticSource/CustomMetricMeterTests.cs
+++ b/tests/Temporalio.Tests/Extensions/DiagnosticSource/CustomMetricMeterTests.cs
@@ -353,4 +353,74 @@ public class CustomMetricMeterTests : WorkflowEnvironmentTestBase
         }
         Assert.Equal(7.89, gaugeFloat.Value);
     }
+
+    [Workflow]
+    public class SimpleCounterWorkflow
+    {
+        [WorkflowRun]
+        public async Task RunAsync() =>
+            Workflow.MetricMeter.CreateCounter<int>("my-counter").Add(10);
+    }
+
+    [Theory]
+    [InlineData(false)]
+    [InlineData(true)]
+    public async Task CustomMetricMeter_Workflow_CanDisableTracing(bool disableTracing)
+    {
+        // Create a meter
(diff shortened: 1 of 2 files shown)


tests/Temporalio.Tests/Extensions/DiagnosticSource/CustomMetricMeterTests.cs
```diff
@@ -353,4 +353,74 @@ await worker.ExecuteAsync(() =>
...
+        using var meterListener = new MeterListener();
+        meterListener.InstrumentPublished = (instrument, listener) =>
+        {
+            if (instrument.Meter.Name == "test-disable-tracing-meter")
+            {
+                listener.EnableMeasurementEvents(instrument);
+            }
+        };
+        meterListener.SetMeasurementEventCallback<long>(
+            (inst, value, tags, state) => Task.Run(() => { }));
```
jmaeagle99, line 383: To be clear, this `Task.Run` is the "bad thing", correct?

jmaeagle99 (approved):

tests/Temporalio.Tests/Extensions/DiagnosticSource/CustomMetricMeterTests.cs
cretz: Right. And we assert it is a bad thing when by confirming task failure when tracing event listener is not disabled.

PR #581, opened 2025-12-19
Upload symbols from build

jmaeagle99:
## What was changed

Upload libraries and symbol files are part of CI build.

## Why?

Having these type of files make it significantly easier to debug crash dumps so that symbols and call frames are matched.


.github/workflows/ci.yml
```diff
@@ -85,6 +85,28 @@ jobs:
 
       - name: Build
         run: dotnet build
+      
+      - name: Upload native symbols
+        uses: actions/upload-artifact@v4
+        if: always()
```
cretz: I wonder if we should give the "Test" step an ID of `test`, move these two steps alongside "Upload test failure", and set `if: steps.test.outcome == 'failure'` for all 3 (or whatever the expression is). That way we don't upload these artifacts all the time for every CI run. Though if we think there is value in doing so for every run, we can.

jmaeagle99: For compiled languages, files that contain symbols are required for debugging (e.g. with a debugger) and they must match the binaries (so, dylib, dll, etc) exactly (e.g. must be built and preserved along with the binaries); these cannot be recreated after the official build is produced. Otherwise, there is no way to get any meaningful call stacks when debugging dumps or live processes. So I would have these uploaded regardless of whether they occurred under test or not. Uploading these files only add about 5-6 seconds of build time. I think the tradeoff for the minimal build time increase is outweighed by the ability to debug and get useful debugging information.

cretz: Sorry, to clarify, I am commenting on the entire collection of artifacts. To be honest, less concerned with build time than I am with so many megs of unnecessary artifacts on our builds. I was just suggesting we don't need any artifacts unless there is a test failure correct? It is at that time, it makes sense to upload all of them (shared libs, debug symbols, etc). I'm just saying we can move these two upload steps to only be on test failure. But if you really want them on every build for every commit to every PR, I suppose we can add them.

jmaeagle99: I'll switch it to only upload on test failures. I was hoping to have it for any build so that if we are testing with unreleased bits and found some bug that requires debugging, then we'd have the symbols. But I suppose we can just rebuild from source locally to reproduce the bug; inconvenient, but not impossible.

cretz (approved):

Issue #579, opened 2025-12-17
[Bug] SIGSEGV crash in temporal_core_worker_poll_workflow_activation on Linux


cfauchere:
### What are you really trying to do?

  Running integration tests for a .NET 8 application that uses Temporal workflows. Tests run in CI on Linux (Ubuntu) with an external Temporal server container (temporalio/temporal:latest).

 ### Describe the bug

  The test host process crashes with SIGSEGV (signal 11) while the Temporal worker is polling for workflow activations. The crash happens intermittently (flaky) - tests pass locally on Windows but crash randomly in Linux CI.

  Shortly before the crash, this warning appears in logs:
  WARN temporal_sdk_core::worker: Activity heartbeat failed. task_token=... details=UnknownActivity

  ### Stack trace from crash dump:
  Temporalio.Bridge.Interop.Methods.temporal_core_worker_poll_workflow_activation(TemporalCoreWorker*, Void*, IntPtr)
  ILStubClass.IL_STUB_PInvoke(TemporalCoreWorker*, Void*, IntPtr)
  Temporalio.Bridge.Worker.PollWorkflowActivationAsync()
  Temporalio.Worker.WorkflowWorker+<ExecuteAsync>d__10.MoveNext()
  ...
  System.Threading.PortableThreadPool+WorkerThread.WorkerThread

  The crash occurs in the native Rust bridge during poll_workflow_activation, not during execution of any user workflow/activity code.

  ### Minimal Reproduction

  Unable to provide minimal repro - the crash is intermittent and only occurs in Linux CI environment, not locally on Windows. The test runs Temporal workflows with activities that use heartbeats.

  ### Environment/Versions

  - OS and processor: Linux x86_64 (Ubuntu, Azure DevOps hosted agent)
  - Temporal .NET SDK Version: 1.8.0
  - Temporal Server: temporalio/temporal:latest (Docker container, server start-dev)
  - .NET Version: 8.0
  - Works on: Windows x64 (local development)
  - Fails on: Linux x64 (CI)

cfauchere:
```
Full stack
        Child SP               IP Call Site
00007EF4E3FFA060 00007f364e710813 [InlinedCallFrame: 00007ef4e3ffa060] Temporalio.Bridge.Interop.Methods.temporal_core_worker_poll_workflow_activation(Temporalio.Bridge.Interop.TemporalCoreWorker*, Void*,
IntPtr)
[...]

PR #583, opened 2025-12-23
Dedicated safe handles for Client and Worker

### LINKED ISSUE #577: [Bug] Investigate test host crash flake
### Describe the bug

CI tests on `main` branch have been running into test host process crashes. The following are examples of workflow runs with the crash and their corresponding change:
| Date | Job | Variant | Commit |
|---|---|---|---|
| 2025-12-12 | https://github.com/temporalio/sdk-dotnet/actions/runs/20175327021/job/57921738811 | ubuntu-arm | https://github.com/temporalio/sdk-dotnet/commit/79d44c3472bc77fe3e294bf5b9cfe01fdf896527 |
| 2025-12-10 | https://github.com/temporalio/sdk-dotnet/actions/runs/20110892146/job/57707557020 | ubuntu-arm | https://github.com/temporalio/sdk-dotnet/commit/2149661d87fb6cf091066d1f6b6056e486696240 |
| 2025-12-02 | https://github.com/temporalio/sdk-dotnet/actions/runs/19900160973/job/57041412266 | ubuntu-arm | https://github.com/temporalio/sdk-dotnet/commit/549285ab87773de9120704b788116f957d76b03a |
| 2025-11-17 | https://github.com/temporalio/sdk-dotnet/actions/runs/19445103513/job/55637681567 | macos-intel | https://github.com/temporalio/sdk-dotnet/commit/5161e0bc814baf0e382cc4e80c465c1e3d2d9669 |
| 2025-11-07 | https://github.com/temporalio/sdk-dotnet/actions/runs/19186311432/job/54853416545 | ubuntu-arm | https://github.com/temporalio/sdk-dotnet/commit/31e754f863a815cc9563d4c85df69c8426f42aa8 |

The test that was running when the crash occurs is not consistent across runs.

jmaeagle99:
## What was changed

- Moved lifetime ownership of unmanaged client and worker pointers to new `SafeHandle` classes.
- Update `Worker` no longer be a `SafeHandle` but rather add-ref to an existing `SafeHandle` or create its own.
- Update `Worker` and `Client` to add-ref to their corresponding handle before calling p/invoke functions. This is done using the `Scope.Pointer` method.

## Why?

Moving ownership of lifetime of unmanaged pointers to `SafeHandle`s and performing add-ref and release operations as necessary will prevent the freeing of unamanaged pointers and their related resources. If an add-ref cannot be fulfilled, a managed exception will be thrown instead of AVing in unmanaged code.

This change can be used as a pattern to update the remaining bridge classes to separate unmanaged resource lifetime from the wrapper bridge classes and prevent use-after-free issues of unmanaged pointers.

## Checklist

1. Closes #577 and #579

2. How was this tested: Repeatedly tested in CI tests (about 50 times); no more crashes observed.

3. Any docs updates needed? No

### DIFF
--- a/src/Temporalio/Bridge/Client.cs
+++ b/src/Temporalio/Bridge/Client.cs
@@ -15,11 +15,11 @@ namespace Temporalio.Bridge
             : base((IntPtr)ptr, true)
         {
             Runtime = runtime;
-            Ptr = ptr;
+            Handle = new SafeClientHandle(ptr);
         }
 
         /// <inheritdoc />
-        public override unsafe bool IsInvalid => Ptr == null;
+        public override bool IsInvalid => Handle.IsInvalid;
 
         /// <summary>
         /// Gets the runtime associated with this client.
@@ -27,9 +27,9 @@ namespace Temporalio.Bridge
         internal Runtime Runtime { get; private init; }
 
         /// <summary>
-        /// Gets the pointer to the client.
+        /// Gets the safe handle for the client.
         /// </summary>
-        internal unsafe Interop.TemporalCoreClient* Ptr { get; private init; }
+        internal SafeClientHandle Handle { get; private init; }
 
         /// <summary>
         /// Connect to Temporal.
@@ -80,7 +80,9 @@ namespace Temporalio.Bridge
             {
                 unsafe
                 {
-                    Interop.Methods.temporal_core_client_update_metadata(Ptr, scope.ByteArrayArray(metadata));
+                    Interop.Methods.temporal_core_client_update_metadata(
+                        scope.Pointer(Handle),
+                        scope.ByteArrayArray(metadata));
                 }
             }
         }
@@ -95,7 +97,9 @@ namespace Temporalio.Bridge
             {
                 unsafe
                 {
-                    Interop.Methods.temporal_core_client_update_binary_metadata(Ptr, scope.ByteArrayArray(metadata));
+                    Interop.Methods.temporal_core_client_update_binary_metadata(
+                        scope.Pointer(Handle),
+                        scope.ByteArrayArray(metadata));
                 }
             }
         }
@@ -110,7 +114,9 @@ namespace Temporalio.Bridge
             {
                 unsafe
                 {
-                    Interop.Methods.temporal_core_client_update_api_key(Ptr, scope.ByteArray(apiKey));
+                    Interop.Methods.temporal_core_client_update_api_key(
+                        scope.Pointer(Handle),
+                        scope.ByteArray(apiKey));
                 }
             }
         }
@@ -148,7 +154,7 @@ namespace Temporalio.Bridge
                 unsafe
                 {
                     Interop.Methods.temporal_core_client_rpc_call(
-                        Ptr,
+                        scope.Pointer(Handle),
                         scope.Pointer(
                             new Interop.TemporalCoreRpcCallOptions()
                             {
@@ -207,7 +213,7 @@ namespace Temporalio.Bridge
         /// <inheritdoc />
         protected override unsafe bool ReleaseHandle()
         {
-            Interop.Methods.temporal_core_client_free(Ptr);
+            Handle.Dispose();
             return true;
         }
     }
--- a/src/Temporalio/Bridge/SafeUnmanagedHandle.cs
+++ b/src/Temporalio/Bridge/SafeUnmanagedHandle.cs
@@ -1,0 +1,29 @@
+using System;
+using Microsoft.Win32.SafeHandles;
+
+namespace Temporalio.Bridge
+{
+    /// <summary>
+    /// Safe handle for unmanaged instance.
+    /// </summary>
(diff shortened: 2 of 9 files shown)


src/Temporalio/Bridge/Client.cs
```diff
@@ -15,21 +15,21 @@ private unsafe Client(Runtime runtime, Interop.TemporalCoreClient* ptr)
             : base((IntPtr)ptr, true)
         {
             Runtime = runtime;
-            Ptr = ptr;
+            Handle = new SafeClientHandle(ptr);
```
cretz, line 18: Hrmm, so the goal of this `Client` class was to be that `SafeClientHandle`. What is the purpose of having both `Client` class and `SafeClientHandle` class both being `SafeHandle`s referencing the same `IntPtr`? Can we either just make this class not a `SafeHandle` or merge this class and `SafeClientHandle` together? (same for `Worker`)

src/Temporalio/Bridge/SafeUnmanagedHandle.cs
```diff
@@ -0,0 +1,29 @@
+using System;
+using Microsoft.Win32.SafeHandles;
+
+namespace Temporalio.Bridge
+{
+    /// <summary>
+    /// Safe handle for unmanaged instance.
+    /// </summary>
+    /// <typeparam name="T">Unmanaged type.</typeparam>
+    internal abstract class SafeUnmanagedHandle<T> :
```
cretz, line 10: I wonder if we should make all of our classes that extend `SafeHandle` instead extend this and stop using `SafeHandle` directly just as good practice

cretz (commented):
LGTM, mostly just code structure discussions on the duplicate client/worker handle classes and if we can apply this safe unmanaged handle base class more broadly.

src/Temporalio/Bridge/Client.cs
jmaeagle99: > Can we either just make this class not a `SafeHandle`

That is what I would like to happen eventually.

tl;dr: Making this change will cause a lot more changes and have wider impact than what is currently proposed to fix the immediate issue.

Details:
Making that change now will create more sprawl in this PR, especially at https://github.com/temporalio/sdk-dotnet/blob/main/src/Temporalio/Worker/TemporalWorker.cs#L66. The returned value from `client.BridgeClientProvider.BridgeClient` would be the `SafeClientHandle`, which means the `TemporalWorker` cannot cast this to `Bridge.Client`. Updating the `Worker` to only take a `SafeClientHandle` instead of `Client` would be ideal, but then it won't have access to a `Runtime` instance (which eventually should be converted to `SafeHandle`). But `TemporalWorker` doesn't have access to a `Runtime` instance outside of casting `client.BridgeClientProvider.BridgeClient` to `Client`. This might require adding another property to `IBridgeClientProvider` to expose the `Runtime`/`SafeRuntimeHandle` or maybe an internal interface that can be used for accessing it so we don't change this public interface; I know there is a bunch of comments telling developers not to use it already.

I didn't want to make a very large change that might affect the public surface area, even if that surface area is documented as "don't touch". I wanted to keep this scoped to fix the immediate issue and then build upon it in subsequent changes.

If you think this should be incorporated now, I can attempt to do it. But I think the current change strikes a good balance and provides a stepping stone toward avoiding use-after-free for unmanaged pointers.

src/Temporalio/Bridge/SafeUnmanagedHandle.cs
jmaeagle99: Eventually, I think it would be best to have purpose built `Safe<Type>Handle` classes for each unmanaged pointer type and pass those around as needed. The existing classes that implement `SafeHandle` can then just implement `IDisposable`. Basically, extend the idea in this PR to all other types that are currently handling unmanaged pointer lifetime.

src/Temporalio/Bridge/Client.cs
cretz: We can do this in phases, no problem. Can we open an issue to track that work? The idea of the unmanaged pointer only being held by one safe handle makes sense to me (and therefore only one disposable client and worker class makes sense to me).

src/Temporalio/Bridge/SafeUnmanagedHandle.cs
cretz: This makes sense to me, I wonder if we can open an issue to track that

cretz (approved):
Approving with the assumption that we will come back and 1) clean up the idea of duplicate client and worker handles pointing to the same pointer, and 2) reuse common pattern for "safe unmanaged handle" approaches instead of having the two separate ones.

src/Temporalio/Bridge/SafeUnmanagedHandle.cs
jmaeagle99: #586

src/Temporalio/Bridge/Client.cs
jmaeagle99: #587

PR #589, opened 2026-01-07
Nexus cancellation types

### LINKED ISSUE #514: [Feature Request] Nexus cancellation type support

cretz:
## What was changed

Support for Nexus cancellation types

## Checklist

1. Closes #514

### DIFF
--- a/src/Temporalio/Workflows/NexusOperationCancellationType.cs
+++ b/src/Temporalio/Workflows/NexusOperationCancellationType.cs
@@ -1,0 +1,33 @@
+namespace Temporalio.Workflows
+{
+    /// <summary>
+    /// How a Nexus operation cancellation is treated by the workflow.
+    /// </summary>
+    public enum NexusOperationCancellationType
+    {
+        /// <summary>
+        /// Wait for operation cancellation completion.
+        /// </summary>
+        WaitCancellationCompleted = Bridge.Api.Nexus.NexusOperationCancellationType.WaitCancellationCompleted,
+
+        /// <summary>
+        /// Do not request cancellation of the nexus operation if already scheduled.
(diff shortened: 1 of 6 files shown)


cretz:
Note, the current CI failure of the `Temporalio.Tests.Client.WorkflowHandleTests.GetResultAsync_Timeout_Throws` test is unrelated to the changes here except that we upgraded dev server to latest. The PR can be reviewed regardless, internal discussions are ongoing w/ server team on changes affecting this other test.

src/Temporalio/Workflows/NexusOperationCancellationType.cs
```diff
@@ -0,0 +1,33 @@
+namespace Temporalio.Workflows
+{
+    /// <summary>
+    /// How a Nexus operation cancellation is treated by the workflow.
+    /// </summary>
+    public enum NexusOperationCancellationType
+    {
+        /// <summary>
+        /// Wait for operation cancellation completion.
```
Quinn-With-Two-Ns:
```suggestion
        /// Wait for operation completion.
```

There is no guarantee the operation will actually return as cancelled, it may return as successful for example

Quinn-With-Two-Ns (approved):

src/Temporalio/Workflows/NexusOperationCancellationType.cs
cretz: :+1: Will update (docs copied from core protos)

cretz: Done

jmaeagle99 (approved):

PR #591, opened 2026-01-12
Include Temporal icon in NuGet packages

jmaeagle99:
## What was changed

Include Temporal icon in NuGet packages

## Why?

Better branding and package recognition


src/Directory.Build.props
cretz, line 1: Personal preference is to put this in the root `.props` file, but would also be ok with this as is, or maybe even better, moving that root one here (and making a smaller copy of it in `tests/`)

cretz (approved):
Nothing blocking

PR #595, opened 2026-01-15
Add support for more timeouts to Nexus operations 

Quinn-With-Two-Ns:
Add schedule-to-start and start-to-close for Nexus operations

NOTE, before this can be merged the API still needs to be tagged, core changes need to be merged and tests need to be added

---

> [!NOTE]
> **Medium Risk**
> Touches workflow command construction for Nexus operations; incorrect mapping could change timeout behavior or break compatibility with older servers, though coverage is added via new/updated timeout tests.
> 
> **Overview**
> Adds `ScheduleToStartTimeout` and `StartToCloseTimeout` to `NexusOperationOptions` and wires them through `WorkflowInstance.StartNexusOperationAsync` so Nexus operation scheduling can enforce these server-side timeouts (in addition to the existing schedule-to-close).
> 
> Extends Nexus worker tests to assert the correct `TimeoutType` surfaces for schedule-to-close, schedule-to-start, and start-to-close failures, and bumps the local test dev-server download version to pick up the required server support.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 94ed30aca651d0830e4c83507a12eaf129197b1c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/Temporalio/Workflows/NexusOperationOptions.cs
+++ b/src/Temporalio/Workflows/NexusOperationOptions.cs
@@ -11,28 +11,28 @@ namespace Temporalio.Workflows
     {
         /// <summary>
         /// Gets or sets the schedule to close timeout.
+        /// </summary>
         /// <remarks>Indicates how long the caller is willing to wait for operation completion.
         /// Calls are retried internally by the server.</remarks>
-        /// </summary>
         public TimeSpan? ScheduleToCloseTimeout { get; set; }
 
         /// <summary>
         /// Gets or sets the schedule to start timeout.
+        /// </summary>
         /// <remarks>Indicates how long the caller is willing to wait for the operation to be started (or completed if synchronous)
         /// by the handler. If the operation is not started within this timeout, it will fail with TIMEOUT_TYPE_SCHEDULE_TO_START.
         /// If not set or zero, no schedule-to-start timeout is enforced.
         /// Requires server version 1.31.0 or later.</remarks>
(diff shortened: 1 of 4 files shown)


src/Temporalio/Workflows/NexusOperationOptions.cs
```diff
@@ -11,9 +11,30 @@ public class NexusOperationOptions : ICloneable
     {
         /// <summary>
         /// Gets or sets the schedule to close timeout.
+        /// Indicates how long the caller is willing to wait for operation completion.
+        /// Calls are retried internally by the server.
```
cretz: Pedantic, but can we move all but the first line of these three docs into a `<remarks>` section instead?

Quinn-With-Two-Ns: Done, let me know if it looks right

jmaeagle99: Remarks should not be nested within summary but be a sibling.

Quinn-With-Two-Ns: Ah thanks, fixed

cretz (approved):

cretz:
LGTM. Merged main and will merge after CI (may be tomorrow)

PR #600, opened 2026-01-22
Fix TemporalWorker::ExecuteAsync(Func<Task>) swallowing exceptions

### LINKED ISSUE #599: [Bug] TemporalWorker::ExecuteAsync(Func<Task>) swallows exceptions, throws TaskCanceledException instead
### Describe the bug
The non-generic `ExecuteAsync(Func<Task>)` overload loses exceptions thrown by the user's task and instead throws `TaskCanceledException`. This only affects workflows that return `Task (void)`, not `Task<T>`.

### Minimal Reproduction
```csharp
[Workflow]
public class VoidWorkflow
{
    [WorkflowRun]
    public async Task RunAsync()  // Returns Task, not Task<T>
    {
        throw new ApplicationFailureException("This exception is lost", nonRetryable: true);
    }
}

[Fact]
public async Task ExceptionIsLost()
{
    await using var env = await WorkflowEnvironment.StartLocalAsync();
    using var worker = new TemporalWorker(
        env.Client,
        new TemporalWorkerOptions($"tq-{Guid.NewGuid()}").AddWorkflow<VoidWorkflow>());

    // This throws TaskCanceledException instead of WorkflowFailedException
    var ex = await Assert.ThrowsAsync<WorkflowFailedException>(async () =>
        await worker.ExecuteAsync(async () =>
            await env.Client.ExecuteWorkflowAsync(
                (VoidWorkflow wf) => wf.RunAsync(),
                new($"wf-{Guid.NewGuid()}", worker.Options.TaskQueue!))));
  }
```

#### Expected behavior
WorkflowFailedException is thrown (wrapping the ApplicationFailureException).

#### Actual behavior
TaskCanceledException: A task was canceled. is thrown. The original exception is lost.

#### Workaround
Use the generic overload by returning a value:

```csharp
await worker.ExecuteAsync(async () =>
{
    await client.ExecuteWorkflowAsync(...);
    return ValueTuple.Create();
});
```

### Environment/Versions
I've tracked this down to #548, which was first released in https://github.com/temporalio/sdk-dotnet/releases/tag/1.10.0

### Additional context
The ExecuteAsync(Func<Task>) overload uses ContinueWith with OnlyOnRanToCompletion:

```csharp
public Task ExecuteAsync(
    Func<Task> untilComplete, CancellationToken stoppingToken = default) =>
    ExecuteInternalAsync(
        () => untilComplete().ContinueWith(
            _ => ValueTuple.Create(),
            default,
            TaskContinuationOptions.ExecuteSynchronously | TaskContinuationOptions.OnlyOnRanToCompletion,
            TaskScheduler.Current),
        stoppingToken);
```

When untilComplete() throws, OnlyOnRanToCompletion prevents the continuation from running, and the resulting task is canceled (not faulted). The original exception is discarded.

adeal:
## What was changed
- Fix `TemporalWorker::ExecuteAsync(Func<Task>)` swallowing exceptions

## Why?
The non-generic `ExecuteAsync(Func<Task>)` overload was using `ContinueWith` with `TaskContinuationOptions.OnlyOnRanToCompletion`. When the caller's task threw an exception, this option prevented the continuation from running, causing the continuation task to be canceled rather than faulted. This resulted in `TaskCanceledException` being thrown instead of the original exception (e.g., `WorkflowFailedException`).

## Checklist

1. Closes #599 

2. How was this tested:
Added regression test `ExecuteAsync_PropagatesException`.

3. Any docs updates needed?
No.

### DIFF
--- a/src/Temporalio/Worker/TemporalWorker.cs
+++ b/src/Temporalio/Worker/TemporalWorker.cs
@@ -260,11 +260,11 @@ namespace Temporalio.Worker
         public Task ExecuteAsync(
             Func<Task> untilComplete, CancellationToken stoppingToken = default) =>
             ExecuteInternalAsync(
-                () => untilComplete().ContinueWith(
-                    _ => ValueTuple.Create(),
-                    default,
-                    TaskContinuationOptions.ExecuteSynchronously | TaskContinuationOptions.OnlyOnRanToCompletion,
-                    TaskScheduler.Current),
+                async () =>
+                {
+                    await untilComplete().ConfigureAwait(false);
+                    return ValueTuple.Create();
+                },
                 stoppingToken);
 
         /// <summary>
--- a/tests/Temporalio.Tests/Testing/WorkflowEnvironmentTests.cs
+++ b/tests/Temporalio.Tests/Testing/WorkflowEnvironmentTests.cs
@@ -50,6 +50,17 @@ public class WorkflowEnvironmentTests : TestBase
         public Task SomeSignalAsync() => Task.CompletedTask;
     }
 
+    [Workflow]
+    public class VoidWorkflowThatFails
+    {
+        [WorkflowRun]
+        public async Task RunAsync()
+        {
+            await Task.Yield();
+            throw new ApplicationFailureException("Intentional failure", nonRetryable: true);
+        }
+    }
+
     [OnlyIntelFact]
     public async Task StartTimeSkippingAsync_SlowWorkflowAutoSkip_ProperlySkips()
     {
@@ -289,5 +300,28 @@ public class WorkflowEnvironmentTests : TestBase
                 TypedSearchAttributes = attrVals,
             });
         Assert.Equal(attrVals, (await handle.DescribeAsync()).TypedSearchAttributes);
+    }
+
+    [Fact]
+    public async Task ExecuteAsync_VoidWorkflowFailure_PropagatesException()
+    {
+        // This test verifies that ExecuteAsync(Func<Task>) properly propagates exceptions
+        // from void-returning workflows. This is a regression test for a bug where
+        // ContinueWith with OnlyOnRanToCompletion caused exceptions to be swallowed.
+        string queue = $"tq-{Guid.NewGuid()}";
+        await using var env = await WorkflowEnvironment.StartLocalAsync();
+        using var worker = new TemporalWorker(
+            env.Client,
+            new TemporalWorkerOptions(queue).AddWorkflow<VoidWorkflowThatFails>());
+
+        var exc = await Assert.ThrowsAsync<WorkflowFailedException>(async () =>
+            await worker.ExecuteAsync(async () =>
+                await env.Client.ExecuteWorkflowAsync(
+                    (VoidWorkflowThatFails wf) => wf.RunAsync(),
+                    new(id: $"wf-{Guid.NewGuid()}", taskQueue: queue))));
+
+        Assert.NotNull(exc.InnerException);
+        var appExc = Assert.IsType<ApplicationFailureException>(exc.InnerException);
+        Assert.Equal("Intentional failure", appExc.Message);
     }
 }
--- a/tests/Temporalio.Tests/Testing/WorkflowEnvironmentTests.cs
+++ b/tests/Temporalio.Tests/Testing/WorkflowEnvironmentTests.cs
@@ -50,17 +50,6 @@ public class WorkflowEnvironmentTests : TestBase
         public Task SomeSignalAsync() => Task.CompletedTask;
     }
 
-    [Workflow]
-    public class VoidWorkflowThatFails
-    {
-        [WorkflowRun]
-        public async Task RunAsync()
-        {
-            await Task.Yield();
-            throw new ApplicationFailureException("Intentional failure", nonRetryable: true);
-        }
-    }
-
     [OnlyIntelFact]
     public async Task StartTimeSkippingAsync_SlowWorkflowAutoSkip_ProperlySkips()
     {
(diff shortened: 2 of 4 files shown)


tests/Temporalio.Tests/Testing/WorkflowEnvironmentTests.cs
```diff
@@ -290,4 +301,27 @@ public async Task StartLocal_SearchAttributes_ProperlyRegistered()
             });
         Assert.Equal(attrVals, (await handle.DescribeAsync()).TypedSearchAttributes);
     }
+
+    [Fact]
+    public async Task ExecuteAsync_VoidWorkflowFailure_PropagatesException()
```
adeal: Please let me know if there is a better location for this test.

tests/Temporalio.Tests/Testing/WorkflowEnvironmentTests.cs
```diff
@@ -290,4 +301,27 @@ public async Task StartLocal_SearchAttributes_ProperlyRegistered()
...
+        // ContinueWith with OnlyOnRanToCompletion caused exceptions to be swallowed.
+        string queue = $"tq-{Guid.NewGuid()}";
+        await using var env = await WorkflowEnvironment.StartLocalAsync();
+        using var worker = new TemporalWorker(
+            env.Client,
+            new TemporalWorkerOptions(queue).AddWorkflow<VoidWorkflowThatFails>());
+
+        var exc = await Assert.ThrowsAsync<WorkflowFailedException>(async () =>
+            await worker.ExecuteAsync(async () =>
+                await env.Client.ExecuteWorkflowAsync(
```
cretz: Pedantic, but this doesn't seem related to workflows at all, can you just throw an exception on this line and not use a client/workflow in here?

tests/Temporalio.Tests/Testing/WorkflowEnvironmentTests.cs
cretz: Yeah, we don't really have a place for general worker tests, am ok leaving it in here (or making new `Worker/GeneralWorkerTests.cs` or putting it in `ActivityWorkerTests` or `WorkflowWorkerTests` though doesn't relate to those any better than it does to this one).

src/Temporalio/Worker/TemporalWorker.cs
```diff
@@ -260,11 +260,11 @@ public Task ExecuteAsync(CancellationToken stoppingToken) =>
         public Task ExecuteAsync(
             Func<Task> untilComplete, CancellationToken stoppingToken = default) =>
             ExecuteInternalAsync(
-                () => untilComplete().ContinueWith(
```
cretz, line 263: Ah yeah, it is confusing that .NET doesn't really have a "map-on-success" combinator and that continue-with inadvertently swallows failures here. Thanks for the fix!

I see one other place we'll need to fix this type of thinking (`Temporalio.Workflows.Semaphore` for a failure in a wait condition), but we can handle that in a separate issue, no worries.

cretz (commented):
Thanks! Just one minor pedantic suggestion to simplify the test.

And don't forget to sign the CLA (link is above here in PR)

tests/Temporalio.Tests/Testing/WorkflowEnvironmentTests.cs
adeal: You're absolutely right. I was too focused on how it surfaced in my projects, which was workflow tests like this. I'll get the PR/issue description updated.

tests/Temporalio.Tests/Testing/WorkflowEnvironmentTests.cs
adeal: I created `Worker/GeneralWorkerTests` and moved there.

tests/Temporalio.Tests/Testing/WorkflowEnvironmentTests.cs
adeal: Updated the following to reflect this
1. PR description
2. Unit test no longer has dependency on executing a workflow

tests/Temporalio.Tests/Worker/GeneralWorkerTests.cs
```diff
@@ -0,0 +1,39 @@
...
+        }
+    }
+
+    [Fact]
+    public async Task ExecuteAsync_VoidWorkflowFailure_PropagatesException()
+    {
+        // This test verifies that ExecuteAsync(Func<Task>) properly propagates exceptions
+        // from void-returning tasks. This is a regression test for a bug where
+        // ContinueWith with OnlyOnRanToCompletion caused exceptions to be swallowed.
+        await using var env = await Temporalio.Testing.WorkflowEnvironment.StartLocalAsync();
```
cretz: We shouldn't use a new start-local here, can we alter this class to extend `WorkflowEnvironmentTestBase` similar to others needing this environment so it can be reused?

tests/Temporalio.Tests/Worker/GeneralWorkerTests.cs
```diff
@@ -0,0 +1,39 @@
...
+        using var worker = new TemporalWorker(
+            env.Client,
+            new TemporalWorkerOptions($"tq-{Guid.NewGuid()}").AddWorkflow<VoidWorkflow>());
```
cretz: Can get rid of the workflow class if you do something like this, but no big deal:

```suggestion
        [Activity]
        static void Ignored() => throw new NotImplementedException();
        using var worker = new TemporalWorker(
            env.Client,
            new TemporalWorkerOptions($"tq-{Guid.NewGuid()}").AddActivity(Ignored));
```

cretz (commented):
One change needed, to use the existing shared environment instead of making a new one for just this test

tests/Temporalio.Tests/Worker/GeneralWorkerTests.cs
adeal: Good call, I like that better. Thanks for the tip.

tests/Temporalio.Tests/Worker/GeneralWorkerTests.cs
adeal: Good catch, thank you.

Updated in https://github.com/temporalio/sdk-dotnet/pull/600/commits/45ed3edcec6fb6cc67a3fdc922f53bd08c08c257.

tests/Temporalio.Tests/Worker/GeneralWorkerTests.cs
adeal: Updated in https://github.com/temporalio/sdk-dotnet/pull/600/commits/45ed3edcec6fb6cc67a3fdc922f53bd08c08c257.

cretz (approved):

cretz:
@adeal - looks like some whitespace formatting failing in CI due to inadvertent commit to `Semaphore.cs`, can you remove that change? I opened #601 separately.

PR #602, opened 2026-01-23
Avoid double Client SafeHandle implementation

### LINKED ISSUE #587: Refactor Bridge.Client to be IDisposable instead of a SafeHandle
#583 introduced the separation of unmanaged pointer lifetime management into `SafeHandle`. The `Bridge.Client` class could not be fully converted to a disposable implementation without other possible API changes; this work was deferred to fix the immediate issue that the PR was addressing. This conversion should be done to match the pattern established for the `Bridge.Worker` class so that there aren't two SafeHandle implementations for `Interop.TemporalCoreClient*`.

jmaeagle99:
## What was changed

Changed the bridge client to no longer implement `SafeHandle` and to only implement `IDisposable`. This required that more information is pass via a new `IBridgeClientProviderInternal` interface that exposes the `SafeClientHandle` and the `Runtime` instances from the `TemporalConnection`.

## Why?

Avoid the double creation of `SafeHandle` implementations for bridge clients.

## Checklist
1. Closes #587 
1. How was this tested: Unit tests

### DIFF
--- a/src/Temporalio/Bridge/Client.cs
+++ b/src/Temporalio/Bridge/Client.cs
@@ -1,6 +1,5 @@
 using System;
 using System.Collections.Generic;
-using System.Runtime.InteropServices;
 using System.Threading.Tasks;
 using Google.Protobuf;
 
@@ -9,17 +8,15 @@ namespace Temporalio.Bridge
     /// <summary>
     /// Core-owned Temporal client.
     /// </summary>
-    internal class Client : SafeHandle
+    internal class Client : IDisposable
     {
+        private readonly SafeClientHandle handle;
+
         private unsafe Client(Runtime runtime, Interop.TemporalCoreClient* ptr)
-            : base((IntPtr)ptr, true)
         {
             Runtime = runtime;
-            Handle = new SafeClientHandle(ptr);
+            handle = new SafeClientHandle(ptr);
         }
-
-        /// <inheritdoc />
-        public override bool IsInvalid => Handle.IsInvalid;
 
         /// <summary>
         /// Gets the runtime associated with this client.
@@ -29,7 +26,7 @@ namespace Temporalio.Bridge
         /// <summary>
         /// Gets the safe handle for the client.
         /// </summary>
-        internal SafeClientHandle Handle { get; private init; }
+        internal SafeClientHandle Handle => handle;
 
         /// <summary>
         /// Connect to Temporal.
@@ -211,10 +208,9 @@ namespace Temporalio.Bridge
         }
 
         /// <inheritdoc />
-        protected override unsafe bool ReleaseHandle()
+        public void Dispose()
         {
-            Handle.Dispose();
-            return true;
+            handle.Dispose();
         }
     }
 }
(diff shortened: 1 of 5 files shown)


Sushisource (approved):
Guess this one fell off the radar somehow huh

src/Temporalio/Bridge/Client.cs
cretz, line 1: Can we just remove this class and only have `SafeClientHandle`? Same for the `Worker` one. I can't really see a benefit of having this thin layer on top.

src/Temporalio/Client/IBridgeClientProviderInternal.cs
cretz, line 1: I am a bit confused on the need for this _and_ `IBridgeClientProvider`. Is this duplicating that one's purpose? What was wrong with what was there and the cast?

jmaeagle99: As I am breaking apart the safe handles from their wrapper classes, I'm also passing around only the safe handle and not the wrapper classes anymore.

So if you look at the bridge `Worker` constructor, I now require passing the `SafeClientHandle` and the `Runtime` instead of the `Client` instance. This requires that I'm able to get the `Runtime` from the bridge provider somehow; but now that `Client` doesn't implement `SafeHandle` anymore, I can't just cast `IBridgeClientProvider.BridgeClient` to `Client`.

I have to either update that public interface (which might be okay, if that's not meant to be implemented by anyone outside of the SDK) or introduce a new internal interface that we can put whatever we want on it.

src/Temporalio/Bridge/Client.cs
jmaeagle99: Possibly. Their methods would have to go _somewhere_ (not the `Safe*Handle`, because that's the whole point of the separation).

cretz: Why can't the methods be on the safe handle? Is the separation beneficial? It's not truly wrapping it if you're still accessing and using the safe handle elsewhere.

src/Temporalio/Client/IBridgeClientProviderInternal.cs
cretz: Why not pass around the wrapper? The reason we made `ITemporalConnection` be an `IBridgeClientProvider` and made that public is so if someone wrapped/implemented the connection, they could return the underlying one's `BridgeClient`. And that can still be used for creating workers. It was all `public` by intention. But at first glance, it looks like this is a breaking change because now a user's `IWorkerClient` that might have returned their `ITemporalConnection` (which implements `IBridgeClientProvider` by maybe delegating to the underlying one) that used to work for creating a worker can no longer work. I admit it is a bit of a contrived use case, but we did make it public on purpose.

> I have to either update that public interface

Why? The worker needs the `SafeHandle`, why can't it continue to provide it? Isn't that what the internal one is doing, just providing the safe handle? Also, arguably the runtime should be on the safe handle for this use case.

Maybe I don't quite get the value of the separation/wrapper vs just having one safe client handle that does everything a bridge client needs to do.

PR #608, opened 2026-02-11
Update sdk-core submodule to latest

jmaeagle99:
## What was changed

Update sdk-core submodule to latest.

## Why?

Get latest APIs and fixes from sdk-core.

## Checklist
1. How was this tested: existing tests


src/Temporalio/Bridge/Cargo.lock
```diff
@@ -1,4003 +0,0 @@
-# This file is automatically @generated by Cargo.
```
jmaeagle99, line 1: I believe that this was unintentionally left over from when the C bridge was moved to sdk-core. Removed it to avoid false positives on scans.

cretz: Looks like it was accidentally added in #509, agreed should be removed if there's no accompanying Cargo.toml

cretz (approved):

PR #609, opened 2026-02-13
💥 Standalone activity support

### LINKED ISSUE #568: [Feature Request] Support standalone activities

Allow executing activities without a workflow. The clients must support the new (upcoming) server APIs for starting, managing and getting results of activities. The workers must support running activities outside of workflow context.

### Additional context

Server API PR (WIP): https://github.com/temporalio/api/pull/640

cretz:
## What was changed

- Added `StartActivityOptions`, `ActivityDescribeOptions`, `ActivityCancelOptions`, `ActivityTerminateOptions`, `ActivityListOptions`, `ActivityListPaginatedOptions`, `ActivityCountOptions`                                                                                                 
- Added `ActivityHandle` and `ActivityHandle<TResult>` with `GetResultAsync`, `DescribeAsync`, `CancelAsync`, `TerminateAsync`
- Added `ActivityExecution`, `ActivityExecutionDescription`, `ActivityExecutionCount`, `ActivityListPage` model classes
- Added `ActivityFailedException` and `ActivityAlreadyStartedException` exceptions
- Added `StartActivityAsync`, `GetActivityHandle`, `ListActivitiesAsync`, `CountActivitiesAsync`, `ListActivitiesPaginatedAsync` on `ITemporalClient`  
- Added `ExecuteActivityAsync` extension methods on `ITemporalClient` (start + get result shortcut, 4 overloads matching workflow pattern)
- Added interceptor inputs: `StartActivityInput`, `DescribeActivityInput`, `CancelActivityInput`, `TerminateActivityInput`, `ListActivitiesInput`, `ListActivitiesPaginatedInput`, `CountActivitiesInput`
- Added activity methods on `ClientOutboundInterceptor`
- Added timeout validation requiring `ScheduleToCloseTimeout` or `StartToCloseTimeout` on `StartActivityOptions`
- Updated `ISerializationContext.Activity` to support standalone activities: `ActivityId` added, `WorkflowId`/`WorkflowType` now nullable, some fields deprecated
  - 💥 BACKWARDS INCOMPATIBLE CHANGE - changing some items to nullable here, though only nullable in standalone situations. Also deprecated some items.
- Updated `ActivityInfo` with `Namespace`, `IsWorkflowActivity`, and nullable workflow fields (`WorkflowId`, `WorkflowNamespace`, `WorkflowRunId`, `WorkflowType`)
  - 💥 BACKWARDS INCOMPATIBLE CHANGE - changing some items to nullable here, though only nullable in standalone situations
- Fixed `NullReferenceException` in `ActivityWorker.StartActivity` when `CurrentAttemptScheduledTime` is null for standalone activities
- Added error handling in activity worker poll loop so `StartActivity` failures complete the task as failed instead of crashing the worker
- Updated dev server to `v1.6.1-server-1.31.0-150.0` with standalone activity dynamic config flags
- Added 12 integration tests covering start/execute, already-started/reuse-policy errors, get handle, describe, cancel, terminate, list/count/paginate, interceptors, and worker activity info

## Checklist

1. Closes #568


.github/workflows/ci.yml
```diff
@@ -149,3 +149,4 @@ jobs:
       dotnet-repo-path: ${{github.event.pull_request.head.repo.full_name}}
       version: ${{github.event.pull_request.head.ref}}
       version-is-repo-ref: true
+      features-repo-ref: standalone-activities-dotnet
```
cretz, line 152: Note, we will need to merge this branch into `features` repo right after this PR is merged.

src/Temporalio/Client/StartActivityOptions.cs
```diff
@@ -0,0 +1,125 @@
...
+using Temporalio.Common;
+
+namespace Temporalio.Client
+{
+    /// <summary>
+    /// Options for starting a standalone activity. <see cref="Id" /> and <see cref="TaskQueue" />
+    /// are required.
+    /// </summary>
+    /// <remarks>WARNING: Standalone activities are experimental.</remarks>
+    public class StartActivityOptions : ICloneable
```
cretz, line 12: Note, after internal discussions, we chose `StartActivityOptions` to not clash with existing `ActivityOptions` which is kinda a uniquely .NET problem of preventing clashes across the entire assembly (instead of just namespace)

src/Temporalio/Activities/ActivityInfo.cs
```diff
@@ -74,10 +86,10 @@ public record ActivityInfo(
             ["ActivityId"] = ActivityId,
             ["ActivityType"] = ActivityType,
             ["Attempt"] = Attempt,
-            ["WorkflowNamespace"] = WorkflowNamespace,
-            ["WorkflowId"] = WorkflowId,
-            ["WorkflowRunId"] = WorkflowRunId,
-            ["WorkflowType"] = WorkflowType,
+            ["Namespace"] = Namespace,
```
jmaeagle99: Should we maybe preserve WorkflowNamespace and only fill it if it's workflow activity? That way we don't wholesale break anyone who is scraping logs, especially for those who aren't using this experimental feature yet.

src/Temporalio/Client/ActivityExecution.cs
```diff
@@ -0,0 +1,144 @@
...
+        /// <param name="rawInfo">Raw proto list info.</param>
+        /// <param name="clientNamespace">Client namespace.</param>
+        /// <remarks>WARNING: This constructor may be mutated in backwards incompatible ways.</remarks>
+        protected internal ActivityExecution(
+            ActivityExecutionListInfo rawInfo, string clientNamespace)
+            : this(
+                clientNamespace: clientNamespace,
+                activityId: rawInfo.ActivityId,
+                activityRunId: string.IsNullOrEmpty(rawInfo.RunId) ? null : rawInfo.RunId,
+                activityType: rawInfo.ActivityType?.Name ?? string.Empty,
```
jmaeagle99, line 29: If this value is null under valid scenarios, then should we make activity type null? Or is this realistically not null and only possibly null because `rawInfo.ActivityType` is a proto object?

src/Temporalio/Client/ActivityExecution.cs
```diff
@@ -0,0 +1,144 @@
...
+                closeTime: rawInfo.CloseTime?.ToDateTime(),
+                executionDuration: rawInfo.ExecutionDuration?.ToTimeSpan(),
+                scheduledTime: rawInfo.ScheduleTime?.ToDateTime() ?? default,
```
jmaeagle99, line 32: Same question as https://github.com/temporalio/sdk-dotnet/pull/609/changes#r2814999948

src/Temporalio/Activities/ActivityInfo.cs
cretz: We can, though the question is whether we should _also_ emit a `Namespace` at that time? Is it worth the redundant scope?

src/Temporalio/Client/ActivityExecution.cs
cretz: Realistically never null and only being extra safe because it's an object in proto

src/Temporalio/Client/ActivityExecution.cs
cretz: Yeah, same here, it will never be null from server

src/Temporalio/Activities/ActivityInfo.cs
jmaeagle99: Thinking this through...

Previously, `WorkflowNamespace` was always added as a scope value and always had a value. Log scraping/queries who look for it a likely relying on it being there. With this change as-is, that key will no longer be there at all. So they would have to adjust their code to find the new kew.

With my suggestion, `WorkflowNamespace` would still be provided under those circumstances IF they don't start using standalone activities. Once they start using standalone activities, some of their log entries won't have `WorkflowNamespace` (for those that are standalone activities). This would avoid a break until they start using the experimental feature.

So questions are:
- do we break now while the feature is still experimental and probably not used in environments where log scraping is occurring? Maybe we can defer the break until standalone activities is GA?
- do we break only for standalone activities?

cretz: Right, all of that makes sense, and I don't have a strong opinion on whether to break existing log scope keys now, later, or not at all. It makes sense for me to put `WorkflowNamespace` back for now.

What I was asking was a followup question - do we emit `Namespace` always or only for standalone activities?

jmaeagle99: I think we always emit `Namespace` and say that we're deprecating `WorkflowNamespace`.

cretz: Do you think we should deprecate with intent to remove one day, or never remove? Is there a meaningful difference removing later vs now (many do not heed deprecations that are release-notes only)? What do we say to users that don't want duplicate `Namespace` _and_ `WorkflowNamespace` tags in their logs?

jmaeagle99: tl;dr: If customers are only supported on the latest version, then I don't think it matters whether the break is made now or later because stating "stay on the prior version until you adjust" is not valid. If that is the case, and this break would likely be unnoticed without specific use or testing, then just break it now and call it out in release notes. This _can_ reduce the likelihood that we need to call out breaking changes for yet another release.

> Do you think we should deprecate with intent to remove one day, or never remove?

My default position is that deprecation shows intent to remove at a later date. If you deprecate and don't remove at some point, that's just additional debt being carried, no matter how small it may be.

> Is there a meaningful difference removing later vs now (many do not heed deprecations that are release-notes only)?

If someone reads the release notes and ignores something that has been called out, they made the decision to not adjust to that information given. If someone doesn't read the release notes at all, then... hopefully they catch any changes that are meaningful to them in testing.

The meaningful difference of removing later is that customers are given time to adjust. The argument against that is customers can just not upgrade immediately. I think that's fair so long as prior versions are not immediately out of support upon publishing a new version. If not given that time to adjust while being supported, its difficult to make immediate adjustments with the risk of unknown breakages. Especially if those breakages are coming in for an experimental feature that a customer wasn't intending to use immediately.

Another aspect that I think about is how obvious is it that there are changes for existing customers. I suspect that the changes on `ISerializationContext.Activity` and `ActivityInfo`, while the can be impactful (because they're likely to be a source breaking change) are not _broadly_ impactful; I could be wrong. If that is true, the the remaining breaking change (the removal of `WorkflowNamespace` scope property) is likely to sneak through unnoticed until it hits a staging environment. I think that may happen just as often regardless if the property is removed now or later.

With all that, I think it's better to break it now and not have to call out yet another breaking change in a future version.

> What do we say to users that don't want duplicate Namespace and WorkflowNamespace tags in their logs?

I think that's relatively the same as "what do we say to users who don't use all of the extra scope properties that they don't ever intend to use" and my default there would be "They should ignore those extra scope properties". Another property of a small size isn't going to break their storage (it shouldn't, hopefully).

jmaeagle99 (approved):

src/Temporalio/Activities/ActivityInfo.cs
cretz: :+1: I will add the redundant log tag for the majority of users

cretz: Meaning all workflow logs will have `Namespace` and `WorkflowNamespace`, but all new standalone activity logs will have just `Namespace`, assuming that is what you meant.

jmaeagle99: I'm fine with theremoval of `WorkflowNamespace` now as you have in this PR. If we are ever going to remove it, I think now is better than later.

cretz: :+1: Same for other workflow tags. Can you take a look at https://github.com/temporalio/sdk-dotnet/pull/609/changes/287ed97c0e123737bafb2dda1e9db4f43cfd32d0 and confirm it is what you were thinking?

jmaeagle99 (approved):

src/Temporalio/Activities/ActivityInfo.cs
jmaeagle99: lgtm :shipit:

PR #611, opened 2026-02-17
Fix local activity backoff double-encoding headers when codec is used

### LINKED ISSUE #610: [Bug] re-attempt of LocalActivity after a LocalActivityBackoff - leading to task failure

Using encryption and some headers to propagate context.
I have a failing LocalActivity - with a reasonably back off > Workflow Task 
- This results in Temporal durable Timer for LocalActivityBackoff

One the Timer fires, I expect the Workflow task and LocalActivity complete successfully 

### Describe the bug

After the Timer fires, on WorkflowTask started- it results in attempt to run LocalActivity
- however the payload converter throws the below error

Image

### Minimal Reproduction
https://github.com/gauravthadani/samples-dotnet/tree/main/src/Encryption

### Environment/Versions

- OS and processor: [e.g. M1 Mac, x86 Windows, Linux]
- Temporal Version: [e.g. 1.14.0?] and/or SDK version
- Are you using Docker or Kubernetes or building Temporal from source?

cretz:
## What was changed

Clone headers for local activities so codec mutating them doesn't cause problems with repeat use

## Checklist

1. Closes #610

### DIFF
--- a/src/Temporalio/Worker/WorkflowInstance.cs
+++ b/src/Temporalio/Worker/WorkflowInstance.cs
@@ -2229,7 +2229,14 @@ namespace Temporalio.Worker
                         };
                         if (input.Headers is IDictionary<string, Payload> headers)
                         {
-                            cmd.Headers.Add(headers);
+                            // Clone each header payload to avoid in-place codec
+                            // encoding from mutating the originals, which would
+                            // cause double-encoding on local activity backoff
+                            // retries.
+                            foreach (var kvp in headers)
+                            {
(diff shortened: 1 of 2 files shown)


jmaeagle99:
Does this only impact local activities? I see other places in this file that consume headers similarly and am wondering if they are impacted as well.

cretz:
Only local activities (because only they reuse the same headers for successive calls w/ the Core do-backoff thing)

jmaeagle99 (approved):

PR #619, opened 2026-02-26
 💥 [BREAKING CHANGE] Rename Nexus Operation classes used in a workflow in preparation for Stand alone Nexus operations

### LINKED ISSUE #618: [Feature Request] Rename Nexus Operation classes used in a workflow in preparation for Stand alone Nexus operations

When doing design prep for Stand alone Nexus operations we found that lots of classes will have name collisions with the existing workflow based Nexus Operation classes.

Rename the workflow based Nexus Operation classes pre-emtively to avoid any conflict in the future. The current thought is to add `workflow` so `NexusClient` becomes `NexusWorkflowClient`

Quinn-With-Two-Ns:
Rename Nexus Operation classes used in a workflow. I went with inserting `Workflow` into the name, but am open to other suggestions.

closes https://github.com/temporalio/sdk-dotnet/issues/618

---

> [!NOTE]
> **Medium Risk**
> Breaking public API rename across workflow Nexus client/handle/options and related interceptor signatures; functional behavior appears unchanged but downstream projects will require code updates and any missed rename could surface as runtime/compile failures.
> 
> **Overview**
> Renames the workflow-scoped Nexus API surface to explicitly include `Workflow` in the type names, updating `Workflow.CreateNexusClient` to `Workflow.CreateNexusWorkflowClient` and renaming `NexusClient*`, `NexusOperationHandle*`, and related options to `NexusWorkflowClient*`, `NexusWorkflowOperationHandle*`, and `NexusWorkflow*Options`.
> 
> Updates all internal plumbing and interception points (workflow outbound interceptors, OpenTelemetry `TracingInterceptor`, and `WorkflowInstance` implementations) plus tests and README references to use the new names.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e01b7bdfe447738eda646a6c3428f4d72e752d2b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/Temporalio/Workflows/NexusWorkflowClient.cs
+++ b/src/Temporalio/Workflows/NexusWorkflowClient.cs
@@ -13,7 +13,7 @@ namespace Temporalio.Workflows
     /// Client for making Nexus service calls from a workflow.
     /// </summary>
     /// <remarks>WARNING: Nexus support is experimental.</remarks>
-    public abstract class NexusClient
+    public abstract class NexusWorkflowClient
     {
         /// <summary>
         /// Gets the service name.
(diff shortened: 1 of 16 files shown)


jmaeagle99 (approved):

src/Temporalio/Workflows/NexusWorkflowClient.cs
```diff
@@ -13,7 +13,7 @@ namespace Temporalio.Workflows
     /// Client for making Nexus service calls from a workflow.
     /// </summary>
     /// <remarks>WARNING: Nexus support is experimental.</remarks>
-    public abstract class NexusClient
+    public abstract class NexusWorkflowClient
```
cretz, line 16: Can we add a 💥 to the PR title and a very clear "BACKWARDS INCOMPATIBLE CHANGE" note to the description? Even though this is experimental, we still like to inform users on release when things like this change.

src/Temporalio/Workflows/Workflow.cs
```diff
@@ -317,14 +317,14 @@ public static ContinueAsNewException CreateContinueAsNewException(
...
-        /// shortcut for <see cref="CreateNexusClient(string, NexusClientOptions)"/>.
+        /// shortcut for <see cref="CreateNexusWorkflowClient(string, NexusWorkflowClientOptions)"/>.
         /// </summary>
         /// <param name="service">Service name.</param>
         /// <param name="endpoint">Endpoint.</param>
         /// <returns>Nexus client.</returns>
         /// <remarks>WARNING: Nexus support is experimental.</remarks>
-        public static NexusClient CreateNexusClient(string service, string endpoint) =>
-            CreateNexusClient(service, new NexusClientOptions(endpoint));
+        public static NexusWorkflowClient CreateNexusWorkflowClient(string service, string endpoint) =>
```
cretz, line 326: Arguably the method name here (and in interceptors) should stay the same. It is clearly in the workflow class, there is no ambiguity, and there is not much value in qualifying the method name. (of course return type being `NexusWorkflowClient` makes sense)

Quinn-With-Two-Ns: I think it'd be odd for the method names to not match the type returned, but i'll defer to you and Justin

jmaeagle99: I agree with Quinn on this. And it leaves room in the future if we need to return some other type of Nexus client that isn't a `NexusWorkflowClient`.

PR #627, opened 2026-03-18
Add upgrade-on-CAN support and expose CAN suggestion reasons

THardy98:
## Summary
- Allow PINNED workflows to detect when the target worker deployment version has changed via `Workflow.TargetWorkerDeploymentVersionChanged`
- Add `InitialVersioningBehavior` option on `ContinueAsNewOptions` to support upgrading to the latest version on continue-as-new (AUTO_UPGRADE)
- Expose `SuggestContinueAsNewReason` flags from the server so workflows can inspect why a CAN was suggested
- Add integration test for the full upgrade-on-CAN flow

### DIFF
--- a/src/Temporalio/Worker/WorkflowInstance.cs
+++ b/src/Temporalio/Worker/WorkflowInstance.cs
@@ -267,6 +267,13 @@ namespace Temporalio.Worker
         public int CurrentHistorySize { get; private set; }
 
         /// <inheritdoc />
+        public IReadOnlyCollection<Workflows.SuggestContinueAsNewReason> SuggestedContinueAsNewReasons { get; private set; } =
+            Array.Empty<Workflows.SuggestContinueAsNewReason>();
+
+        /// <inheritdoc />
+        public bool TargetWorkerDeploymentVersionChanged { get; private set; }
+
+        /// <inheritdoc />
         public WorkflowUpdateInfo? CurrentUpdateInfo => CurrentUpdateInfoLocal.Value;
 
         /// <inheritdoc />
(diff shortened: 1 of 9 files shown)


src/Temporalio/Worker/WorkflowInstance.cs
```diff
@@ -627,6 +634,11 @@ public WorkflowActivationCompletion Activate(WorkflowActivation act)
                 ContinueAsNewSuggested = act.ContinueAsNewSuggested;
                 CurrentHistoryLength = checked((int)act.HistoryLength);
                 CurrentHistorySize = checked((int)act.HistorySizeBytes);
+                SuggestedContinueAsNewReasons = act.SuggestContinueAsNewReasons
```
jmaeagle99, line 637: If `WorkflowActivation.SuggestContinueAsNewReasons` contains an enum value that is not in the set of named enums, this is going to wrap it in the SDK representation the same way. For example:

```cs
var serverEnums = new List<Api.Enums.V1.SuggestContinueAsNewReason>()
{
    Api.Enums.V1.SuggestContinueAsNewReason.HistorySizeTooLarge,
    (Api.Enums.V1.SuggestContinueAsNewReason)8, // Not a named enum field
};

var clientEnums = serverEnums
    .Select(r => (Temporalio.Workflows.SuggestContinueAsNewReason)(int)r)
    .Where(r => r != Temporalio.Workflows.SuggestContinueAsNewReason.Unspecified)
    .ToList();

foreach (var value in clientEnums)
{
    Console.WriteLine($"Value: {value}");
}
```

Will print out:
```
Value: HistorySizeTooLarge
Value: 8
```

Is that intended? Should unknown values be filtered out?

src/Temporalio/Worker/WorkflowInstance.cs
```diff
@@ -266,6 +266,13 @@ public WorkflowInstance(WorkflowInstanceDetails details)
         /// <inheritdoc />
         public int CurrentHistorySize { get; private set; }
 
+        /// <inheritdoc />
+        public IReadOnlyCollection<Workflows.SuggestContinueAsNewReason> SuggestedContinueAsNewReasons { get; private set; } =
```
jmaeagle99, line 270: What's the intent for exposing these in the public API? How would a customer actually use these in a meaningful way? I don't see any usage of these in the change other than just making it publicly available.

src/Temporalio/Worker/WorkflowInstance.cs
THardy98: I don't see this as that much of an issue. It's coming from core/server so it's something we own.
I suppose the issue is if we add a new reason but forget to plumb it to lang, but I think in that case, I'd prefer that we list the reason than obfuscate it.

src/Temporalio/Worker/WorkflowInstance.cs
THardy98: For context, this [thread](https://temporaltechnologies.slack.com/archives/C091S4EUZPT/p1773858186955009?thread_ts=1773849990.678549&cid=C091S4EUZPT)

TLDR: additional visibility

jmaeagle99 (approved):

PR #629, opened 2026-03-20
💥 Nexus - Fix operation token on NexusWorkflowRunHandle

Quinn-With-Two-Ns:
Fix operation token on NexusWorkflowRunHandle. Code had an obvious typo and more subtle url encoding issue. I also added some regression tests that test cross SDK token deserialization.  Discussed among the Nexus team and we are okay making this change due to the low usage of dotnet Nexus and its current release stage.
  

---

> [!NOTE]
> **Medium Risk**
> Changes the on-the-wire operation token format (base64url + corrected JSON key), which may break consumers relying on the previous encoding/typo despite added compatibility tests.
> 
> **Overview**
> Fixes Nexus operation token serialization/deserialization in `NexusWorkflowRunHandle` by switching to **base64url without padding** and correcting the token JSON field name from `t'` to `t`.
> 
> `FromToken` now distinguishes invalid base64 vs invalid JSON (wrapping `JsonException`) and continues to reject unsupported non-zero versions; new unit tests cover round-trips, base64url formatting, version handling, invalid inputs, and cross-SDK (Python) token compatibility.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7a0358b52d280c5c8077899d99fb913ed8ef0850. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


tests/Temporalio.Tests/Nexus/NexusWorkflowRunHandleTests.cs
```diff
@@ -0,0 +1,126 @@
...
+        switch (base64.Length % 4)
+        {
+            case 2: base64 += "=="; break;
+            case 3: base64 += "="; break;
+        }
+        var json = Encoding.UTF8.GetString(Convert.FromBase64String(base64));
+        using var doc = JsonDocument.Parse(json);
+        var root = doc.RootElement;
+
+        // Key must be "t" (not "t'")
```
VegetarianOrc, line 45: Nit: This comment makes sense in the context of this change but will be weird to read later.
```suggestion
```

VegetarianOrc (approved):

tests/Temporalio.Tests/Nexus/NexusWorkflowRunHandleTests.cs
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    [Fact]
+    public void ToToken_JsonUsesCorrectKeys()
+    {
+        var handle = new NexusWorkflowRunHandle("ns", "wid", 0);
+        var token = handle.ToToken();
+
+        // Decode manually to inspect JSON
+        var base64 = token.Replace('-', '+').Replace('_', '/');
```
jmaeagle99: Maybe just make the handle's `Base64UrlEncode` and `Base64UrlDecode` methods internal and call them from the test instead of duplicating them?

tests/Temporalio.Tests/Nexus/NexusWorkflowRunHandleTests.cs
```diff
@@ -0,0 +1,125 @@
...
+    public void FromToken_RejectsUnsupportedVersion()
+    {
+        var json = """{"t":1,"ns":"ns","wid":"wid","v":2}""";
+        var token = ToBase64Url(json);
+
+        Assert.Throws<ArgumentException>(() => NexusWorkflowRunHandle.FromToken(token));
+    }
+
+    [Fact]
+    public void FromToken_RejectsInvalidBase64()
```
jmaeagle99, line 85: Slightly similar to this test, maybe add a test that calls `FromToken` on something that is a valid base64 url encoding but invalidate JSON. I think a `JsonException` will propagate out of https://github.com/temporalio/sdk-dotnet/pull/629/changes#diff-cc1d8ad90218cfbebdfc95e9e9605f658a5d92a7f26537c7dbf6d7654495c04dR72

tests/Temporalio.Tests/Nexus/NexusWorkflowRunHandleTests.cs
```diff
@@ -0,0 +1,125 @@
...
+using System.Text.Json;
+using Temporalio.Nexus;
+using Xunit;
+
+public class NexusWorkflowRunHandleTests
+{
+    [Fact]
+    public void ToToken_RoundTrips()
+    {
+        var handle = new NexusWorkflowRunHandle("my-namespace", "my-workflow-id", 0);
```
jmaeagle99, line 13: There are tests for version `null`, `0`, and `2` but not for `1`. Is the version actually significant to test, or did we just happen to "skip" `1`?

tests/Temporalio.Tests/Nexus/NexusWorkflowRunHandleTests.cs
```diff
@@ -0,0 +1,125 @@
...
+    [Fact]
+    public void CrossSdkCompatibility_PythonToken()
+    {
+        // This token was generated by Python SDK with {"t":1,"ns":"default","wid":"my-wf"} and base64url encoding
+        var pythonToken = "eyJ0IjoxLCJucyI6ImRlZmF1bHQiLCJ3aWQiOiJteS13ZiJ9";
+
+        var handle = NexusWorkflowRunHandle.FromToken(pythonToken);
+
+        Assert.Equal("default", handle.Namespace);
+        Assert.Equal("my-wf", handle.WorkflowId);
```
jmaeagle99, line 106: Assert the expected version

tests/Temporalio.Tests/Nexus/NexusWorkflowRunHandleTests.cs
Quinn-With-Two-Ns: Can do

tests/Temporalio.Tests/Nexus/NexusWorkflowRunHandleTests.cs
Quinn-With-Two-Ns: I can test 1 here but it's not really providing any unique test coverage. We could equally ask why did we not test 3 or higher.

tests/Temporalio.Tests/Nexus/NexusWorkflowRunHandleTests.cs
Quinn-With-Two-Ns: Yeah that's a good catch we should be throwing `ArgumentException`

tests/Temporalio.Tests/Nexus/NexusWorkflowRunHandleTests.cs
jmaeagle99: Just making sure that version wasn't significant for parsing other than it just parses correctly.

jmaeagle99 (approved):

PR #637, opened 2026-04-07
generate apis from 1.13.0

robcao:
## What was changed

I ran `dotnet run --project src/Temporalio.Api.Generator` to add the new cloud service endpoints

## Why?

Users want to be able to use the new cloud service endpoints like `GetAuditLogs` and `GetBillingReport` without having to maintain a fork now that the official repository has the protobuf definitions, since protobuf messages are global and the cloud protobuf messages are packaged in the `Temporalio` assembly

## Checklist

1. Closes 

2. How was this tested:

auto-generated code added, routine activity

3. Any docs updates needed?

no

### DIFF
--- a/src/Temporalio/Client/CloudService.Rpc.cs
+++ b/src/Temporalio/Client/CloudService.Rpc.cs
@@ -32,6 +32,17 @@ namespace Temporalio.Client
         }
 
         /// <summary>
+        /// Invoke CreateAccountAuditLogSink.
+        /// </summary>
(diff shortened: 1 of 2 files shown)


jmaeagle99:
Thanks for doing this! I've added these files to the diff generator in https://github.com/temporalio/sdk-dotnet/pull/645 so these get included on every API update.

jmaeagle99 (approved):

PR #639, opened 2026-04-08
💥 Add support for Stand Alone Nexus Operations

Quinn-With-Two-Ns:
# Summary

  - Adds stand-alone Nexus operation support, allowing Nexus operations to be started, managed, and queried directly from the client
  (outside of workflows)
  - Updates sdk-core submodule to stand-alone-nexus-operations branch and regenerates proto bindings (8 new RPCs)
  - Introduces NexusClient / NexusClient<TService> with both string-based and expression-based operation invocation, mirroring the existing
   NexusWorkflowClient pattern

  ## API Surface

  On ITemporalClient:
  - CreateNexusClient(service, options) / CreateNexusClient<TService>(options) — factory methods
  - GetNexusOperationHandle(operationId, runId?) — handle for existing operations
  - ListNexusOperationsAsync(query) / CountNexusOperationsAsync(query) — visibility queries

  On NexusClient:
  - StartNexusOperationAsync / ExecuteNexusOperationAsync — start or start+await operations

  On NexusOperationHandle:
  - GetResultAsync — polls PollNexusOperationExecution until completion
  - DescribeAsync / CancelAsync / TerminateAsync

  All operations flow through the ClientOutboundInterceptor chain, consistent with stand-alone activities.

  ## Test plan

  - Existing Nexus worker tests pass (30/30 verified)
  - Build succeeds with 0 errors across all target frameworks
  - Integration tests for start, get result, describe, cancel, terminate, list, count
  - Verify expression-based invocation with NexusClient<TService>
  - Verify NexusOperationAlreadyStartedException on ID conflict

---

> [!NOTE]
> **Medium Risk**
> Adds a new experimental client-facing API surface and wires it into new RPCs for starting/polling/listing/canceling/terminating Nexus operations; behavior changes could impact Nexus operation execution, visibility queries, and interceptor integrations.
> 
> **Overview**
> Adds **experimental standalone Nexus operations** to the .NET SDK, enabling Nexus operations to be started and managed directly from `ITemporalClient` via new `NexusClient`/`NexusOperationHandle` types (start/execute, poll for result, describe, cancel, terminate) plus list/count APIs (including manual pagination).
> 
> Introduces a full interceptor surface for these client operations (`ClientOutboundInterceptor.*NexusOperation*` inputs/methods), maps them to new workflow-service RPCs in `TemporalClient.NexusOperation`, and adds dedicated exceptions for already-started and failed operations.
> 
> Also updates workflow-side naming from `StartNexusOperation*` to `ScheduleNexusOperation*`, ensures Nexus task cancellation reason is stable by honoring only the first cancel, and adds a comprehensive new test suite plus local test environment config to enable the standalone Nexus operations server feature.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a1371af3621a04ccdc336b18686bfa8904a78ca6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/Temporalio/Client/NexusOperationDescribeOptions.cs
```diff
@@ -0,0 +1,36 @@
...
+        /// <summary>
+        /// Gets or sets RPC options for describing the operation.
+        /// </summary>
+        public RpcOptions? Rpc { get; set; }
+
+        /// <summary>
+        /// Create a shallow copy of these options.
+        /// </summary>
+        /// <returns>A shallow copy of these options and any transitive options fields.</returns>
+        public virtual object Clone()
```
jmaeagle99, line 20: Need to deep copy the `LongPollToken` value. While the array itself is immutable, and would be captured by the `MemberwiseClone`, that means the original and new instances share the same array instance. The elements of the array are still mutable.

src/Temporalio/Client/NexusOperationExecutionCount.cs
```diff
@@ -0,0 +1,70 @@
...
+            Count = raw.Count;
+            Groups = raw.Groups?.Select(
+                g => new AggregationGroup(g)).ToArray() ?? Array.Empty<AggregationGroup>();
+        }
+
+        /// <summary>
+        /// Gets the approximate number of operations matching the original query. If the query had a
+        /// group-by clause, this is simply the sum of all the counts in <see cref="Groups" />.
+        /// </summary>
+        public long Count { get; private init; }
```
jmaeagle99:
```suggestion
        public long Count { get; }
```
Unless this are used someway outside of the constructor, they do not need to be `init` setters. In fact, the setter can be completely dropped, making them only assignable in the constructor.

src/Temporalio/Client/NexusOperationOptions.cs
```diff
@@ -0,0 +1,81 @@
...
+        /// </summary>
+        public RpcOptions? Rpc { get; set; }
+
+        /// <summary>
+        /// Create a shallow copy of these options.
+        /// </summary>
+        /// <returns>A shallow copy of these options and any transitive options fields.</returns>
+        public virtual object Clone()
+        {
+            var copy = (NexusOperationOptions)MemberwiseClone();
```
jmaeagle99, line 73: Do we need to deep clone the `SearchAttributes` too?

src/Temporalio/Client/NexusClientOptions.cs
```diff
@@ -0,0 +1,38 @@
...
+        /// <param name="endpoint">Endpoint name.</param>
+        public NexusClientOptions(string endpoint)
+        {
+            Endpoint = endpoint;
+        }
+
+        /// <summary>
+        /// Gets or sets the endpoint name, resolved to a URL via the cluster's endpoint registry.
+        /// </summary>
+        public string? Endpoint { get; set; }
```
jmaeagle99: The consumers of this property require that it is not `null`. I would suggest making this property not null and removing the default constructor.

src/Temporalio/Client/TemporalClient.NexusOperation.cs
```diff
@@ -0,0 +1,322 @@
...
+namespace Temporalio.Client
+{
+    public partial class TemporalClient
+    {
+        /// <inheritdoc />
+        public NexusClient CreateNexusClient(string service, NexusClientOptions options) =>
+            new NexusClientImpl(
+                client: this,
+                service: service,
+                endpoint: options.Endpoint ?? throw new ArgumentException("Endpoint is required"));
```
jmaeagle99: If https://github.com/temporalio/sdk-dotnet/pull/639/changes#r3119275089 is adopted, then the "null coalesce and throw" is not necessary.

src/Temporalio/Client/TemporalClient.NexusOperation.cs
```diff
@@ -0,0 +1,322 @@
...
+                StartNexusOperationInput input)
+            {
+                try
+                {
+                    var dataConverter = Client.Options.DataConverter;
+
+                    var req = new StartNexusOperationExecutionRequest()
+                    {
+                        Namespace = Client.Options.Namespace,
+                        OperationId = input.Options.Id ?? throw new ArgumentException("ID required to start Nexus operation"),
```
jmaeagle99:
```suggestion
                        OperationId = string.IsNullOrEmpty(input.Options.Id) ? throw new ArgumentException("ID required to start Nexus operation") : input.Options.Id,
```
Should this check for empty string as well?

src/Temporalio/Client/NexusOperationHandle.cs
```diff
@@ -0,0 +1,163 @@
...
+        /// <summary>
+        /// Wait for the result of the operation, discarding the return value.
+        /// </summary>
+        /// <param name="options">Options for the call.</param>
+        /// <returns>Task that completes when the operation completes.</returns>
+        /// <exception cref="NexusOperationFailedException">
+        /// Exception thrown for unsuccessful operation result.
+        /// </exception>
+        /// <exception cref="RpcException">Server-side error.</exception>
+        /// <remarks>WARNING: Standalone Nexus operations are experimental.</remarks>
```
jmaeagle99, line 33: In .NET, it is idiomatic that async methods have a `CancellationToken` as the last parameter on the method itself rather than buried inside other parameters. If using `NexusOperationGetResultOptions.Rpc.CancellationToken` is going to be the way to specify this, then this at least should be clearly documented in the top-level API where .NET developers would normally expect the `CancellationToken` parameter to be.

src/Temporalio/Client/NexusClient.cs
```diff
@@ -0,0 +1,199 @@
...
+
+        /// <summary>
+        /// Gets the endpoint name.
+        /// </summary>
+        public abstract string Endpoint { get; }
+
+        /// <summary>
+        /// Gets the underlying Temporal client.
+        /// </summary>
+        public abstract ITemporalClient Client { get; }
```
Quinn-With-Two-Ns: Probably don't need to expose this

src/Temporalio/Client/NexusOperationDescribeOptions.cs
```diff
@@ -0,0 +1,36 @@
...
+    /// Options for describing a standalone Nexus operation.
+    /// </summary>
+    /// <remarks>WARNING: Standalone Nexus operations are experimental.</remarks>
+    public class NexusOperationDescribeOptions : ICloneable
+    {
+        /// <summary>
+        /// Gets or sets a long-poll token from a previous describe response. When set, the
+        /// describe call will long-poll for state changes.
+        /// </summary>
+        public byte[]? LongPollToken { get; set; }
```
Quinn-With-Two-Ns: Not sure we should expose this on the SDK, we did for SAA but maybe that was a bug

src/Temporalio/Client/NexusOperationOptions.cs
```diff
@@ -0,0 +1,81 @@
...
+        /// <param name="id">Operation ID.</param>
+        public NexusOperationOptions(string id)
+        {
+            Id = id;
+        }
+
+        /// <summary>
+        /// Gets or sets the unique operation identifier. This is required.
+        /// </summary>
+        public string? Id { get; set; }
```
jmaeagle99, line 32: I find it strange that this is nullable but yet required. I think the argument was that it should be nullable just in case `NexusOperationOptions` is composed via `IOptions<T>`. But that can be easily be done if the property was non-nullable and its default value is set to `string.Empty` in the default constructor. We should probably be checking that the Id property is not null or empty when we go to consume it, right?

src/Temporalio/Client/NexusOperationExecutionDescription.cs
```diff
@@ -0,0 +1,126 @@
...
+        /// Gets the raw proto description.
+        /// </summary>
+        internal DescribeNexusOperationExecutionResponse RawDescription { get; private init; }
+
+        /// <summary>
+        /// Gets the single-line fixed summary for this operation that may appear in UI/CLI.
+        /// This can be in single-line Temporal markdown format.
+        /// </summary>
+        /// <returns>Static summary.</returns>
+        public async Task<string?> GetStaticSummaryAsync() =>
```
jmaeagle99, line 123: Should there be an analogous method to get the description?

src/Temporalio/Client/NexusOperationExecutionDescription.cs
```diff
@@ -0,0 +1,126 @@
...
+            ExpirationTime = info.ExpirationTime?.ToDateTime();
+            LastAttemptCompleteTime = info.LastAttemptCompleteTime?.ToDateTime();
+            LongPollToken = rawDescription.LongPollToken.IsEmpty ? null : rawDescription.LongPollToken.ToByteArray();
+            NextAttemptScheduleTime = info.NextAttemptScheduleTime?.ToDateTime();
+            OperationToken = string.IsNullOrEmpty(info.OperationToken) ? null : info.OperationToken;
+            RequestId = string.IsNullOrEmpty(info.RequestId) ? null : info.RequestId;
+            ScheduleToCloseTimeout = info.ScheduleToCloseTimeout?.ToTimeSpan();
+            State = info.State;
+#pragma warning disable VSTHRD011 // This should not be able to deadlock
+            userMetadata = new(() => dataConverter.FromUserMetadataAsync(info.UserMetadata));
```
jmaeagle99, line 59:
```suggestion
            userMetadata = new(() => dataConverter.FromUserMetadataAsync(info.UserMetadata), LazyThreadSafetyMode.PublicationOnly);
```

The safety mode is being used in `NexusOperationExecution` for search attributes, so why not here? Might be able to remove the pragma.

src/Temporalio.Extensions.OpenTelemetry/TracingInterceptor.cs
```diff
@@ -701,7 +701,7 @@ public override Task<ChildWorkflowHandle<TWorkflow, TResult>> StartChildWorkflow
             }
 
             public override Task<NexusWorkflowOperationHandle<TResult>> StartNexusOperationAsync<TResult>(
-                StartNexusOperationInput input)
+                Temporalio.Worker.Interceptors.StartNexusOperationInput input)
```
jmaeagle99: Any particular reason for the fully qualified type here or could this be reverted?

src/Temporalio/Client/NexusOperationExecutionDescription.cs
```diff
@@ -0,0 +1,126 @@
...
+            : base(
+                clientNamespace: clientNamespace,
+                operationId: rawDescription.Info.OperationId,
+                runId: string.IsNullOrEmpty(rawDescription.Info.RunId) ? null : rawDescription.Info.RunId,
+                endpoint: rawDescription.Info.Endpoint,
+                service: rawDescription.Info.Service,
+                operation: rawDescription.Info.Operation,
+                closeTime: rawDescription.Info.CloseTime?.ToDateTime(),
+                executionDuration: rawDescription.Info.ExecutionDuration?.ToTimeSpan(),
+                scheduledTime: rawDescription.Info.ScheduleTime?.ToDateTime() ?? default,
```
jmaeagle99, line 39: Is `ScheduleTime` possibly nullable even though it is not annotated as nullable and that's why null coalescing is being used? The `?? default` is going to result in `DateTime.MinValue`, which is something like "Jan 1, 0001". That feels like a very non-obvious value to expect. Maybe `NexusOperationExecution.ScheduleTime` need to be nullable instead.

src/Temporalio/Client/CloudService.Rpc.cs
```diff
@@ -31,6 +31,17 @@ public Task<AddUserGroupMemberResponse> AddUserGroupMemberAsync(AddUserGroupMemb
             return InvokeRpcAsync("AddUserGroupMember", req, AddUserGroupMemberResponse.Parser, options);
         }
 
+        /// <summary>
```
jmaeagle99, line 34: When you merge main, these will go away.

src/Temporalio/Exceptions/NexusOperationFailedException.cs
```diff
@@ -0,0 +1,20 @@
+using System;
+
+namespace Temporalio.Exceptions
+{
+    /// <summary>
+    /// Exception thrown when a standalone Nexus operation has failed while waiting for the result.
+    /// </summary>
+    /// <remarks>WARNING: Standalone Nexus operations are experimental.</remarks>
+    public class NexusOperationFailedException : TemporalException
```
jmaeagle99, line 9: Consider adding `OperationId` and `RunId` properties.

jmaeagle99 (commented):
Left a bunch of feedback, nothing major. However, tests should also be added for the new capabilities introduced in this change.

src/Temporalio/Client/NexusOperationExecutionCount.cs
Quinn-With-Two-Ns: Yeah this was copied from `WorkflowExecutionCount` so maybe it can be removed there as well?

src/Temporalio/Client/NexusOperationOptions.cs
Quinn-With-Two-Ns: We didn't in other places like `WorkflowOptions` so I didn't here

src/Temporalio/Client/NexusOperationHandle.cs
Quinn-With-Two-Ns: Yeah I'll change it to be consistent with how stand alone activities did it

src/Temporalio/Client/NexusOperationOptions.cs
Quinn-With-Two-Ns: I copied the same pattern we had for other `IDs` in other options like `WorkflowOptions` and `StartActivityOptions`

src/Temporalio/Client/NexusOperationExecutionDescription.cs
Quinn-With-Two-Ns: Yep, good catch!

src/Temporalio/Client/NexusOperationExecutionDescription.cs
Quinn-With-Two-Ns: I copied this from `ActivityExecutionDescription`. I think making `NexusOperationExecution.ScheduleTime` nullable would be weird since from the users perspective

src/Temporalio/Client/NexusOperationExecutionDescription.cs
Quinn-With-Two-Ns: I copied this from the stand alone activity code. we also did the same thing in Workflow code
```
        protected internal WorkflowExecutionDescription(
            DescribeWorkflowExecutionResponse rawDescription, DataConverter dataConverter)
            : base(rawDescription.WorkflowExecutionInfo, dataConverter, null)
        {
            RawDescription = rawDescription;
#pragma warning disable VSTHRD011 // This should not be able to deadlock
            userMetadata = new(() => dataConverter.FromUserMetadataAsync(
                rawDescription.ExecutionConfig?.UserMetadata));
#pragma warning restore VSTHRD011
        }
```

src/Temporalio.Extensions.OpenTelemetry/TracingInterceptor.cs
Quinn-With-Two-Ns: I used the  fully qualified type since there is a duplicate name in the ` Temporalio.Client.Interceptors`  I think we need to rename the workfloiw interceptor type

src/Temporalio/Client/NexusOperationOptions.cs
```diff
@@ -0,0 +1,81 @@
...
+        /// <returns>A shallow copy of these options and any transitive options fields.</returns>
+        public virtual object Clone()
+        {
+            var copy = (NexusOperationOptions)MemberwiseClone();
+            if (Rpc != null)
+            {
+                copy.Rpc = (RpcOptions)Rpc.Clone();
+            }
+            return copy;
+        }
```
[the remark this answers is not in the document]
Quinn-With-Two-Ns: Discussed above

jmaeagle99 (approved):

src/Temporalio/Client/NexusOperationHandle.cs
```diff
@@ -0,0 +1,163 @@
...
+                        if (typeof(TResult) == typeof(ValueTuple) || resp.Result == null)
+                        {
+                            return default!;
+                        }
+                        return await dataConverter.ToSingleValueAsync<TResult>(
+                            new[] { resp.Result }).ConfigureAwait(false);
+                    case PollNexusOperationExecutionResponse.OutcomeOneofCase.Failure:
+                        var failure = await dataConverter.ToExceptionAsync(
+                            resp.Failure).ConfigureAwait(false);
+                        throw new NexusOperationFailedException(Id, resp.RunId, failure);
```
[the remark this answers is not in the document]
Quinn-With-Two-Ns: This matches other workflow code like `WorkflowAlreadyStartedException`

PR #641, opened 2026-04-13
Add Temporalio.Extensions.ToolRegistry with AgenticSession support

lex00:
## What was changed

New `Temporalio.Extensions.ToolRegistry` package for running LLM tool-calling loops inside Temporal activities.

- `ToolRegistry` — maps tool names to JSON Schema definitions and handler functions
- `AgenticSession` — crash-safe wrapper that checkpoints conversation history via `ActivityExecutionContext.HeartbeatAsync` on each turn and restores on retry; session survives both activity crashes and provider-side session expiry since state is stored locally
- Built-in `AnthropicProvider` and `OpenAIProvider`
- `Temporalio.Extensions.ToolRegistry.Testing` with `MockProvider` and `FakeToolRegistry` for unit testing without a live API key
- `ToolRegistry.FromMcpTools` — converts a sequence of MCP tool descriptors into a populated registry

## Bug fix included

The Anthropic provider now sets `is_error: true` on tool result maps when a handler throws, matching the Anthropic API spec. A new test (`AnthropicProvider_HandlerError_SetsIsError`) verifies this using an `HttpListener`-based mock server. OpenAI has no equivalent field.

## Why?

Temporal activities are a natural fit for LLM tool-calling loops, but every team reimplements the same boilerplate. This extension package standardizes the pattern across all six Temporal SDKs. .NET has no framework-level LLM integration (Python's `openai_agents`/`google_adk_agents` and TypeScript's `@temporalio/ai-sdk` are Python/TypeScript only), so this is the recommended path for direct Anthropic/OpenAI calls in .NET.

Proposal: temporalio/proposals#107

## Checklist

1. Related to: temporalio/proposals#107
2. How was this tested: unit tests with `MockProvider` (no API key required); new `AnthropicProvider_HandlerError_SetsIsError` test uses `HttpListener` mock; integration tests gated on `RUN_INTEGRATION_TESTS=1`
3. Any docs updates needed? Yes — docs.temporal.io update to follow after merge

### DIFF
--- a/Temporalio.sln
+++ b/Temporalio.sln
@@ -17,6 +17,12 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Temporalio.Extensions.Hosti
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Temporalio.Extensions.DiagnosticSource", "src\Temporalio.Extensions.DiagnosticSource\Temporalio.Extensions.DiagnosticSource.csproj", "{CC7EA7CD-BBE7-448C-8A4B-F8B2D1E55990}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Temporalio.Extensions.ToolRegistry", "src\Temporalio.Extensions.ToolRegistry\Temporalio.Extensions.ToolRegistry.csproj", "{A1B2C3D4-E5F6-7890-ABCD-EF1234567890}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Temporalio.Extensions.ToolRegistry.Testing", "src\Temporalio.Extensions.ToolRegistry.Testing\Temporalio.Extensions.ToolRegistry.Testing.csproj", "{B2C3D4E5-F6A7-8901-BCDE-F12345678901}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Temporalio.Extensions.ToolRegistry.Tests", "tests\Temporalio.Extensions.ToolRegistry.Tests\Temporalio.Extensions.ToolRegistry.Tests.csproj", "{C3D4E5F6-A7B8-9012-CDEF-012345678902}"
+EndProject
 Global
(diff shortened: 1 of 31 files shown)


lex00:
Real .NET handlers tend to do I/O. HTTP for MCP forward, Anthropic for fallback turns, Temporal client for HITL signalWithStart. With the current `Func<IReadOnlyDictionary<string, object?>, string>` signature, the only escape is `.GetAwaiter().GetResult()` inside the handler body. The .NET port of the incident-triage example accumulated 8 of those calls.

The proposal's "Known limitations" deliberately accepts sync handlers across Go/Java/Ruby/.NET for cross-SDK consistency. .NET's `Task<string>` is more idiomatically async than Java's `CompletableFuture` or Go's goroutine equivalents. Worth revisiting just for .NET, or hold the line on consistency?

If revisited, the shape would be:

```csharp
public void Register(ToolDef def, Func<IReadOnlyDictionary<string, object?>, Task<string>> handler);
public Task<string> DispatchAsync(string name, IReadOnlyDictionary<string, object?> input);
```

`MockProvider.RunTurnAsync` already returns `Task` so the plumbing exists. Filing as a question, not a request to block the PR.

lex00:
Reporting consumer-side observation. `<ProjectReference>` consumers of `Temporalio.Extensions.ToolRegistry` get 134 build errors in SDK source files (SA1101, CA1062, CA1848) under `dotnet build -c Debug`. Cause: `Directory.Build.props` sets `<TreatWarningsAsErrors>true</TreatWarningsAsErrors>` and references `StyleCop.Analyzers` plus `Microsoft.VisualStudio.Threading.Analyzers` as `PackageReference`. Both inherit through `ProjectReference` consumers and run analyzers over the SDK's own source under Debug.

The local workaround is `dotnet pack` plus a local feed plus switch to `<PackageReference>`. Workable but adds a 10-minute setup to first-time consumption of the unmerged branch.

Two ways to support `<ProjectReference>` consumers if either is wanted:

1. Publish a SNAPSHOT `.nupkg` for the unmerged branch.
2. Scope `TreatWarningsAsErrors` to Release / in-repo builds: `<TreatWarningsAsErrors Condition="'$(Configuration)' == 'Release' or '$(BuildingInsideTemporalRepo)' == 'true'">true</TreatWarningsAsErrors>`.

Filing as a question.

PR #644, opened 2026-04-16
Central package management and lock files

jmaeagle99:
## What was changed

- Centralize package versioning in the Directory.Packages.props file at the repository root.
- Use package lock files to pin package versions and require explicit package version updates.

## Why?

Unify the "source of truth" of the package version information.
Allow easier package version updating with fewer source locations.
Better deterministic builds with explicitly known package versions.

## Checklist

1. How was this tested: Local and CI Build

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,6 +11,8 @@ permissions:
 
 jobs:
   build-lint-test:
+    env:
+      RestoreLockedMode: true
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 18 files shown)


.github/workflows/ci.yml
```diff
@@ -11,6 +11,8 @@ permissions:
 
 jobs:
   build-lint-test:
+    env:
+      RestoreLockedMode: true
```
Sushisource, line 15: What's this guy do?

Sushisource (approved):

.github/workflows/ci.yml
jmaeagle99: Forces nuget to restore the packages with the exact versions specified in the lock files; will error otherwise.

PR #647, opened 2026-04-22
feat(otel): add otel tracing for standalone activities.

cconstable:
## What was changed
Implemented OTel interceptor methods for client standalone activities:
- start_activity

## Why?
Users now have OTel tracing info available for standalone activities.

## Checklist
1. Closes SDK-5175
2. How was this tested: new tests + `dotnet test`
3. Any docs updates needed? no


tests/Temporalio.Tests/Extensions/OpenTelemetry/TracingInterceptorTests.cs
```diff
@@ -886,7 +923,7 @@ private async Task<IReadOnlyCollection<Activity>> WithTracingWorkerAsync(
         var activities = new List<Activity>();
 
         // Setup provider
-        using var tracerProvider = global::OpenTelemetry.Sdk.CreateTracerProviderBuilder().
+        using var _ = global::OpenTelemetry.Sdk.CreateTracerProviderBuilder().
```
cconstable, line 910: Took me a second to realize that `tracerProvider` was not actually being used we were just using `using` for lifetime management. Swapped to discard variable instead which preserves lifetime management but better signals intent.

src/Temporalio.Extensions.OpenTelemetry/TracingInterceptor.cs
```diff
@@ -395,6 +410,112 @@ public override async Task<WorkflowUpdateHandle<TResult>> StartWorkflowUpdateAsy
...
             }
 
+            public override async Task<ActivityHandle<TResult>> StartActivityAsync<TResult>(
+                StartActivityInput input)
+            {
+                using (var activity = ClientSource.StartActivity(
+                    $"StartActivity:{input.Activity}",
+                    kind: ActivityKind.Client,
+                    parentContext: default,
+                    tags: input.Options.Id is string id ? root.CreateStandaloneActivityTags(id) : null))
```
cconstable, line 419: `input.Options.Id` is optional because it uses the shared `StartActivityInput`. AFAICT, it will never be null for standalone activities.

src/Temporalio.Extensions.OpenTelemetry/TracingInterceptor.cs
```diff
@@ -231,6 +231,21 @@ protected virtual IDictionary<string, string> HeadersFromContext(
             return ret;
         }
 
+        /// <summary>
+        /// Create tag collection for the given standalone activity ID.
+        /// </summary>
+        /// <param name="activityId">Standalone activity ID.</param>
+        /// <returns>Tags.</returns>
+        protected virtual IEnumerable<KeyValuePair<string, object?>> CreateStandaloneActivityTags(
```
cconstable, line 239: I could just inline this logic now that start_activity is the only event we are adding tracing to.

src/Temporalio.Extensions.OpenTelemetry/TracingInterceptor.cs
jmaeagle99: It shouldn't be `null` (and is documented that it is required), but customers can still pass `null`. Given the current type declaration, you are doing the right thing here. The validation of the value is done within the innermost interceptor, which hasn't been called at this point.

src/Temporalio.Extensions.OpenTelemetry/TracingInterceptor.cs
jmaeagle99: I think it's fine to leave here as-is. There's precedent for this pattern.

tests/Temporalio.Tests/Extensions/OpenTelemetry/TracingInterceptorTests.cs
```diff
@@ -797,6 +797,27 @@ public async Task TracingInterceptor_Nexus_HasTracing()
...
     }
 
+    [Fact]
+    public async Task TracingInterceptor_StandaloneActivity_HasProperSpans()
+    {
+        var activityId = $"act-{Guid.NewGuid()}";
+        var spans = await WithTracingWorkerAsync(async (client, _) =>
+        {
+            await client.StartActivityAsync(
+                "StandaloneActivity",
```
jmaeagle99: nit: choose a different name so we don't accidentally happen to match the well known primitive name

jmaeagle99 (approved):

PR #671, opened 2026-05-01
ci: overwrite uploaded artifacts for ci jobs.

cconstable:
overwrite previous "ci artifacts" whenever the ci workflow is triggered. this prevents test results, etc from previous runs showing up and muddying things.


Sushisource (approved):
Wow, kind of unintuitive that they aren't automatically different per run

PR #676, opened 2026-05-06
expose DnsLoadBalancingOptions

robcao:
## What was changed

https://github.com/temporalio/sdk-rust/issues/1250

DNS load balancing was introduced in https://github.com/temporalio/sdk-rust/pull/1212, and it defaults to enabled for users of the C bridge.

These options should be exposed through the C bridge, so that users of the C bridge, such as .NET SDK users, can configure DNS load balancing.

## Why?

.NET users want to be able to change the DNS load balancing options of the gRPC client

## Checklist

1. Closes 

2. How was this tested:

I ran the tests locally

3. Any docs updates needed?

If the Rust side change is merged, will need to update the submodule to point to the upstream repo instead of my fork: https://github.com/temporalio/sdk-rust/issues/1250


jmaeagle99:
Thank you for bringing the issue to our attention and proactively providing a fix.

I ended up reimplementing the change in https://github.com/temporalio/sdk-dotnet/pull/689 without the enabled field and aligned it with the behavior of the other SDKs. I'll close out this PR at this time since it has been superseded.

PR #690, opened 2026-05-13
Generic Temporal Nexus Operation Handler

Quinn-With-Two-Ns:
## What was changed

Added a generic Nexus operation handler that consolidates the common pattern of "either return a synchronous
result or start a workflow as an async operation."

## Why?

  The existing `WorkflowRun` /` SyncOperation` split forces users to pick the operation shape up front, which is awkward when:  
  - The start handler decides at runtime whether to respond sync or async.
  - The handler needs access to the Temporal client.                          
  - The user wants a custom cancel implementation (e.g. terminate instead of cancel). 
  - Difficult to extend as we add new operations                                                   
                                                                                                                                           
  The new TemporalOperation unifies these cases behind one API and makes the common path much shorter. 

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

---

> [!NOTE]
> **Medium Risk**
> Changes experimental Nexus operation start/cancel and callback token wiring shared by old and new handlers; mistakes could break async completion or cancel the wrong workflow, though coverage is broad.
> 
> **Overview**
> Adds a **unified experimental Nexus API** (`TemporalOperationHandler.FromHandleFactory`) so one handler can return either a **sync** result or an **async** workflow-run token, with `ITemporalNexusClient` / `TemporalNexusClient` for starting workflows and accessing `TemporalClient` for signals and similar work.
> 
> Shared workflow-start behavior (links, completion callbacks, `Nexus-Operation-Token` header with case-insensitive dedup, `UseExisting` conflict options, outbound links) moves into **`NexusWorkflowStartHelper`**; **`WorkflowRunOperationContext`** now delegates there instead of inlining that logic.
> 
> Operation tokens gain **`ParseToken`**, an **`OperationToken`** record, and a **`WorkflowRunOperationTokenType`** constant; cancel on the generic handler validates namespace and token type, then calls overridable **`CancelWorkflowRunAsync`** (default: cancel workflow by ID). Unit and worker integration tests cover tokens, sync/async paths, cancel, links, and conflict policy.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c354d2ada5e331f140437e0f83c4f39748007653. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/Temporalio/Nexus/TemporalNexusOperationHandler.cs
```diff
@@ -0,0 +1,148 @@
...
+            {
+                throw new HandlerException(HandlerErrorType.BadRequest, e.Message);
+            }
+            if (token.Namespace != NexusOperationExecutionContext.Current.Info.Namespace)
+            {
+                throw new HandlerException(HandlerErrorType.BadRequest, "Invalid namespace");
+            }
+            return token.Type switch
+            {
+                1 => CancelWorkflowRunAsync(context, token.WorkflowId),
```
jmaeagle99: This feels like a magic number. I would suggest moving it to a constant so that the creation and the decoding of this value is consistent and documenting any meaningful notions about it.

src/Temporalio/Nexus/NexusWorkflowStartHelper.cs
```diff
@@ -0,0 +1,121 @@
...
+                workflow, args, options).ConfigureAwait(false);
+
+            // Add the outbound link
+            nexusStartContext.OutboundLinks.Add(new Link.Types.WorkflowEvent
+            {
+                Namespace = namespace_,
+                WorkflowId = workflowId,
+                RunId = wfHandle.FirstExecutionRunId ??
+                    throw new InvalidOperationException("Handle unexpectedly missing run ID"),
+                EventRef = new() { EventId = 1, EventType = EventType.WorkflowExecutionStarted },
```
jmaeagle99, line 116: Another magic value. Make const and document.

src/Temporalio/Nexus/TemporalNexusClient.cs
```diff
@@ -0,0 +1,103 @@
...
+        {
+            this.nexusStartContext = nexusStartContext;
+            temporalContext = NexusOperationExecutionContext.Current;
+            temporalClient = temporalContext.TemporalClient;
+        }
+
+        /// <inheritdoc/>
+        public ITemporalClient TemporalClient => temporalClient;
+
+        /// <summary>
```
jmaeagle99: These appear to be exactly the same as those on the interface at first glance. Let's use `/// <inheritdoc/>` instead of writing them out again. Apply to all interface method impls.

src/Temporalio/Nexus/TemporalNexusOperationHandler.cs
```diff
@@ -0,0 +1,148 @@
...
+    /// <see cref="IOperationHandler{TInput, TResult}"/> and provides a composable way to map
+    /// Temporal operations to Nexus operations.
+    /// </summary>
+    /// <typeparam name="TInput">Operation input type.</typeparam>
+    /// <typeparam name="TResult">Operation result type.</typeparam>
+    /// <remarks>
+    /// <para>WARNING: Nexus support is experimental.</para>
+    /// <para>This class supports inheritance to customize cancel behavior. Override
+    /// <see cref="CancelWorkflowRunAsync"/> to change how workflow-run cancellations are handled.
+    /// The <see cref="StartAsync"/> and <see cref="CancelAsync"/> methods should not be
```
jmaeagle99: These methods cannot be overridden because they are not virtual. I would drop this comment.

jmaeagle99: If there is some specific concern that derivations of the class will shadow these methods (and callers are not directly invoking these methods via the `TemporalNexusOperationHandler` class itself), then make them explicit interface implementations e.g. `async Task<OperationStartResult<TResult>> IOperationHandler<TInput, TResult>.StartAsync(...)`

jmaeagle99: If you want to keep it to doc comments, I would suggest adding these to each of the methods instead. They should compose with the inheritdoc, especially since the interface methods do not use `<remarks>`.

src/Temporalio/Nexus/TemporalNexusClient.cs
```diff
@@ -0,0 +1,103 @@
...
+        private readonly NexusOperationExecutionContext temporalContext;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="TemporalNexusClient"/> class.
+        /// </summary>
+        /// <param name="nexusStartContext">Nexus start context for callbacks and links.</param>
+        internal TemporalNexusClient(OperationStartContext nexusStartContext)
+        {
+            this.nexusStartContext = nexusStartContext;
+            temporalContext = NexusOperationExecutionContext.Current;
```
jmaeagle99: Have this dependency on an async local might make it difficult to reason about where the `TemporalNexusClient` can be created. I would suggest making this a ctor parameter and passing in the value. The handler is already taking a dependency on this async local, so making it pass the value along isn't difficult or out of the ordinary of the existing implementation.

src/Temporalio/Nexus/TemporalOperationResult.cs
```diff
@@ -0,0 +1,69 @@
...
+        /// </summary>
+        internal string? AsyncToken { get; }
+
+        /// <summary>
+        /// Create a synchronous result with the given value.
+        /// </summary>
+        /// <param name="value">The result value.</param>
+        /// <returns>A synchronous operation result.</returns>
+#pragma warning disable CA1000 // Intentional static factory on generic type
+        public static TemporalOperationResult<TResult> Sync(TResult? value) =>
```
jmaeagle99: Rename to `SyncResult` or `FromValue`.

src/Temporalio/Nexus/TemporalOperationResult.cs
```diff
@@ -0,0 +1,69 @@
...
+#pragma warning restore CA1000
+
+        /// <summary>
+        /// Create an asynchronous result with the given operation token.
+        /// </summary>
+        /// <param name="token">The operation token.</param>
+        /// <returns>An asynchronous operation result.</returns>
+        /// <exception cref="ArgumentException">If the token is null or empty.</exception>
+#pragma warning disable CA1000, VSTHRD200 // Intentional: static factory on generic type; "Async" refers to the result kind, not the method signature
+        public static TemporalOperationResult<TResult> Async(string token)
```
jmaeagle99: Rename to `AsyncResult` or `FromToken`.

src/Temporalio/Nexus/TemporalOperationResult.cs
```diff
@@ -0,0 +1,69 @@
...
+        /// <summary>
+        /// Gets a value indicating whether this is a synchronous result.
+        /// </summary>
+        public bool IsSyncResult { get; }
+
+        /// <summary>
+        /// Gets the synchronous result value. Only meaningful when <see cref="IsSyncResult"/> is
+        /// true.
+        /// </summary>
+        internal TResult? SyncValue { get; }
```
jmaeagle99: Rather than these be auto-generated properties, should we stored them in backing fields and have the properties throw an exception if `IsSyncResult` is not the expected state?

src/Temporalio/Nexus/TemporalNexusOperationHandler.cs
```diff
@@ -0,0 +1,148 @@
...
+        /// <summary>
+        /// Called when a cancel request is received for a workflow-run token (type=1). Override to
+        /// customize cancel behavior.
+        /// <para>Default behavior: cancels the underlying workflow.</para>
+        /// </summary>
+        /// <param name="context">The cancel context.</param>
+        /// <param name="workflowId">The workflow ID extracted from the operation token.</param>
+        /// <returns>Task for cancel completion.</returns>
+        protected virtual Task CancelWorkflowRunAsync(
+            OperationCancelContext context, string workflowId) =>
```
jmaeagle99: The `context` parameter is not used at all in this default implementation. Is this desired? How does user cancellation flow into the cancel gesture?

jmaeagle99 (commented):
Still reviewing but wanted to leave my initial comments.

src/Temporalio/Nexus/TemporalNexusOperationHandler.cs
Quinn-With-Two-Ns: Ah okay I think I was a bit confused about dotnets behaviour if a user made a subclass

src/Temporalio/Nexus/TemporalNexusClient.cs
Quinn-With-Two-Ns: done!

src/Temporalio/Nexus/TemporalOperationResult.cs
Quinn-With-Two-Ns: Sure can do

src/Temporalio/Nexus/TemporalNexusOperationHandler.cs
Quinn-With-Two-Ns: these are run by the worker so users don't cancel these methods and the server doesn't have a way cancel these method either. It is assumed they are fast (<10s)

src/Temporalio/Nexus/NexusWorkflowStartHelper.cs
```diff
@@ -0,0 +1,121 @@
...
+        /// </summary>
+        /// <param name="client">Temporal client.</param>
+        /// <param name="nexusStartContext">Nexus start context for callbacks and links.</param>
+        /// <param name="temporalContext">Temporal operation context for info and logging.</param>
+        /// <param name="workflow">Workflow type name.</param>
+        /// <param name="args">Workflow arguments.</param>
+        /// <param name="options">Workflow start options. ID and TaskQueue are required.</param>
+        /// <returns>Base64url-encoded operation token.</returns>
+        internal static async Task<string> StartWorkflowAndGetTokenAsync(
+            ITemporalClient client,
```
jmaeagle99: Could we eliminate this parameter and pull it from the `temporalContext` parameter? What would also allow deferring the resolution of the `ITemporalClient` until it's actually necessary.

src/Temporalio/Nexus/WorkflowRunOperationContext.cs
jmaeagle99, line 12: Do we want to guide customers to use `TemporalNexusOperationHandler` instead of `WorkflowRunOperationHandler`? Obsoleting would be a forcing function. If not that, maybe update the comments on `WorkflowRunOperationHandler` and related types to recommend using the new types.

src/Temporalio/Nexus/NexusWorkflowStartHelper.cs
```diff
@@ -0,0 +1,121 @@
...
+        /// adding outbound links.
+        /// </summary>
+        /// <param name="client">Temporal client.</param>
+        /// <param name="nexusStartContext">Nexus start context for callbacks and links.</param>
+        /// <param name="temporalContext">Temporal operation context for info and logging.</param>
+        /// <param name="workflow">Workflow type name.</param>
+        /// <param name="args">Workflow arguments.</param>
+        /// <param name="options">Workflow start options. ID and TaskQueue are required.</param>
+        /// <returns>Base64url-encoded operation token.</returns>
+        internal static async Task<string> StartWorkflowAndGetTokenAsync(
```
jmaeagle99: Consider returning the `NexusWorkflowRunHandle` instead of the token. The `WorkflowRunOperationContext` callers can then immediately return the returned result and the `TemporalNexusOperationHandler` callers can invoke `ToToken()` themselves.

tests/Temporalio.Tests/Worker/NexusWorkerTests.cs
```diff
@@ -1315,6 +1315,330 @@ await Workflow.CreateNexusWorkflowClient<IStringService>(endpoint).
...
+        var handle = await RunInWorkflowAsync(workerOptions, async () =>
+        {
+            var result = await Workflow.CreateNexusWorkflowClient<IStringService>(endpoint).
+                ExecuteNexusOperationAsync(svc => svc.DoSomething("some-name"));
+            Assert.Equal("Hello from workflow, some-name", result);
+        });
+
+        // Context fields populated
+        Assert.NotNull(capturedContext);
+        Assert.Equal("StringService", capturedContext!.Service);
```
jmaeagle99: Is the `!` assertion necessary? I think that the `Assert.NotNull` call should cover it. If not, that's a nullability enhancement we can consider making, at least for tests.

tests/Temporalio.Tests/Worker/NexusWorkerTests.cs
```diff
@@ -1315,6 +1315,330 @@ await Workflow.CreateNexusWorkflowClient<IStringService>(endpoint).
...
+        Assert.True(Guid.TryParse(capturedContext.RequestId, out _));
+        Assert.False(string.IsNullOrEmpty(capturedContext.Underlying.CallbackUrl));
+        // Inbound link points to the caller workflow's scheduled event
+        var wfEvent = Assert.Single(capturedContext.InboundLinks).ToWorkflowEvent();
+        Assert.Equal(handle.Id, wfEvent.WorkflowId);
+        Assert.Equal(Api.Enums.V1.EventType.NexusOperationScheduled, wfEvent.EventRef.EventType);
+
+        // Nexus client exposes the temporal client
+        Assert.NotNull(capturedClient);
+        Assert.Equal(Env.Client.Options.Namespace, capturedClient!.TemporalClient.Options.Namespace);
```
jmaeagle99: Same as https://github.com/temporalio/sdk-dotnet/pull/690/changes#r3344810394

tests/Temporalio.Tests/Worker/NexusWorkerTests.cs
```diff
@@ -1315,6 +1315,330 @@ await Workflow.CreateNexusWorkflowClient<IStringService>(endpoint).
...
+                        svc => svc.DoSomething("some-name"),
+                        new() { CancellationToken = cancelSource.Token });
+#pragma warning disable CA1849, VSTHRD103
+                cancelSource.Cancel();
+#pragma warning restore CA1849, VSTHRD103
+                await handle.GetResultAsync();
+            }));
+
+        Assert.NotNull(capturedHandler);
+        Assert.True(capturedHandler!.CancelCallCount > 0);
```
jmaeagle99: Same as https://github.com/temporalio/sdk-dotnet/pull/690/changes#r3344810394

jmaeagle99 (approved):

PR #698, opened 2026-05-15
fix(worker): allow Nexus-only workers

jmaeagle99:
## What was changed

Allow workers to only host Nexus services.

### DIFF
--- a/tests/Temporalio.Tests/Worker/GeneralWorkerTests.cs
+++ b/tests/Temporalio.Tests/Worker/GeneralWorkerTests.cs
@@ -13,6 +13,16 @@ public class GeneralWorkerTests : WorkflowEnvironmentTestBase
     }
 
     [Fact]
+    public void Constructor_NoTaskTypes_Throws()
+    {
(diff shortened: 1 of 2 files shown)


tests/Temporalio.Tests/Worker/GeneralWorkerTests.cs
```diff
@@ -12,6 +12,16 @@ public GeneralWorkerTests(ITestOutputHelper output, WorkflowEnvironment env)
...
     }
 
+    [Fact]
+    public void Constructor_NoTaskTypes_Throws()
+    {
+        var exc = Assert.Throws<ArgumentException>(() =>
+            new TemporalWorker(
+                Env.Client,
+                new TemporalWorkerOptions($"tq-{Guid.NewGuid()}")));
+        Assert.Equal("Must have at least one workflow, activity, or Nexus service", exc.Message);
```
maciejdudko: Nit: it's better not to rely on precise wording of exception message. Checking just exception type is entirely sufficient.

maciejdudko (approved):

PR #720, opened 2026-05-28
Make README.md more specific about which AWS Lambda images cause CA

eamsden:
issues

## What was changed
README section about CA loading issues on AWS lambda was changed to reflect that the issue should only appear for the AWS Lambda .NET 8 image

## Why?
The verbiage was slightly stale and didn't reflect that the performance regression was fixed in .NET 9 and thus AWS only kept the change for their .NET 8 image.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1265,7 +1265,7 @@ See the earlier part of this section for details on what environments are suppor
 
 Some platforms are not loading the system CA list properly due to quirks with how the platforms work.
 
-#### AWS Lambda .NET CA Loading Issues
+#### AWS Lambda .NET 8 CA Loading Issues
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1267,7 +1267,7 @@ Some platforms are not loading the system CA list properly due to quirks with ho
 
 #### AWS Lambda .NET CA Loading Issues
```
jmaeagle99:
```suggestion
#### AWS Lambda .NET 8 CA Loading Issues
```

jmaeagle99 (approved):

PR #721, opened 2026-05-29
AWS Lambda

eamsden:
## What was changed
Add helpers for serverless workers running on AWS Lambda

## Why?
SDKs require specific helpers to run effectively on AWS Lambda.

## Checklist

1. Closes 

2. How was this tested:
Code in a separate PR in samples-dotnet ( temporalio/samples-dotnet#152 )  was used to test deploying and running workflows and activities on a Lambda worker.

3. Any docs updates needed?


src/Temporalio.Extensions.Aws.Lambda/README.md
```diff
@@ -0,0 +1,132 @@
...
+    public Task HandlerAsync(object? input, ILambdaContext context) =>
+        WorkerHandler(input, context);
+}
+```
+
+`configure` runs once when the delegate is created, which is normally during Lambda cold start. Each invocation creates a
+fresh Temporal client and worker, runs until the Lambda deadline minus `ShutdownDeadlineBuffer`, then shuts down and runs
+configured shutdown hooks.
+
+## Configuration
```
02strich, line 64: how do I configure the shutdown timing?

src/Temporalio.ApiDoc/api/index.md
```diff
@@ -16,5 +16,7 @@ Commonly used namespaces:
 Extensions:
 
 * [Temporalio.Extensions.DiagnosticSource](/api/Temporalio.Extensions.DiagnosticSource.html)
+* [Temporalio.Extensions.Aws.Lambda](/api/Temporalio.Extensions.Aws.Lambda.html)
```
jmaeagle99, line 18: Please alphabetize

.github/workflows/nuget-package.yml
```diff
@@ -171,6 +171,10 @@ jobs:
             src/Temporalio/bin/Release/*.snupkg
             src/Temporalio.Extensions.DiagnosticSource/bin/Release/*.nupkg
             src/Temporalio.Extensions.DiagnosticSource/bin/Release/*.snupkg
+            src/Temporalio.Extensions.Aws.Lambda/bin/Release/*.nupkg
```
jmaeagle99, line 188: Please alphabetize

src/Temporalio.ApiDoc/docfx.json
```diff
@@ -6,6 +6,8 @@
           "files": [
             "Temporalio/*.csproj",
             "Temporalio.Extensions.DiagnosticSource/*.csproj",
+            "Temporalio.Extensions.Aws.Lambda/*.csproj",
```
jmaeagle99, line 8: Please alphabetize

README.md
```diff
@@ -276,6 +276,26 @@ var client = await TemporalClient.ConnectAsync(new("my-namespace.a1b2c.tmprl.clo
 });
 ```
 
+#### Client Configuration From Environment
```
jmaeagle99, line 279: Please separate this out from this change and open a separate PR

src/Temporalio/Temporalio.csproj
```diff
@@ -28,6 +28,9 @@
     <AssemblyAttribute Include="System.Runtime.CompilerServices.InternalsVisibleTo">
       <_Parameter1>Temporalio.Tests</_Parameter1>
     </AssemblyAttribute>
+    <AssemblyAttribute Include="System.Runtime.CompilerServices.InternalsVisibleTo">
```
jmaeagle99, line 31: Please alphabetize (by `_Parameter1` element)

src/Temporalio/Bridge/EnvConfig.cs
```diff
@@ -24,7 +24,7 @@ internal static class EnvConfig
 
             try
             {
-                var envVarsRef = options.OverrideEnvVars?.Count > 0
+                var envVarsRef = options.OverrideEnvVars != null
```
jmaeagle99: Are the EnvConfig changes unrelated to AWS Lambda support? Even if not, I would probably still make these a separate change.

src/Temporalio.Extensions.Aws.Lambda/TemporalLambdaWorker.cs
```diff
@@ -0,0 +1,429 @@
...
+                {
+                    LoadClientConnectOptions = options => LoadClientConnectOptions(options),
+                });
+
+        /// <summary>
+        /// Load Temporal client connection options using AWS Lambda-aware config file resolution.
+        /// </summary>
+        /// <param name="options">Options for loading the configuration profile.</param>
+        /// <returns>Client connection options.</returns>
+        public static TemporalClientConnectOptions LoadClientConnectOptions(
```
jmaeagle99: This isn't a public API in the Go, Python, or TypeScript lambda workers. Does this need to be so? There might be a need to bring things up to parity in terms of API exposure and behavior.

src/Temporalio.Extensions.Aws.Lambda/LambdaWorkerConfig.cs
```diff
@@ -0,0 +1,95 @@
...
+using System.Threading.Tasks;
+using Temporalio.Client;
+using Temporalio.Worker;
+
+namespace Temporalio.Extensions.Aws.Lambda
+{
+    /// <summary>
+    /// Configuration for <see cref="TemporalLambdaWorker.CreateHandler(Temporalio.Common.WorkerDeploymentVersion, Action{LambdaWorkerConfig})" />.
+    /// </summary>
+    public class LambdaWorkerConfig
```
jmaeagle99: `TemporalLambdaWorkerOptions` to be more consistent with naming convention in the .NET SDK:
- It seems that top-level entry points are prefixed with `Temporal`
- Nearly all of these types of classes in the .NET SDK are suffixed with `Options`
- It should match the naming of the thing is is options of e.g. `TemporalLambdaWorker`

src/Temporalio.Extensions.Aws.Lambda/TemporalLambdaWorker.cs
```diff
@@ -0,0 +1,429 @@
...
+            WorkerDeploymentVersion version,
+            object configure,
+            string configureParamName,
+            TemporalLambdaWorkerHandlerOptions handlerOptions)
+        {
+            if (version == null)
+            {
+                throw new ArgumentNullException(nameof(version));
+            }
+            if (configure == null)
```
jmaeagle99, line 141: Can you leave a comment here as to why we are not using `CallerArgumentExpression` attribute on the `object configure` parameter? We'd need to bump the language version to 10+, which is something I'm thinking of doing so we can leverage some of the QOL. Would be similar to how we enabled `init` properties with the `IsExternalInit` [polly-fill](https://github.com/temporalio/sdk-dotnet/blob/main/src/Temporalio/Common/_LanguageHelpers.cs#L8).

src/Temporalio.Extensions.Aws.Lambda/LambdaWorkerConfig.cs
```diff
@@ -0,0 +1,95 @@
...
+    {
+        /// <summary>
+        /// Default time reserved after the worker run budget for worker shutdown and hooks.
+        /// </summary>
+        public static readonly TimeSpan DefaultShutdownDeadlineBuffer = TimeSpan.FromSeconds(7);
```
jmaeagle99: Can this be made `private` or `internal`? I don't think a consumer needs to programmatically access this default value.

src/Temporalio.Extensions.Aws.Lambda/LambdaWorkerConfig.cs
```diff
@@ -0,0 +1,95 @@
...
+        /// <summary>
+        /// Gets or sets the deadline buffer reserved for worker shutdown and hooks.
+        /// </summary>
+        public TimeSpan ShutdownDeadlineBuffer { get; set; } = DefaultShutdownDeadlineBuffer;
+
+        /// <summary>
+        /// Gets or sets hooks to run after each invocation's worker has shut down.
+        /// </summary>
+#pragma warning disable CA2227 // The public API intentionally allows replacing the list during configuration.
+        public IList<Func<CancellationToken, Task>> ShutdownHooks { get; set; } =
```
jmaeagle99: - I would remove the setter, to avoid accidentally dropping any already-existing hooks when setting a new list (which will now be disallowed). This eliminates the pragma.
- I would actually change the property type to `IReadOnlyCollection<T>` and use a private readonly `List<T>` backing field; return it from the property getter. Add a `AddShutdownHook(Func<T>)` method that (1) validates the hook is not null, and (2) adds it to the private backing field. This would eliminate the need for downstream item null checking and prevent resets or mutations of the collection.
- And to take it even further, if consumers don't need to inspect the hook list, just make it `internal`.

src/Temporalio.Extensions.Aws.Lambda/TemporalLambdaWorker.cs
```diff
@@ -0,0 +1,429 @@
...
+            var postPluginConfiguration = config.WorkerOptions.PostPluginConfiguration;
+            config.WorkerOptions.PostPluginConfiguration = options =>
+            {
+                postPluginConfiguration?.Invoke(options);
+                ApplyDeploymentVersion(options, version);
+                ClearConcurrencyLimitsIfTunerSet(options);
+            };
+            config.WorkerOptions.ApplyPostPluginConfiguration();
+
+            foreach (var hook in config.ShutdownHooks)
```
jmaeagle99: Becomes unnecessary if https://github.com/temporalio/sdk-dotnet/pull/721/changes#r3438966388 is adopted.

src/Temporalio.Extensions.Aws.Lambda/TemporalLambdaWorker.cs
```diff
@@ -0,0 +1,429 @@
...
+
+        /// <summary>
+        /// Create an AWS Lambda handler that runs a Temporal worker for each invocation.
+        /// </summary>
+        /// <param name="version">Worker deployment version for this Lambda worker.</param>
+        /// <param name="configure">Callback to configure client and worker options.</param>
+        /// <returns>A Lambda handler delegate.</returns>
+        public static Func<object?, ILambdaContext, Task> CreateHandler(
+            WorkerDeploymentVersion version,
+            Action<LambdaWorkerConfig> configure) =>
```
jmaeagle99: This should be invoked per-invocation as well. I don't think there should be a difference of call behavior just because one is sync and the other is async.

src/Temporalio.Extensions.Aws.Lambda/TemporalLambdaWorker.cs
```diff
@@ -0,0 +1,429 @@
...
+            }
+
+            var postPluginConfiguration = config.WorkerOptions.PostPluginConfiguration;
+            config.WorkerOptions.PostPluginConfiguration = options =>
+            {
+                postPluginConfiguration?.Invoke(options);
+                ApplyDeploymentVersion(options, version);
+                ClearConcurrencyLimitsIfTunerSet(options);
+            };
+            config.WorkerOptions.ApplyPostPluginConfiguration();
```
jmaeagle99: I think forcing the application here is a production smell (because it's doubly called: here and in the worker) and is only necessary for testing without a real worker. I think this should be removed and then have the affected tests simulate the worker behavior.

src/Temporalio/Worker/TemporalWorkerOptions.cs
```diff
@@ -376,6 +376,14 @@ public TemporalWorkerOptions()
         internal Func<WorkflowInstanceDetails, IWorkflowInstance> WorkflowInstanceFactory { get; set; } =
             DefaultWorkflowInstanceFactory;
 
+        /// <summary>
+        /// Gets or sets a function to run after worker plugins configure options.
+        /// </summary>
+        /// <remarks>
+        /// Don't expose this until there's a use case.
+        /// </remarks>
+        internal Action<TemporalWorkerOptions>? PostPluginConfiguration { get; set; }
```
jmaeagle99: Instead of this, could we write our own internal plugin to do it? Would eliminate this new surface area, guarantee last application, show up in worker heartbeat, etc.

jmaeagle99 (commented):
Wanted to provide some feedback; still reviewing.

src/Temporalio.Extensions.Aws.Lambda/TemporalLambdaWorker.cs
```diff
@@ -0,0 +1,437 @@
...
+        {
+            if (options.OverrideEnvVars != null)
+            {
+                return options.OverrideEnvVars.TryGetValue(name, out var value) ? value : null;
+            }
+
+            return Environment.GetEnvironmentVariable(name);
+        }
+
+        private static void LogLine(ILambdaContext context, string message) =>
```
jmaeagle99, line 288: I would replace this utility method with direct calls to `context.Logger?.LogX()` where X is the appropriate log level e.g. `LogWarning` and `LogError`. Remove the "WARNING: " and "ERROR: " prefixes from the message.

src/Temporalio.Extensions.Aws.Lambda/TemporalLambdaWorkerHandlerOptions.cs
```diff
@@ -0,0 +1,40 @@
...
+        /// </summary>
+        public Func<string, string?> GetEnvironmentVariable { get; set; } =
+            Environment.GetEnvironmentVariable;
+
+        /// <summary>
+        /// Gets or sets the client configuration loader.
+        /// </summary>
+        public Func<ClientEnvConfig.ProfileLoadOptions?, TemporalClientConnectOptions>?
+            LoadClientConnectOptions
+        { get; set; }
```
jmaeagle99: nit: This looks strange being on its own line. Please put it on the same line as the property name.

tests/Temporalio.Tests/Common/EnvConfig/ClientConfigTests.cs
```diff
@@ -287,6 +287,32 @@ public void Test_Load_Profile_Disable_Env()
             Assert.Equal("default-address", profile.Address);
         }
 
+        [Fact]
+        public void EmptyOverrideEnvVarsSuppressesSystemEnvironment()
```
jmaeagle99: This probably should be moved to https://github.com/temporalio/sdk-dotnet/pull/767

tests/Temporalio.Tests/Common/EnvConfig/ClientConfigTests.cs
```diff
@@ -287,6 +287,32 @@ public void Test_Load_Profile_Disable_Env()
...
+            var previousNamespace = Environment.GetEnvironmentVariable("TEMPORAL_NAMESPACE");
+            try
+            {
+                Environment.SetEnvironmentVariable("TEMPORAL_ADDRESS", "system-address");
+                Environment.SetEnvironmentVariable("TEMPORAL_NAMESPACE", "system-namespace");
+
+                var profile = ClientEnvConfig.Profile.Load(new ClientEnvConfig.ProfileLoadOptions
+                {
+                    DisableFile = true,
+                    OverrideEnvVars = new Dictionary<string, string>(),
```
jmaeagle99: Does this test pass if you remove the `OverrideEnvVars` property setting? I think you'll run into what I described in https://github.com/temporalio/sdk-dotnet/pull/767#discussion_r3539715194

src/Temporalio/Common/EnvConfig/ClientEnvConfig.cs
```diff
@@ -206,7 +206,7 @@ public sealed record Tls(
             {
                 if (Disabled == true)
                 {
-                    return null;
+                    return new TlsOptions { Disabled = true };
```
jmaeagle99: This is probably another change that should be in a separate PR. Feels like its fixing behavioral issue that is separate from Lambda itself, even it if Lambda exhibits the problem.

tests/Temporalio.Tests/Extensions/Aws/Lambda/TemporalLambdaWorkerNonParallelDefinition.cs
```diff
@@ -0,0 +1,8 @@
+namespace Temporalio.Tests.Extensions.Aws.Lambda;
+
+using Xunit;
+
+[CollectionDefinition("TemporalLambdaWorkerNonParallel", DisableParallelization = true)]
+public sealed class TemporalLambdaWorkerNonParallelDefinition
+{
```
jmaeagle99, line 7: I would recommend turning "TemporalLambdaWorkerNonParallel" into a public const on this class and consuming it in the attribute usage and on the test class to avoid spelling mistakes and make it easier to discover to which definition it is associated.

src/Temporalio.Extensions.Aws.Lambda.OpenTelemetry/LambdaWorkerOpenTelemetry.cs
```diff
@@ -0,0 +1,204 @@
...
+        /// Configure OpenTelemetry metrics and tracing with AWS Lambda defaults.
+        /// </summary>
+        /// <param name="config">Lambda worker configuration to mutate.</param>
+        /// <param name="options">Optional OpenTelemetry configuration.</param>
+        /// <remarks>
+        /// This creates an OTLP trace exporter and tracer provider, configures Core SDK metrics
+        /// through a Temporal runtime, adds the Temporal tracing interceptor, and registers a
+        /// per-invocation shutdown hook to force-flush traces before the Lambda invocation ends.
+        /// </remarks>
+        public static void ApplyDefaults(
```
jmaeagle99, line 43: This method is called `ApplyDefaults` but it doing more than just applying defaults due to the setting of the runtime. Whatever runtime that was set by the caller is gone. I think we should, at the very least, document this. Maybe consider logging a warning (somehow, I know a logger is not plumbed through at this time) or raising an exception.

src/Temporalio.Extensions.Aws.Lambda.OpenTelemetry/LambdaWorkerOpenTelemetry.cs
```diff
@@ -0,0 +1,204 @@
...
+            var resolvedOptions = ResolveOptions(options);
+#pragma warning disable CA2000 // Provider is intentionally retained for Lambda warm invocations.
+            var tracerProvider = CreateTracerProvider(resolvedOptions);
+#pragma warning restore CA2000
+
+            config.ClientOptions.Interceptors = AddTracingInterceptor(
+                config.ClientOptions.Interceptors);
+            config.ClientOptions.Runtime = CreateRuntime(resolvedOptions);
+            config.AddShutdownHook(
+                cancellationToken => ForceFlushAsync(
```
jmaeagle99: Should this callback dispose of the provider after the flush?

src/Temporalio.Extensions.Aws.Lambda.OpenTelemetry/LambdaWorkerOpenTelemetry.cs
```diff
@@ -0,0 +1,204 @@
...
+            if (cancellationToken.IsCancellationRequested)
+            {
+                return;
+            }
+
+            var flushTask = Task.Run(
+                () => tracerProvider.ForceFlush(ToTimeoutMilliseconds(shutdownDeadlineBuffer)));
+            if (flushTask == await Task.WhenAny(
+                flushTask,
+                Task.Delay(Timeout.Infinite, cancellationToken)).ConfigureAwait(false))
```
jmaeagle99: nit: The infinitely delayed task dangles for the lifetime of the CancellationTokenSource since the task never completes successfully.

A _better_ way to do this is something like:

```csharp
public static async Task WaitAsync(this Task task, CancellationToken cancellationToken)
{
    if (task == null)
    {
        throw new ArgumentNullException(nameof(task));
    }
    if (task.IsCompleted || !cancellationToken.CanBeCanceled)
    {
        await task.ConfigureAwait(false);
        return;
    }
    cancellationToken.ThrowIfCancellationRequested();

    TaskCompletionSource<bool> cancellationTaskSource = new TaskCompletionSource<bool>(TaskCreationOptions.RunContinuationsAsynchronously);
    using (cancellationToken.Register(() => cancellationTaskSource.TrySetResult(true)))
    {
        if (task != await Task.WhenAny(task, cancellationTaskSource.Task).ConfigureAwait(false))
        {
            throw new OperationCanceledException(cancellationToken);
        }
    }
    
    await task.ConfigureAwait(false);
}
```

This allows the cancellation registration to be disposed when it's not really needed anymore. It also eliminates the need for the separate observation of the flush vs the cancellation task.

Newer versions of modern .NET have [this](https://learn.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.waitasync?view=net-6.0) built into the `Task` class already but we aren't there yet and we'd need to polyfill the other runtimes anyway.

Would suggest moving to a separate `TaskExtensions.cs` file.

src/Temporalio.Extensions.Aws.Lambda.OpenTelemetry/LambdaWorkerOpenTelemetry.cs
```diff
@@ -0,0 +1,204 @@
...
+            else
+            {
+                ObserveTaskException(flushTask);
+            }
+        }
+
+        private static string FirstNonEmpty(params string?[] values) =>
+            values.First(value => !string.IsNullOrEmpty(value))!;
+
+        private static void ObserveTaskException(Task task) =>
```
jmaeagle99: This is usually called `Forget`, is an extension method, and can go right next to the `WaitAsync` extension method. Regardless, you'll want to check that `task` is not null and raise `ArgumentNullException` when it is.

jmaeagle99 (commented):
Some more feedback. I think once these are addressed, this PR will be in good shape to merge.

src/Temporalio.Extensions.Aws.Lambda/TemporalLambdaWorker.cs
eamsden: This would require moving the lambda package from targeting `netstandard2.0` (in line with the rest of the SDK) to at least `net8.0`. We could do that, but I want to call this out explicitly. `ILambdaLogger` at `netstandard2.0` only exposes `Log` and `LogLine`.

jmaeagle99: One possibility is to have both a `netstandard2.0` and `net8.0` target. But then we'd have to rewrite the call twice, one for each target. Probably not worth it until we have more reasons to have a `net8.0` or higher target. So I'll suggest we leave as-is and adjust in the future if it makes sense.

src/Temporalio.Extensions.Aws.Lambda.OpenTelemetry/LambdaWorkerOpenTelemetry.cs
eamsden: Yes, done

src/Temporalio.Extensions.Aws.Lambda/TemporalLambdaWorkerHandlerOptions.cs
jmaeagle99: Sorry, I could have been more clear. It's the `{ get; set; }` being on its own line that looks strange.

CHANGELOG.md
```diff
@@ -23,6 +23,15 @@ to docs, or any other relevant information.
 
 - Fixed `ClientEnvConfig` empty `OverrideEnvVars` handling so an explicit empty dictionary no
   longer falls back to process environment variables.
+### Breaking Changes
```
jmaeagle99: Neither of these changes are accurate. All of this code is new so it cannot be breaking nor changed. There should be one entry in Added that describes the creation of these Lambda support packages.

Directory.Packages.props
```diff
@@ -15,8 +16,11 @@
     <PackageVersion Include="Microsoft.SourceLink.GitHub" Version="8.0.0" />
     <PackageVersion Include="Microsoft.VisualStudio.Threading.Analyzers" Version="18.7.23" />
     <PackageVersion Include="NexusRpc" Version="0.3.0" />
+    <PackageVersion Include="OpenTelemetry" Version="1.15.3" />
```
jmaeagle99, line 19: Can you add the new `OpenTelemetry.*` packages to the ignore group in https://github.com/temporalio/sdk-dotnet/blob/main/.github/dependabot.yml#L68 so that way we don't get constantly pestered to update but still get security updates.

src/Temporalio.Extensions.Aws.Lambda/TemporalLambdaWorkerHandlerOptions.cs
```diff
@@ -0,0 +1,39 @@
...
+        /// Gets or sets the environment variable reader.
+        /// </summary>
+        public Func<string, string?> GetEnvironmentVariable { get; set; } =
+            Environment.GetEnvironmentVariable;
+
+        /// <summary>
+        /// Gets or sets the client configuration loader.
+        /// </summary>
+        public Func<ClientEnvConfig.ProfileLoadOptions?, TemporalClientConnectOptions>? LoadClientConnectOptions
+        { get; set; }
```
jmaeagle99: Sorry, I could have been more clear. It's the `{ get; set; }` being on its own line that looks strange.

jmaeagle99 (approved):
Just a few more, but approving.

CHANGELOG.md
eamsden: I missed that the AI had updated that. 😳

PR #731, opened 2026-06-06
feat: gzip compression options

jmaeagle99:
## What was changed

Add options to configure gRPC compression. Default to gzip compression.

## Why?

Options good!

## Checklist
1. How was this tested: Simple translation unit tests. Rely on sdk-core and sdk-core-c-bridge testing.


src/Temporalio/Client/GrpcCompression.cs
```diff
@@ -0,0 +1,19 @@
...
+    /// <summary>
+    /// Transport-level gRPC compression used for client calls.
+    /// </summary>
+    /// <seealso cref="TemporalConnectionOptions.GrpcCompression" />
+    public enum GrpcCompression
+    {
+        /// <summary>
+        /// Do not compress requests or advertise acceptance of compressed responses.
+        /// </summary>
+        None,
```
Sushisource: I think we need to use something other than enum here - in the other PRs I wanted to keep open the possibility of specifying information along with the compression kind in a backwards compat way. IE: `Zstd { level: 3 }` or something, even though we don't have that yet.

jmaeagle99: Refactored to be a discriminated union-like closed type set.

Sushisource (approved):

PR #741, opened 2026-06-11
Adding links to Nexus signals

Evanthx:
This adds links to Nexus signals - the receiving workflow has a link to the workflow sending the signal. and the sending workflow has a link to the receiving workflow.

### DIFF
--- a/src/Temporalio/Client/TemporalClient.Workflow.cs
+++ b/src/Temporalio/Client/TemporalClient.Workflow.cs
@@ -369,7 +369,7 @@ namespace Temporalio.Client
                         }
                     }
                 }
-                ForwardNexusLinks(req.Links);
+                ForwardNexusRequestLinks(req.Links);
                 var resp = await Client.Connection.WorkflowService.SignalWorkflowExecutionAsync(
                     req, DefaultRetryOptions(input.Options?.Rpc)).ConfigureAwait(false);
                 CaptureNexusResponseLink(resp.Link);
@@ -766,7 +766,7 @@ namespace Temporalio.Client
                         throw new ArgumentException("Cannot have start signal args without start signal");
                     }
                     // A plain start does not produce a response link.
-                    ForwardNexusLinks(req.Links);
+                    ForwardNexusRequestLinks(req.Links);
                     var resp = await Client.Connection.WorkflowService.StartWorkflowExecutionAsync(
                         req, DefaultRetryOptions(input.Options.Rpc)).ConfigureAwait(false);
                     return new WorkflowHandle<TWorkflow, TResult>(
@@ -809,7 +809,7 @@ namespace Temporalio.Client
                     signalReq.SignalInput.Payloads_.AddRange(
                         await dataConverter.ToPayloadsAsync(input.Options.StartSignalArgs).ConfigureAwait(false));
                 }
-                ForwardNexusLinks(signalReq.Links);
+                ForwardNexusRequestLinks(signalReq.Links);
                 var signalResp = await Client.Connection.WorkflowService.SignalWithStartWorkflowExecutionAsync(
                     signalReq, DefaultRetryOptions(input.Options.Rpc)).ConfigureAwait(false);
                 CaptureNexusResponseLink(signalResp.SignalLink);
--- a/src/Temporalio/Client/TemporalClient.Workflow.cs
+++ b/src/Temporalio/Client/TemporalClient.Workflow.cs
@@ -752,37 +752,15 @@ namespace Temporalio.Client
                     {
                         throw new ArgumentException("Cannot have start signal args without start signal");
                     }
-                    var resp = await Client.Connection.WorkflowService.StartWorkflowExecutionAsync(
-                        req, DefaultRetryOptions(input.Options.Rpc)).ConfigureAwait(false);
+                    // If this start is being issued from inside a Nexus operation handler, forward
+                    // the inbound Nexus task links so the callee's WorkflowExecutionStarted history
+                    // event links back to the caller. A plain start does not produce a backlink.
                     if (NexusOperationExecutionContext.HasCurrent)
                     {
-                        // Auto-capture the start-workflow backlink so the task handler drains it onto
-                        // the StartOperationResponse, the same path used for signal/signalWithStart
-                        // responses.
-                        if (resp.Link != null)
-                        {
-                            NexusOperationExecutionContext.Current.AddBacklink(resp.Link);
-                        }
-                        else
-                        {
-                            // Older servers (pre-1.31) don't return a link on the start response.
-                            // Fabricate one pointing at the started workflow's
-                            // WorkflowExecutionStarted event so the caller still gets a backlink.
-                            NexusOperationExecutionContext.Current.AddBacklink(new Api.Common.V1.Link
-                            {
-                                WorkflowEvent = new()
-                                {
-                                    Namespace = Client.Options.Namespace,
-                                    WorkflowId = req.WorkflowId,
-                                    RunId = resp.RunId,
-                                    EventRef = new()
-                                    {
-                                        EventType = Api.Enums.V1.EventType.WorkflowExecutionStarted,
-                                    },
-                                },
-                            });
-                        }
+                        req.Links.AddRange(NexusOperationExecutionContext.Current.NexusOperationLinks);
                     }
+                    var resp = await Client.Connection.WorkflowService.StartWorkflowExecutionAsync(
+                        req, DefaultRetryOptions(input.Options.Rpc)).ConfigureAwait(false);
                     return new WorkflowHandle<TWorkflow, TResult>(
                         Client: Client,
                         Id: req.WorkflowId,
--- a/src/Temporalio/Client/TemporalClient.Workflow.cs
+++ b/src/Temporalio/Client/TemporalClient.Workflow.cs
@@ -372,7 +372,7 @@ namespace Temporalio.Client
                 ForwardNexusLinks(req.Links);
                 var resp = await Client.Connection.WorkflowService.SignalWorkflowExecutionAsync(
                     req, DefaultRetryOptions(input.Options?.Rpc)).ConfigureAwait(false);
-                CaptureNexusBacklink(resp.Link);
+                CaptureNexusResponseLink(resp.Link);
             }
 
             /// <inheritdoc />
@@ -686,20 +686,20 @@ namespace Temporalio.Client
             {
                 if (NexusOperationExecutionContext.HasCurrent)
                 {
-                    links.AddRange(NexusOperationExecutionContext.Current.ForwardLinks);
+                    links.AddRange(NexusOperationExecutionContext.Current.RequestLinks);
                 }
             }
 
-            // Capture a backlink returned by an outbound RPC (signal / signalWithStart) onto the
-            // current Nexus operation context so the task handler can attach it to the operation's
-            // StartOperationResponse. No-op outside a Nexus operation context; null/unset links are
-            // dropped by TryAddBacklink (servers older than 1.31, or without
+            // Capture a response link returned by an outbound RPC (signal / signalWithStart) onto
+            // the current Nexus operation context so the task handler can attach it to the
+            // operation's StartOperationResponse. No-op outside a Nexus operation context; null/unset
+            // links are dropped by TryAddResponseLink (servers older than 1.31, or without
             // EnableCHASMSignalBacklinks, leave the response link unset).
-            private static void CaptureNexusBacklink(Link? backlink)
+            private static void CaptureNexusResponseLink(Link? responseLink)
             {
                 if (NexusOperationExecutionContext.HasCurrent)
                 {
-                    NexusOperationExecutionContext.Current.TryAddBacklink(backlink);
+                    NexusOperationExecutionContext.Current.TryAddResponseLink(responseLink);
                 }
             }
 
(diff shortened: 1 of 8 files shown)


src/Temporalio/Client/TemporalClient.Workflow.cs
```diff
@@ -367,8 +368,22 @@ public override async Task SignalWorkflowAsync(SignalWorkflowInput input)
                         }
                     }
                 }
-                await Client.Connection.WorkflowService.SignalWorkflowExecutionAsync(
+                // If this signal is being issued from inside a Nexus operation handler, forward the
```
jmaeagle99: Would all of the forward and backlink adding specifically for Nexus be better served as a client interceptor? Might help centralize the concerns rather than wrapping every callsite.

tests/Temporalio.Tests/packages.lock.json
```diff
@@ -356,7 +356,8 @@
         "resolved": "17.4.1",
```
jmaeagle99, line 356: Please revert all changes in this file. I don't think they are necessary for this PR.

src/Temporalio/Worker/NexusWorker.cs
```diff
@@ -199,14 +199,46 @@ private async Task<StartOperationResponse> HandleStartOperationAsync(
...
+                // caller workflow's history event links to each event on the callee.
+                var backlinks = executionContext.ResponseBacklinks
+                    .Where(l => l.VariantCase == Api.Common.V1.Link.VariantOneofCase.WorkflowEvent)
+                    .Select(l => l.WorkflowEvent.ToNexusLink())
+                    .Select(l => new Api.Nexus.V1.Link() { Type = l.Type, Url = l.Uri.ToString(), })
+                    .ToList();
                 var links = context.OutboundLinks.Select(l =>
-                    new Api.Nexus.V1.Link() { Type = l.Type, Url = l.Uri.ToString(), });
+                    new Api.Nexus.V1.Link() { Type = l.Type, Url = l.Uri.ToString(), })
+                    .Concat(backlinks);
```
jmaeagle99: I would concatenate the `NexusLink` enumerables first and do one final `.Select(l => new Api.Nexus.V1.Link() { ... })` call.

src/Temporalio/Worker/NexusWorker.cs
```diff
@@ -199,14 +199,46 @@ private async Task<StartOperationResponse> HandleStartOperationAsync(
...
             try
             {
                 var result = await handler.StartOperationAsync(
                     context,
                     new HandlerContent(startOp.Payload.ToByteArray())).ConfigureAwait(false);
+                // Drain any backlinks captured from outbound RPCs the handler issued (e.g. signal,
+                // signalWithStart, start) and attach them to both the sync and async response so the
+                // caller workflow's history event links to each event on the callee.
+                var backlinks = executionContext.ResponseBacklinks
+                    .Where(l => l.VariantCase == Api.Common.V1.Link.VariantOneofCase.WorkflowEvent)
```
jmaeagle99: Is this filter necessary since `AddBacklink` is already filtering for this case?

src/Temporalio/Nexus/NexusOperationExecutionContext.cs
```diff
@@ -96,5 +114,50 @@ internal NexusOperationExecutionContext(
...
+            }
+        }
+
+        /// <summary>
+        /// Append a backlink returned by an outbound RPC the operation handler issued (e.g. signal,
+        /// signalWithStart, start). The task handler drains the list when building the operation's
+        /// StartOperationResponse. Null and non-WorkflowEvent links are ignored.
+        /// </summary>
+        /// <param name="link">Backlink to add.</param>
+        internal void AddBacklink(Api.Common.V1.Link? link)
```
jmaeagle99: Maybe `bool TryAddBacklink(Link? link)` instead so that it's more clear than nulls are dropped and then the callers don't need to null check the link before passing.

src/Temporalio/Nexus/NexusOperationExecutionContext.cs
```diff
@@ -18,6 +19,23 @@ public class NexusOperationExecutionContext
...
 
+        // Backlinks returned by outbound RPCs the operation handler issues (such as
+        // SignalWorkflowExecutionResponse.Link or SignalWithStartWorkflowExecutionResponse.SignalLink).
+        // One entry per outbound RPC that returned a link. Drained by the task handler when building
+        // the StartOperationResponse so each RPC the handler issued gets a corresponding link on the
+        // caller workflow's history event.
+        //
+        // Access is synchronized via backlinksLock: AddBacklink appends under the lock and the
+        // ResponseBacklinks getter returns a copy under the lock.
+        private readonly List<Api.Common.V1.Link> responseBacklinks = new();
```
jmaeagle99: Consider using `ConcurrentQueue<T>` instead of `List<T>` so you can drop the dedicated lock object.

src/Temporalio/Nexus/NexusOperationExecutionContext.cs
```diff
@@ -96,5 +114,50 @@ internal NexusOperationExecutionContext(
...
+
+        /// <summary>
+        /// Gets or sets the <c>common.v1.Link</c>s extracted from the inbound Nexus task so they can
+        /// be attached to RPCs issued by the operation handler. Only WorkflowEvent-shaped links are
+        /// stored; empty if none.
+        /// </summary>
+        internal IReadOnlyCollection<Api.Common.V1.Link> NexusOperationLinks
+        {
+            get => nexusOperationLinks;
+            set => nexusOperationLinks = value ?? Array.Empty<Api.Common.V1.Link>();
```
jmaeagle99: I don't think the null coalescing is necessary since the type is not nullable. Also, the only caller is setting a non-null value. Consider simplifying to a standard auto property.

src/Temporalio/Nexus/NexusOperationExecutionContext.cs
```diff
@@ -96,5 +114,50 @@ internal NexusOperationExecutionContext(
         /// Gets the Nexus context.
         /// </summary>
         internal OperationContext HandlerContext { get; private init; }
+
+        /// <summary>
+        /// Gets or sets the <c>common.v1.Link</c>s extracted from the inbound Nexus task so they can
+        /// be attached to RPCs issued by the operation handler. Only WorkflowEvent-shaped links are
+        /// stored; empty if none.
+        /// </summary>
+        internal IReadOnlyCollection<Api.Common.V1.Link> NexusOperationLinks
```
jmaeagle99: This and `ResponseBacklinks` appear to be symmetrical. Consider naming them symmetrically e.g. `ForwardLinks` and `Backlinks` or `RequestLinks` and `ResponseLinks`. I think you can drop `Nexus` and `Operation` from the names since all of this is on `NexusOperationExecutionContext` and should be understood implicitly that they are for Nexus operations.

tests/Temporalio.Tests/packages.lock.json
Evanthx: Reverted!

src/Temporalio/Worker/NexusWorker.cs
Evanthx: Good catch!

src/Temporalio/Worker/NexusWorker.cs
Evanthx: Done

src/Temporalio/Nexus/NexusOperationExecutionContext.cs
Evanthx: Good idea - done

src/Temporalio/Nexus/NexusOperationExecutionContext.cs
Evanthx: Done

src/Temporalio/Nexus/NexusOperationExecutionContext.cs
Evanthx: Done

src/Temporalio/Nexus/NexusOperationExecutionContext.cs
Evanthx: Renamed to Request and Response links, and propagated that change to the the corresponding PRs for the other SDKs.

src/Temporalio/Client/TemporalClient.Workflow.cs
Evanthx: This is an interesting idea! I think I want to push back on it though - it is modifying req.Links which is in protoBuf, and i don't think interceptors can access that?

But it did make me think about the wrapping of every callsite, and I broke that out into a private method, so hopefully that is a lot simpler now

jmaeagle99: Yeah, totally fine. Didn't investigate whether interceptors could fully work here, but just wanted to provide another idea.

jmaeagle99 (approved):

src/Temporalio/Client/TemporalClient.Workflow.cs
```diff
@@ -675,6 +679,30 @@ public override async Task<WorkflowListPage> ListWorkflowsPaginatedAsync(ListWor
                     NextPageToken: resp.NextPageToken.IsEmpty ? null : resp.NextPageToken.ToByteArray());
             }
 
+            // Forward the inbound Nexus task links onto an outbound request when this call is issued
+            // from inside a Nexus operation handler, so the resulting history event links back to
+            // the caller. No-op outside a Nexus operation context.
+            private static void ForwardNexusLinks(RepeatedField<Link> links)
```
jmaeagle99:
```suggestion
            private static void ForwardNexusRequestLinks(RepeatedField<Link> links)
```

jmaeagle99:
Also, please update https://github.com/temporalio/sdk-dotnet/blob/main/CHANGELOG.md with a blurb about this change that will be used in release notes.

src/Temporalio/Nexus/NexusOperationExecutionContext.cs
```diff
@@ -18,6 +20,14 @@ public class NexusOperationExecutionContext
         private readonly Lazy<MetricMeter> metricMeter;
         private readonly ITemporalClient? temporalClient;
 
+        // Response links returned by outbound RPCs the operation handler issues (such as
+        // SignalWorkflowExecutionResponse.Link or SignalWithStartWorkflowExecutionResponse.SignalLink).
+        // One entry per outbound RPC that returned a link. Drained by the task handler when building
+        // the StartOperationResponse so each RPC the handler issued gets a corresponding link on the
+        // caller workflow's history event. A ConcurrentQueue handles enqueues from outbound RPCs the
+        // handler may issue in parallel; the ResponseLinks getter snapshots it.
+        private readonly ConcurrentQueue<Api.Common.V1.Link> responseLinks = new();
```
Quinn-With-Two-Ns, line 29: Why did we need to add `responseLinks` when we already have `OutboundLinks` on `OperationContext`

src/Temporalio/Nexus/NexusOperationExecutionContext.cs
```diff
@@ -96,5 +106,39 @@ internal NexusOperationExecutionContext(
...
+        /// signalWithStart, start). The task handler drains the accumulated links when building the
+        /// operation's StartOperationResponse. Null and non-WorkflowEvent links are ignored.
+        /// </summary>
+        /// <param name="link">Response link to add.</param>
+        /// <returns><c>true</c> if the link was added; <c>false</c> if it was null or not a
+        /// WorkflowEvent-shaped link.</returns>
+        internal bool TryAddResponseLink(Api.Common.V1.Link? link)
+        {
+            if (link == null ||
+                link.VariantCase != Api.Common.V1.Link.VariantOneofCase.WorkflowEvent)
```
Quinn-With-Two-Ns, line 136: Why are we restricting these links to only WorkflowEvents?

Quinn-With-Two-Ns: I would expect this code to be generics for links so we can use it for all types of links we may get from a response

Evanthx: The protobufs have a NexusLink and methods to convert a workflow event to a Nexus link and back (in ProtoLinkExtensions). But that does restrict me to Workflow events unless we write more converters and make a bigger change - but then that would also apply to the other SDKs, so would be a non-minor change. I'd ask for it to be in a separate ticket and set of PRs if we decided to do that.

Quinn-With-Two-Ns: Then I'd would document that this function does drop unknown link types

src/Temporalio/Nexus/NexusOperationExecutionContext.cs
Evanthx: Honestly just for thread safety. If I used the OutboundLinks directly then I would need to wrap all the adds in a lock, and while that would protect any adds inside this file, that means there are now multiple threads adding things to OutboundLinks so THAT starts to need protecting and it gets weird. And OutboundLinks is brought in from Nexus so I can't just change that to a concurrent list, which would be the simpler answer.

If my understanding of this doesn't seem right to you let me know, but that is the reasoning behind it!

Quinn-With-Two-Ns: hm IMO if the `OutboundLinks` on `OperationContext` are not thread safe then we should fix that, we own the Nexus SDK and can change it. If you don't want to fix that before this PR, i'd at least open an issue and link to that here to explain why we don't use the "correct" approach

src/Temporalio/Nexus/NexusOperationExecutionContext.cs
Evanthx: /// <returns><c>true</c> if the link was added; <c>false</c> if it was null or not a
        /// WorkflowEvent-shaped link.</returns>
        
        That work?
        
        Added https://temporalio.atlassian.net/browse/NEXUS-468 to make this work for all links later

jmaeagle99:
@Evanthx the acceptance of [this suggestion](https://github.com/temporalio/sdk-dotnet/pull/741#discussion_r3417292028) required the callers to be updated as well.

PR #743, opened 2026-06-12
Add Standalone Nexus Operation links

Quinn-With-Two-Ns:
## What was changed

Add Standalone nexus operation links

## Why?

So SANO has nice links in the UI

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

---

> [!NOTE]
> **Medium Risk**
> Changes affect Nexus link propagation across worker, workflow start, and client RPC paths; incorrect conversion could break UI cross-links, though invalid links are dropped with warnings and coverage is expanded.
> 
> **Overview**
> Extends Nexus **link propagation** so standalone Nexus operations can round-trip both **workflow event** and **nexus operation** `common.v1.Link` variants, not only workflow-event-shaped links.
> 
> `ProtoLinkExtensions` gains bidirectional conversion (`ToProtoLink` / `ToNexusLink` on proto and `NexusLink`), nexus-operation URI encoding, shared path validation, and **canonical `temporal:///…` URIs** (replacing `UriBuilder` behavior that produced a non-standard single-slash form). Inbound links on workflow starts and Nexus tasks use `ToProtoLink()`; unsupported shapes are **logged and dropped** at conversion time.
> 
> `NexusOperationExecutionContext.TryAddResponseLink` now enqueues any non-null proto link; `NexusWorker` drains response links through `ToNexusLink()` and drops variants it cannot convert. Integration coverage asserts workflow-backed operations expose a workflow-event link on the operation and a **nexus-operation backlink** on the started workflow’s completion callback. Test env adds permissive **callback allowed addresses** for arbitrary callback URLs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a1493b2e3ee123b3f0679d4020bb18fc41ad12f3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/Temporalio/Nexus/NexusWorkflowStartHelper.cs
+++ b/src/Temporalio/Nexus/NexusWorkflowStartHelper.cs
@@ -64,6 +64,10 @@ namespace Temporalio.Nexus
                 {
                     try
                     {
+                        if (link.Type == Api.Common.V1.Link.Types.NexusOperation.Descriptor.FullName)
+                        {
+                            return new Link { NexusOperation = link.ToNexusOperationLink() };
+                        }
                         return new Link { WorkflowEvent = link.ToWorkflowEvent() };
                     }
                     catch (ArgumentException e)
--- a/src/Temporalio/Nexus/NexusWorkflowStartHelper.cs
+++ b/src/Temporalio/Nexus/NexusWorkflowStartHelper.cs
@@ -64,11 +64,7 @@ namespace Temporalio.Nexus
                 {
                     try
                     {
-                        if (link.Type == Api.Common.V1.Link.Types.NexusOperation.Descriptor.FullName)
-                        {
-                            return new Link { NexusOperation = link.ToNexusOperationLink() };
-                        }
-                        return new Link { WorkflowEvent = link.ToWorkflowEvent() };
+                        return link.ToProtoLink();
                     }
                     catch (ArgumentException e)
                     {
--- a/src/Temporalio/Nexus/ProtoLinkExtensions.cs
+++ b/src/Temporalio/Nexus/ProtoLinkExtensions.cs
@@ -86,6 +86,38 @@ namespace Temporalio.Nexus
         }
 
         /// <summary>
+        /// Convert a proto Link to a Nexus link, dispatching on the populated oneof variant.
+        /// </summary>
+        /// <param name="link">Proto link.</param>
+        /// <returns>Nexus link.</returns>
+        /// <exception cref="ArgumentException">If the link variant is not set or unknown.</exception>
+        public static NexusLink ToNexusLink(this Api.Common.V1.Link link) => link.VariantCase switch
+        {
+            Api.Common.V1.Link.VariantOneofCase.WorkflowEvent => link.WorkflowEvent.ToNexusLink(),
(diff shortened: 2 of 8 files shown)


src/Temporalio/Nexus/ProtoLinkExtensions.cs
```diff
@@ -57,17 +57,64 @@ public static NexusLink ToNexusLink(this Api.Common.V1.Link.Types.WorkflowEvent
...
+            return new(new Uri(uriStr), Api.Common.V1.Link.Types.NexusOperation.Descriptor.FullName);
+        }
+
+        /// <summary>
+        /// Convert a Nexus link to a nexus operation link.
+        /// </summary>
+        /// <param name="link">Nexus link.</param>
+        /// <returns>Nexus operation link.</returns>
+        /// <exception cref="ArgumentException">If the link is invalid.</exception>
+        public static Api.Common.V1.Link.Types.NexusOperation ToNexusOperationLink(this NexusLink link)
```
jmaeagle99: To be consistent with the other naming of similar methods, should be called `ToNexusOperation`.

src/Temporalio/Nexus/ProtoLinkExtensions.cs
```diff
@@ -57,17 +57,64 @@ public static NexusLink ToNexusLink(this Api.Common.V1.Link.Types.WorkflowEvent
                 queryParams["requestID"] = reqIdRef.RequestId;
             }
 
-            // Build URI
-            var builder = new UriBuilder
+            // Build URI with empty authority so there is no host. UriBuilder cannot be used here
```
jmaeagle99: `UriBuilder` does support empty host, you just have to manually specify that. But I suspect you won't get the triple forward slash (I think it gets a single forward slash). So the comment is a bit misleading. If the triple forward slash is not necessary, I'd switch back to `UriBuilder`, but if it is needed, then what you have is fine.

src/Temporalio/Nexus/NexusWorkflowStartHelper.cs
```diff
@@ -64,6 +64,10 @@ internal static async Task<NexusWorkflowRunHandle> StartWorkflowAsync(
                 {
                     try
                     {
+                        if (link.Type == Api.Common.V1.Link.Types.NexusOperation.Descriptor.FullName)
+                        {
+                            return new Link { NexusOperation = link.ToNexusOperationLink() };
```
jmaeagle99: Can you add a test that flexes this integration?

src/Temporalio/Nexus/ProtoLinkExtensions.cs
Quinn-With-Two-Ns: I updated the comment, let me know if that is clearer

src/Temporalio/Nexus/NexusWorkflowStartHelper.cs
Quinn-With-Two-Ns: Done, I updated `StartNexusOperationAsync_WorkflowBacked_OutboundLinkPointsToStartedWorkflow`

jmaeagle99 (approved):

PR #748, opened 2026-06-16
Add Standalone Activities support to Temporal Nexus Operation Handler

Quinn-With-Two-Ns:
## What was changed

Add Standalone Activities support to Temporal Nexus Operation Handler

## Why?

Allow calling Standalone Activities as a Nexus Operation

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

---

> [!NOTE]
> **Medium Risk**
> Touches experimental Nexus operation lifecycle (tokens, callbacks, cancel, link propagation) and extends standalone activity client RPC shaping; behavior is covered by new tests but cross-server link support may still vary.
> 
> **Overview**
> Adds **standalone activities as async Nexus operations** via new `ITemporalNexusClient.StartActivityAsync` overloads (expression and by-name), with activity-execution operation tokens, completion callbacks, and idempotent start plumbing aligned with workflow-backed operations.
> 
> `TemporalOperationHandler` **cancel** now routes by token type and supports activity tokens; default cancel calls `CancelAsync` on the underlying activity, with an overridable `CancelActivityExecutionAsync`. `StartActivityOptions` gains internal fields for Nexus (request ID, links, callbacks, on-conflict options), and activity/workflow starts inside a Nexus handler attach **outbound links** from server responses (workflow start also falls back to a synthetic started-event link on older servers).
> 
> Shared Nexus start logic is **factored** into `NexusOperationStartHelper`; workflow docs now say task queue defaults to the operation queue when omitted. **Activity proto links** round-trip through `ProtoLinkExtensions`, and package-validation suppressions cover the new public API.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit abff167bfca2b727a7c8f2ec05b75bf688a2b03c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/Temporalio/Nexus/NexusActivityStartHelper.cs
```diff
@@ -0,0 +1,74 @@
...
+            }
+            if (NexusOperationStartCommon.BuildCallback(
+                    nexusStartContext, callbackToken, options.Links) is { } callback)
+            {
+                options.CompletionCallbacks = new[] { callback };
+            }
+            options.RequestId = nexusStartContext.RequestId;
+
+            // Do the start call
+            var handle = await client.StartActivityAsync(
```
VegetarianOrc, line 67: In the Go implementation, we check for timeout presence to raise a handler error and surface the validation error. Is that necessary here as well?

src/Temporalio/Nexus/TemporalOperationHandler.cs
```diff
@@ -170,6 +210,36 @@ public class CancelWorkflowRunInput
...
+    /// </summary>
+    /// <remarks>WARNING: Nexus support is experimental.</remarks>
+    public class CancelActivityExecutionInput
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="CancelActivityExecutionInput"/> class.
+        /// </summary>
+        /// <param name="activityId">Activity ID extracted from the operation token.</param>
+        /// <param name="runId">Activity run ID extracted from the operation token. <c>null</c> for
+        /// tokens generated by SDKs that don't embed the run ID.</param>
```
VegetarianOrc: Should this be nullable? I think we settled on that it won't be possible for this to be null and also get a cancel.

VegetarianOrc (approved):

src/Temporalio/Nexus/TemporalOperationHandler.cs
Quinn-With-Two-Ns: Yeah good point

src/Temporalio/Nexus/NexusActivityStartHelper.cs
Quinn-With-Two-Ns: The SDK already does some of this validation, Started a discussion in slack on where we want to do the validation.

src/Temporalio/Nexus/NexusActivityExecutionToken.cs
```diff
@@ -0,0 +1,95 @@
...
+
+        /// <summary>
+        /// Build a base64url-encoded activity-execution operation token.
+        /// </summary>
+        /// <param name="namespace_">Activity namespace.</param>
+        /// <param name="activityId">Activity ID.</param>
+        /// <param name="runId">Activity run ID. May be <c>null</c> when building the token used in
+        /// the completion-callback header (which is sent before the run ID is known).</param>
+        /// <returns>Base64url-encoded operation token.</returns>
+        internal static string Build(string namespace_, string activityId, string? runId) =>
```
jmaeagle99: Rename to `Create`. The `Build` name is really only used for builders.

src/Temporalio/Nexus/NexusOperationStartCommon.cs
```diff
@@ -0,0 +1,95 @@
...
+        internal const string NexusOperationTokenHeader = "Nexus-Operation-Token";
+
+        /// <summary>
+        /// Translate inbound Nexus links into Temporal proto links. Unrecognized link types are
+        /// dropped silently; malformed links are logged and dropped.
+        /// </summary>
+        /// <param name="nexusStartContext">Nexus start context.</param>
+        /// <param name="temporalContext">Temporal operation context (used for logging).</param>
+        /// <returns>List of proto links, or <c>null</c> if there are no inbound links.</returns>
+        internal static IReadOnlyCollection<Link>? BuildInboundLinks(
```
jmaeagle99: Rename to `CreateInboundLinks`. The `Build` name is usually only used for builders.

src/Temporalio/Nexus/NexusOperationStartCommon.cs
```diff
@@ -0,0 +1,95 @@
...
+        /// <summary>
+        /// Build the completion callback from the start context, attaching the operation token
+        /// header (unless the caller already supplied one) and any inbound links.
+        /// </summary>
+        /// <param name="nexusStartContext">Nexus start context.</param>
+        /// <param name="token">Operation token to inject as a callback header.</param>
+        /// <param name="links">Links to attach to the callback (typically from
+        /// <see cref="BuildInboundLinks"/>).</param>
+        /// <returns>Callback to attach to start options, or <c>null</c> if no callback URL is set.</returns>
+        internal static Callback? BuildCallback(
```
jmaeagle99: Rename to `CreateCallback`. The `Build` name is usually only used for builders.

src/Temporalio/Nexus/NexusWorkflowRunHandle.cs
```diff
@@ -84,6 +87,44 @@ internal static byte[] Base64UrlDecode(string s)
...
         }
 
+        /// <summary>
+        /// Decode just the type field from a base64url-encoded operation token. Used by cancel
+        /// dispatch to route by token type without committing to a particular per-type schema.
+        /// </summary>
+        /// <param name="token">Base64url-encoded token string.</param>
+        /// <returns>Token type code (the JSON <c>t</c> field).</returns>
+        /// <exception cref="ArgumentException">If the token is empty or malformed.</exception>
+        internal static int LoadTokenType(string token)
```
jmaeagle99: Rename to `ParseTokenType` or `ReadTokenType`. The `Load` name usually means it involves some kind of I/O or transforming data into something executable.

src/Temporalio/Nexus/NexusActivityStartHelper.cs
```diff
@@ -0,0 +1,74 @@
...
+        /// </summary>
+        /// <param name="client">Temporal client.</param>
+        /// <param name="nexusStartContext">Nexus start context for callbacks and links.</param>
+        /// <param name="temporalContext">Temporal operation context for info and logging.</param>
+        /// <param name="activity">Activity type name.</param>
+        /// <param name="args">Activity arguments.</param>
+        /// <param name="options">Activity start options. Either ScheduleToCloseTimeout or
+        /// StartToCloseTimeout must be set; TaskQueue defaults to the operation's task queue.</param>
+        /// <returns>Base64url-encoded operation token.</returns>
+        internal static async Task<string> StartActivityAndGetTokenAsync(
```
jmaeagle99: Rename to `StartActivityAsync`. Typically do not name methods based on return type.

src/Temporalio/Nexus/NexusOperationStartCommon.cs
```diff
@@ -0,0 +1,95 @@
...
+using Temporalio.Api.Common.V1;
+
+namespace Temporalio.Nexus
+{
+    /// <summary>
+    /// Plumbing shared by <see cref="NexusWorkflowStartHelper"/> and
+    /// <see cref="NexusActivityStartHelper"/> for translating inbound Nexus links and building the
+    /// completion callback.
+    /// </summary>
+    internal static class NexusOperationStartCommon
```
jmaeagle99: Rename to `NexusOperationStartHelper` to be consistent with other utility classes.

src/Temporalio/Nexus/TemporalOperationHandler.cs
```diff
@@ -112,29 +112,56 @@ public async Task<OperationStartResult<TResult>> StartAsync(
         /// <inheritdoc/>
         public Task CancelAsync(OperationCancelContext context)
         {
-            NexusWorkflowRunHandle.OperationToken token;
+            int tokenType;
             try
             {
-                token = NexusWorkflowRunHandle.ParseToken(context.OperationToken);
+                tokenType = NexusWorkflowRunHandle.LoadTokenType(context.OperationToken);
```
jmaeagle99: I think the token parsing and related logic could use a refactor such that we use a discriminated union, something like:

```
[JsonPolymorphic(TypeDiscriminatorPropertyName = "t")]
[JsonDerivedType(typeof(WorkflowExecutionToken), 1)]
[JsonDerivedType(typeof(ActivityExecutionToken), 2)]
internal abstract record NexusOperationToken(string Namespace, int? Version);
```

Then when parsing the STJ, you'll get the exact token class out rather than having to pick out the `t` field yourself and then construct each type separately.

src/Temporalio/Nexus/TemporalOperationHandler.cs
```diff
@@ -112,29 +112,56 @@ public async Task<OperationStartResult<TResult>> StartAsync(
...
-            if (token.Namespace != NexusOperationExecutionContext.Current.Info.Namespace)
+            switch (tokenType)
             {
-                throw new HandlerException(HandlerErrorType.BadRequest, "Invalid namespace");
-            }
-            return token.Type switch
-            {
-                NexusWorkflowRunHandle.WorkflowRunOperationTokenType =>
-                    CancelWorkflowRunAsync(
+                case NexusWorkflowRunHandle.WorkflowRunOperationTokenType:
```
jmaeagle99, line 126: Looking at the naming of the this token and helper methods vs the new one for standalone activities, they feel very asymmetrical when they really share a lot in common. For example, I think that the workflow run operation token should be separated from the handle notion and standard on either the "run" or "execution" name for both token types.

jmaeagle99: I would be fine with deferring a refactor on this along with https://github.com/temporalio/sdk-dotnet/pull/748/changes/BASE..8520a4bcc34de23a824cf065f2e377f5f615a09b#r3477886756 as a separate change.

src/Temporalio/Nexus/TemporalOperationHandler.cs
jmaeagle99: I would be fine with deferring a refactor of this along with https://github.com/temporalio/sdk-dotnet/pull/748/changes/BASE..8520a4bcc34de23a824cf065f2e377f5f615a09b#r3477893944 as a separate change.

src/Temporalio/Client/TemporalClient.Workflow.cs
```diff
@@ -765,10 +738,21 @@ private async Task<WorkflowHandle<TWorkflow, TResult>> StartWorkflowInternalAsyn
...
                     }
-                    // A plain start does not produce a response link.
-                    ForwardNexusRequestLinks(req.Links);
                     var resp = await Client.Connection.WorkflowService.StartWorkflowExecutionAsync(
                         req, DefaultRetryOptions(input.Options.Rpc)).ConfigureAwait(false);
+                    if (NexusOperationExecutionContext.HasCurrent)
+                    {
+                        // Prefer the link returned by the server; fall back to a
+                        // WorkflowExecutionStarted link for older servers that don't populate it.
+                        var nexusLink = resp.Link?.ToNexusLink() ?? new Link.Types.WorkflowEvent
```
jmaeagle99, line 777: Why fallback to synthesize a link for workflows but not for activities? Is that because the server version that supports nexus + standalone activities always returns a link?

jmaeagle99 (approved):
Mostly rename requests and I think there is opportunity for unifying the token types or at least refactoring to be more symmetrical and untied from handles (which can all be deferred).

src/Temporalio/Client/TemporalClient.Workflow.cs
Quinn-With-Two-Ns: Yes, this is for old server essentially. We already have this capability to try to synthesize a link so I didn't remove this in the refactor. For activities we should always have a link

src/Temporalio/Nexus/NexusWorkflowRunHandle.cs
Quinn-With-Two-Ns: Good point, ill rename to `ParseTokenType `

jmaeagle99:
I assume that `ITemporalNexusClient` is not meant for customers to actually implement but just for us to provide easy wire up of Temporal operations as Nexus operations. If that's the case, then the API breaks are fine. Run `dotnet pack -c Debug /p:ApiCompatGenerateSuppressionFile=true` from the repo root to update the API compat suppression file.

src/Temporalio/Nexus/ProtoLinkExtensions.cs
```diff
@@ -70,6 +108,55 @@ public static NexusLink ToNexusLink(this Api.Common.V1.Link.Types.WorkflowEvent
...
         }
 
+        /// <summary>
+        /// Convert an activity link to a Nexus link.
+        /// </summary>
+        /// <param name="act">Activity link to convert.</param>
+        /// <returns>Nexus link.</returns>
+        public static NexusLink ToNexusLink(this Api.Common.V1.Link.Types.Activity act)
+        {
+            var builder = new UriBuilder
```
jmaeagle99: Might need to fix this based on what was done here: https://github.com/temporalio/sdk-dotnet/pull/743/changes#diff-f041f09e4ee02b38caef08ec16893331f18864d25213ec951945bbc6bf23ea1f

PR #750, opened 2026-06-23
Expose continue-as-new backoff start interval

Sushisource:
Expose the continue-as-new `BackoffStartInterval` option and pass it through to the Core bridge command.

Also updates the Core submodule to pick up the bridge proto field.

Validation:
- `dotnet build src/Temporalio/Temporalio.csproj`
- `dotnet test tests/Temporalio.Tests/Temporalio.Tests.csproj --filter FullyQualifiedName~ExecuteWorkflowAsync_ContinueAsNew_ProperlyContinues`


jmaeagle99 (approved):
Please update change log if you feel this is worth the callout.

PR #753, opened 2026-06-29
perf: use interpreted expression compilation on more targets

### LINKED ISSUE #751: [Bug] Expression compilation is slow for non-constant members

Mitigate a CPU performance bottleneck for extracting workflow/activity/operation methods and arguments.

### Describe the bug

Customers are advised to use type-safe calls to start workflows/activities/operations/etc (e.g. `ExecuteChildWorkflowAsync(w => w.RunAsync(input))`) rather than using the registration name and separate argument overload (e.g. `ExecuteChildWorkflowAsync("MyWorkflow", [input])`).

The former requires extracting method and argument information from call expressions. This is CPU bound expensive work because the SDK is [reflection emitting and JIT compiling](https://github.com/temporalio/sdk-dotnet/blob/main/src/Temporalio/Common/ExpressionUtil.cs#L120) every expression call rather than using the [IL interpreter](https://github.com/temporalio/sdk-dotnet/blob/main/src/Temporalio/Common/ExpressionUtil.cs#L118) for the expression calls. The JIT compilation takes locks and is invoked every time a customer uses this expression-style invocation path; there is no long term caching of the JIT compiled methods.

### Proposed Solution

Fix the compilation branch to use the interpreter-based solution in as many places as possible (namely, .NET Standard and .NET Core targets). Using simple benchmarking, the IL interpreter and invoke path is ~19x times faster than the JIT-compilation path, without caching.

jmaeagle99:
## What was changed

Include more compilation targets in the interpreted expression compilation path. This does not guarantee that the expressions will use a lightweight interpreter but does enable it if the runtime supports it. See https://learn.microsoft.com/en-us/dotnet/api/system.linq.expressions.expression-1.compile?view=net-10.0#system-linq-expressions-expression-1-compile(system-boolean)

## Why?

Better CPU performance; do not need to pay for full IL compilation (Reflection.Emit and JIT) for single-use expressions.

## Checklist

1. Closes #751 

2. How was this tested: Build and use existing tests; tested that a .NET Standard 2.0 library with interpreted expression compilation will load and execute on a .NET Framework 4.6.2 container on Windows.

3. Any docs updates needed? No

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,3 +27,8 @@ to docs, or any other relevant information.
   signaled event is attached to the caller workflow's Nexus operation history event. This makes the
   caller and callee mutually navigable in the UI for signal-based Nexus operations.
 - Exposed `BackoffStartInterval` for continue-as-new, to allow the new workflow to start after a delay.
+
+### Changed
+
+- Reduced CPU usage of type-safe calls (e.g. `ExecuteChildWorkflowAsync(wf => wf.RunAsync(arg))`)
+  by evaluating non-constant arguments via expression interpretation instead of full IL compilation, when supported by the runtime.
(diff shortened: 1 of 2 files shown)


Sushisource (approved):
Neat, good find

PR #764, opened 2026-07-07
Add NexusOperationHandler annotation

Quinn-With-Two-Ns:
## What was changed
Add `NexusOperationHandler` annotation that can be used in Nexus Service Handlers on Temporal Workers

## Why?
Makes it easier to declare Nexus Service Handlers with less boilerplate

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

Note to reviewers:
https://github.com/nexus-rpc/sdk-dotnet/pull/19 needs to land first and most of the package lock changes are only here to test locally

---

> [!NOTE]
> **Medium Risk**
> Touches Nexus operation registration, hosting DI lifetimes, and a NexusRpc dependency upgrade in experimental Nexus code; behavior changes are localized but affect how handlers are built and scoped at runtime.
> 
> **Overview**
> Adds **`[TemporalOperation]`** so a method on a `[NexusServiceHandler]` class can be the Temporal-backed Nexus **start** handler directly (signature with `TemporalOperationStartContext`, `ITemporalNexusClient`, and optional input), instead of returning an `IOperationHandler` from `[NexusOperationHandler]`. The two attributes are mutually exclusive on the same method; operations are still matched by method name to the service interface.
> 
> Registration paths now wire this through a **`TemporalOperationMethodExtension`** (`IMethodExtension` for NexusRpc) and pass it from **`TemporalWorkerOptions.AddNexusService`**, **`SimplePluginOptions.AddNexusService`**, and hosting. Hosting **`ServiceHandlerInstanceHelper`** / **`CreateNexusServiceHandlerInstance`** were refactored to an operation factory that supports both handler styles, creates a **fresh DI scope (and service instance) per start/cancel** for `[TemporalOperation]` (aligned with activities), and fails registration when `[NexusOperationHandler]` names a method with no matching operation.
> 
> **NexusRpc** is bumped to **0.4.0** (lockfiles + central package version). Changelog documents the new attribute; tests cover validation, workflow/sync/void inputs, mixed handlers, DI lifetimes, and registration errors.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 922a22e4990f75ad47d240384735dee8c2ab210d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/Temporalio/Nexus/TemporalOperationMethodExtension.cs
```diff
@@ -0,0 +1,139 @@
...
+        {
+            var ctxParam = Expression.Parameter(typeof(TemporalOperationStartContext), "ctx");
+            var clientParam = Expression.Parameter(typeof(ITemporalNexusClient), "client");
+            var inputParam = Expression.Parameter(inputType, "input");
+            var instanceExpr = Expression.Constant(instance, method.DeclaringType!);
+            var call = hasInputParam
+                ? Expression.Call(instanceExpr, method, ctxParam, clientParam, inputParam)
+                : Expression.Call(instanceExpr, method, ctxParam, clientParam);
+            var lambda = Expression.Lambda(funcType, call, ctxParam, clientParam, inputParam);
+            return lambda.Compile();
```
jmaeagle99: Is `NexusRpc` going to cache the result of calling `IMethodExtension.Extract`? If not (and maybe regardless), consider using `Expression.Compile(bool)` for interpreted execution rather than JIT compilation e.g. https://github.com/temporalio/sdk-dotnet/blob/main/src/Temporalio/Common/ExpressionUtil.cs#L119

jmaeagle99 (commented):
Small amount of feedback, but looks good overall. Will approve when dependency has been updated.

Side note: this is ripe for source generation, which would create compile-time safe operation registration, but still use the reflection and expression compilation as a fallback. I'm going to log an issue to track this, as there are other areas that can benefit as well.

src/Temporalio/Nexus/TemporalOperationMethodExtension.cs
```diff
@@ -0,0 +1,139 @@
...
+            {
+                throw new ArgumentException(
+                    $"[TemporalOperation] method {method.DeclaringType}.{method.Name} must be public");
+            }
+            if (method.IsStatic)
+            {
+                throw new ArgumentException(
+                    $"[TemporalOperation] method {method.DeclaringType}.{method.Name} must not be static");
+            }
+            var opDef = serviceDefinition.Operations.Values
```
jmaeagle99: Should this be provided by the `NexusRpc` library via the `IMethodExtension.Extract` method rather than having the `Temporalio` library figure it out? It would also eliminate the `MethodExtensionResult` return type since only the operation handler would need to be returned.

jmaeagle99:
Please update CHANGELOG.md with a blurb in the Added section.

jmaeagle99 (approved):
I think the one pitfall of this design is that if a method is attributed with `[TemporalOperation]` but it's name doesn't match a method in the service interface, then it is silently ignored. This might lead to typo-created mismatches given that the service implementation doesn't implemented the service interface. Not sure what to do with that without given the entire service definition to Temporalio to validate all attributed methods at once. Maybe could be done better with an analyzer.

PR #766, opened 2026-07-07
Document client configuration from environment

eamsden:
## What changed

- Adds root README documentation for loading client connection options from temporal.toml and environment variables.

## Notes

- Split out from the Lambda branch so the README client configuration docs can be reviewed separately.

### DIFF
--- a/README.md
+++ b/README.md
@@ -276,6 +276,26 @@ var client = await TemporalClient.ConnectAsync(new("my-namespace.a1b2c.tmprl.clo
 });
 ```
 
+#### Client Configuration From Environment
+
+Client connection options can be loaded from a `temporal.toml` file and environment variables:
+
+```csharp
+using Temporalio.Client;
+using Temporalio.Common.EnvConfig;
+
+var client = await TemporalClient.ConnectAsync(
+    ClientEnvConfig.LoadClientConnectOptions());
+```
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -276,6 +276,26 @@ var client = await TemporalClient.ConnectAsync(new("my-namespace.a1b2c.tmprl.clo
...
+
+```csharp
+using Temporalio.Client;
+using Temporalio.Common.EnvConfig;
+
+var client = await TemporalClient.ConnectAsync(
+    ClientEnvConfig.LoadClientConnectOptions());
+```
+
+By default, the loader checks `TEMPORAL_CONFIG_FILE`; if unset, it looks for `temporal.toml` in the
```
jmaeagle99, line 291: I think we should mention the behavior when the toml file doesn't exist at the expected/specified location, which I believe is that it just loads the defaults.

README.md
```diff
@@ -276,6 +276,26 @@ var client = await TemporalClient.ConnectAsync(new("my-namespace.a1b2c.tmprl.clo
...
+user config directory under `temporalio`. The selected profile is `TEMPORAL_PROFILE`, or `default`
+when unset. Environment variables such as `TEMPORAL_ADDRESS`, `TEMPORAL_NAMESPACE`,
```
jmaeagle99: Do we want to add a link to the public docs so that users can find the precise environment variables? https://docs.temporal.io/develop/environment-configuration

jmaeagle99 (approved):

PR #767, opened 2026-07-07
Align bridge env config loading behavior

eamsden:
## What was changed
Pass empty `OverrideEnvVars` through to bridge environment config loading instead of treating an empty override dictionary as unset.

## Why?
This keeps bridge config-loading behavior aligned with other SDKs and lets an explicit empty override suppress the process environment. This PR was split out from `ea/aws-lambda` so the `EnvConfig.cs` change can be reviewed independently.

## Checklist

1. Closes N/A

2. How was this tested:
Not run; branch split only.

3. Any docs updates needed?
No.

### DIFF
--- a/src/Temporalio/Bridge/EnvConfig.cs
+++ b/src/Temporalio/Bridge/EnvConfig.cs
@@ -24,7 +24,7 @@ namespace Temporalio.Bridge
 
             try
             {
-                var envVarsRef = options.OverrideEnvVars?.Count > 0
+                var envVarsRef = options.OverrideEnvVars != null
                     ? scope.ByteArray(JsonSerializer.Serialize(options.OverrideEnvVars))
                     : ByteArrayRef.Empty.Ref;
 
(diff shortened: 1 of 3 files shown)


src/Temporalio/Bridge/EnvConfig.cs
```diff
@@ -24,7 +24,7 @@ internal static class EnvConfig
 
             try
             {
-                var envVarsRef = options.OverrideEnvVars?.Count > 0
+                var envVarsRef = options.OverrideEnvVars != null
```
jmaeagle99: This is _technically_ a breaking change, even if the comments on the `OverrideEnvVars` property were correct on the intended behavior. This very minor breaking change needs to be docuemnted in the changelog and the PR title updated with a :boom:.

jmaeagle99: And since this is changing a behavior, a small test would be warranted, if possible.

jmaeagle99 (approved):
Approving, but please update the changelog at the very least.

src/Temporalio/Bridge/EnvConfig.cs
jmaeagle99: While the vast majority of the env application is done in core, the marshaling decision is owned by the .NET side. Making a test for this might prove difficult because calling `Environment.SetEnvironmentVariable` will not have observable result in core; the environment block is loaded by the .NET runtime once and mutated in memory in the .NET runtime without being reflected into the native process environment.

PR #769, opened 2026-07-08
chore: pin .NET SDK with latest patch

### LINKED ISSUE #563: Use global.json file to centralize SDK version
There are several places in build pipelines that specify the SDK version and there is no clear documentation as to which version of the SDK should be used to build the .NET projects.

## Proposed Solution

Add a [global.json](https://learn.microsoft.com/en-us/dotnet/core/tools/global-json) file to the repository that specifies the version of the .NET SDK to use. This has the benefit of:
- Removing the need for build pipelines to know which version of the .NET SDK is necessary.
- Having a central location that allow .NET tooling understands how to determine which version of the .NET SDK to use.
- Allow Dependabot to potentially update the .NET SDK version if using the full triplet version.

jmaeagle99:
## What was changed

Pin the .NET SDK using global.json file.

## Why?

Naturally define the minimum .NET SDK needed to build the product in one place.

## Checklist

1. Closes #563 
1. How was this tested: CI build

### DIFF
--- a/.github/workflows/nuget-package.yml
+++ b/.github/workflows/nuget-package.yml
@@ -6,10 +6,6 @@ on:
       - "releases/*"
   workflow_dispatch: {}
   workflow_call: {}
-  # Temporary PR trigger for testing workflow changes. Remove before merging.
-  pull_request:
-    branches:
-      - main
 permissions:
   contents: read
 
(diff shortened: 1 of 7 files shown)


.github/workflows/nuget-package.yml
```diff
@@ -6,6 +6,10 @@ on:
       - "releases/*"
   workflow_dispatch: {}
   workflow_call: {}
+  # Temporary PR trigger for testing workflow changes. Remove before merging.
+  pull_request:
+    branches:
+      - main
```
maciejdudko: Just to be safe, I'll approve after this is removed.

maciejdudko (approved):

PR #782, opened 2026-07-10
Add patch activation callback

Sushisource:
## What was changed

Added an experimental `TemporalWorkerOptions.PatchActivationCallback` that lets a worker decide whether the first newly encountered, non-replay `Workflow.Patched` call activates a patch.

The callback receives workflow information and the patch ID. Its boolean decision is memoized for the workflow run. Declined patches return `false` without recording a marker, while replay, existing history markers, and deprecated patches retain their existing behavior.

This ports the behavior introduced by https://github.com/temporalio/sdk-ruby/pull/481.

The work is intentionally split into two commits:

1. `Add patch activation callback` contains only the callback API, plumbing, activation semantics, and rollout tests.
2. `Harden read-only workflow contexts` makes the callback read-only and completes the workflow context's read-only enforcement. This includes guarding nested patches, workflow metadata and handler mutation, continue-as-new, and patch calls from queries and update validators.

> [!NOTE]
> We can drop the second commit if it feels like too much to put in the PR, but, it's done and I think adds some nice safety. 

## Why?

This allows workflow changes using `Workflow.Patched` to roll out gradually without every new-code worker immediately recording a marker that older workers cannot replay.

Keeping the read-only hardening in a subsequent commit separates the callback's intended behavior from the broader workflow-context consistency fixes discovered during review.

## Testing

- .NET SDK build across all target frameworks with zero warnings
- First callback-only commit independently formatted and tested
- 21 focused callback, read-only, mutation, handler, query, and update-validator cases
- Changed-file formatting verification
- Diff and sdk-core submodule cleanliness checks

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,10 @@ to docs, or any other relevant information.
   connections now compress outbound requests and accept gzip-compressed responses by default. If the remote
   service does not support gzip compression, the connection is downgraded to uncompressed requests. Set it
   to `GrpcCompression.None` to opt out.
+- Added AWS Lambda worker support packages, including OpenTelemetry helpers for Lambda workers.
+- Added the experimental `TemporalWorkerOptions.PatchActivationCallback`, allowing workers to
+  decide whether a first non-replay `Workflow.Patched` call should activate a patch during rolling
+  deployments.
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,18 @@ to docs, or any other relevant information.
 
 ## [Unreleased]
 
+### Added
+
+- Added the experimental `TemporalWorkerOptions.PatchActivationCallback`, allowing workers to
+  decide whether a first non-replay `Workflow.Patched` call should activate a patch during rolling
+  deployments.
+
+### Changed
+
+- Hardened read-only workflow context enforcement so queries, update validators, and patch activation
+  callbacks cannot mutate handlers or workflow details, invoke patches, or schedule workflow work.
+  Patch activation callbacks also cannot use workflow randomness or issue workflow commands.
+
 ### [1.17.0] - 2026-07-13
 
 ### Added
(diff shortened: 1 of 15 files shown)


CHANGELOG.md
```diff
@@ -32,6 +32,10 @@ to docs, or any other relevant information.
   connections now compress outbound requests and accept gzip-compressed responses by default. If the remote
   service does not support gzip compression, the connection is downgraded to uncompressed requests. Set it
   to `GrpcCompression.None` to opt out.
+- Added AWS Lambda worker support packages, including OpenTelemetry helpers for Lambda workers.
```
jmaeagle99: Looks to be accidentally duplicated. Please remove.

CHANGELOG.md
```diff
@@ -32,6 +32,10 @@ to docs, or any other relevant information.
...
+- Added the experimental `TemporalWorkerOptions.PatchActivationCallback`, allowing workers to
```
jmaeagle99: Please move to Unreleased section.

src/Temporalio/Worker/WorkflowInstance.cs
```diff
@@ -403,14 +441,26 @@ public object Instance
         /// </summary>
         internal Action<string> AssertValidLocalActivity { get; private init; }
 
+        /// <inheritdoc/>
+        public void AssertNotReadOnly(string operation)
+        {
+            if (contextFrozen || inQueryOrValidator)
```
jmaeagle99, line 447: I think the hardening is fine to add, but maybe add something in the changelog to raise that this is intentionally being checked, especially for queries and validators.

src/Temporalio/Worker/WorkflowInstance.cs
```diff
@@ -444,13 +494,33 @@ public ExternalWorkflowHandle<TWorkflow> GetExternalWorkflowHandle<TWorkflow>(
         /// <inheritdoc />
         public bool Patch(string patchId, bool deprecated)
         {
+            AssertNotReadOnly("patch");
```
jmaeagle99:
```suggestion
            AssertNotReadOnly(deprecated ? "deprecate patch" : "create patch");
```

jmaeagle99 (approved):
Few minor comments but looks good overall.

PR #783, opened 2026-07-10
Support Workflow Update as a Nexus Operation

Evanthx:
The linter failed as I added three StartUpdateWorkflowAsync overloads to ITemporalNexusClient and that gets flagged. I suppressed the errors, just calling this out here to make sure that was the right thing to do. My understanding is that when PackageValidationBaselineVersion gets bumped the suppressions will remove themselves.

### DIFF
--- a/src/Temporalio/Client/WorkflowUpdateStartOptions.cs
+++ b/src/Temporalio/Client/WorkflowUpdateStartOptions.cs
@@ -1,4 +1,4 @@
-#pragma warning disable SA1402 // We allow multiple types of the same name
+using System.Collections.Generic;
 
 namespace Temporalio.Client
 {
@@ -45,30 +45,12 @@ namespace Temporalio.Client
         /// Gets or sets the completion callbacks. Only settable by the SDK, e.g. when starting an
         /// update-workflow-backed Nexus operation.
         /// </summary>
-        internal System.Collections.Generic.IReadOnlyCollection<Api.Common.V1.Callback>? CompletionCallbacks { get; set; }
+        internal IReadOnlyCollection<Api.Common.V1.Callback>? CompletionCallbacks { get; set; }
 
         /// <summary>
         /// Gets or sets the links. Only settable by the SDK, e.g. when starting an
         /// update-workflow-backed Nexus operation.
         /// </summary>
-        internal System.Collections.Generic.IReadOnlyCollection<Api.Common.V1.Link>? Links { get; set; }
-
-        /// <summary>
-        /// Gets or sets a holder that captures the link returned on the start-update response. Only
-        /// set by the SDK, e.g. when starting an update-workflow-backed Nexus operation.
-        /// </summary>
-        internal UpdateResponseInfo? ResponseInfo { get; set; }
-    }
-
-    /// <summary>
-    /// Mutable holder used to capture the link returned on a start-update response, so callers such
-    /// as the Nexus update-workflow operation can build outbound links.
-    /// </summary>
-    internal sealed class UpdateResponseInfo
-    {
-        /// <summary>
-        /// Gets or sets the link returned on the start-update response, if any.
-        /// </summary>
-        internal Api.Common.V1.Link? Link { get; set; }
+        internal IReadOnlyCollection<Api.Common.V1.Link>? Links { get; set; }
     }
 }
--- a/src/Temporalio/Client/WorkflowUpdateStartOptions.cs
+++ b/src/Temporalio/Client/WorkflowUpdateStartOptions.cs
@@ -1,3 +1,5 @@
+#pragma warning disable SA1402 // We allow multiple types of the same name
+
 namespace Temporalio.Client
 {
     /// <summary>
@@ -32,5 +34,41 @@ namespace Temporalio.Client
         /// <c>None</c> or <c>Admitted</c> at this time.
         /// </summary>
         public WorkflowUpdateStage WaitForStage { get; set; }
+
+        /// <summary>
+        /// Gets or sets the request ID for server de-duplication. Only settable by the SDK, e.g.
+        /// when starting an update-workflow-backed Nexus operation.
+        /// </summary>
+        internal string? RequestId { get; set; }
+
+        /// <summary>
+        /// Gets or sets the completion callbacks. Only settable by the SDK, e.g. when starting an
+        /// update-workflow-backed Nexus operation.
+        /// </summary>
+        internal System.Collections.Generic.IReadOnlyCollection<Api.Common.V1.Callback>? CompletionCallbacks { get; set; }
+
+        /// <summary>
+        /// Gets or sets the links. Only settable by the SDK, e.g. when starting an
+        /// update-workflow-backed Nexus operation.
+        /// </summary>
+        internal System.Collections.Generic.IReadOnlyCollection<Api.Common.V1.Link>? Links { get; set; }
+
+        /// <summary>
+        /// Gets or sets a holder that captures the link returned on the start-update response. Only
+        /// set by the SDK, e.g. when starting an update-workflow-backed Nexus operation.
+        /// </summary>
+        internal UpdateResponseInfo? ResponseInfo { get; set; }
+    }
+
+    /// <summary>
+    /// Mutable holder used to capture the link returned on a start-update response, so callers such
+    /// as the Nexus update-workflow operation can build outbound links.
+    /// </summary>
+    internal sealed class UpdateResponseInfo
+    {
+        /// <summary>
+        /// Gets or sets the link returned on the start-update response, if any.
+        /// </summary>
+        internal Api.Common.V1.Link? Link { get; set; }
     }
 }
--- a/src/Temporalio/Nexus/ITemporalNexusClient.cs
+++ b/src/Temporalio/Nexus/ITemporalNexusClient.cs
@@ -72,5 +72,75 @@ namespace Temporalio.Nexus
         /// <returns>An async operation result containing the workflow-run token.</returns>
         Task<TemporalOperationResult<TResult>> StartWorkflowAsync<TResult>(
             string workflow, IReadOnlyCollection<object?> args, WorkflowOptions options);
+
+        /// <summary>
+        /// Start a workflow update via a lambda invoking the update method, backing this Nexus
+        /// operation with the update.
+        /// </summary>
+        /// <remarks>
+        /// <para>Only <see cref="WorkflowUpdateStage.Accepted"/> is supported for
+        /// <c>WaitForStage</c>. The operation requires a callback URL to be present.</para>
+        /// <para>Returns an async result carrying an update-workflow token, unless the update has
+        /// already completed (e.g. a retried request with the same update ID), in which case a sync
+        /// result is returned; an update that completed with an error surfaces as a failed
+        /// operation.</para>
+        /// </remarks>
+        /// <typeparam name="TWorkflow">Workflow class type.</typeparam>
+        /// <typeparam name="TResult">Update result type.</typeparam>
+        /// <param name="workflowId">Target workflow ID.</param>
+        /// <param name="updateCall">Invocation of the workflow update method with a result.</param>
+        /// <param name="options">Update start options. <c>WaitForStage</c> must be
+        /// <see cref="WorkflowUpdateStage.Accepted"/>. If the update ID is unset, the Nexus request
+        /// ID is used.</param>
+        /// <returns>An operation result for the update.</returns>
+        Task<TemporalOperationResult<TResult>> StartUpdateWorkflowAsync<TWorkflow, TResult>(
+            string workflowId,
+            Expression<Func<TWorkflow, Task<TResult>>> updateCall,
+            WorkflowUpdateStartOptions options);
+
+        /// <summary>
+        /// Start a workflow update with no result via a lambda invoking the update method, backing
+        /// this Nexus operation with the update.
+        /// </summary>
+        /// <remarks>
+        /// <para>Only <see cref="WorkflowUpdateStage.Accepted"/> is supported for
+        /// <c>WaitForStage</c>. The operation requires a callback URL to be present.</para>
+        /// </remarks>
+        /// <typeparam name="TWorkflow">Workflow class type.</typeparam>
+        /// <param name="workflowId">Target workflow ID.</param>
+        /// <param name="updateCall">Invocation of the workflow update method with no result.</param>
+        /// <param name="options">Update start options. <c>WaitForStage</c> must be
+        /// <see cref="WorkflowUpdateStage.Accepted"/>. If the update ID is unset, the Nexus request
+        /// ID is used.</param>
+        /// <returns>An operation result for the update.</returns>
+        Task<TemporalOperationResult<NoValue>> StartUpdateWorkflowAsync<TWorkflow>(
+            string workflowId,
+            Expression<Func<TWorkflow, Task>> updateCall,
+            WorkflowUpdateStartOptions options);
+
+        /// <summary>
+        /// Start a workflow update by name, backing this Nexus operation with the update.
+        /// </summary>
+        /// <remarks>
+        /// <para>Only <see cref="WorkflowUpdateStage.Accepted"/> is supported for
+        /// <c>WaitForStage</c>. The operation requires a callback URL to be present.</para>
+        /// <para>Returns an async result carrying an update-workflow token, unless the update has
+        /// already completed (e.g. a retried request with the same update ID), in which case a sync
+        /// result is returned; an update that completed with an error surfaces as a failed
+        /// operation.</para>
+        /// </remarks>
+        /// <typeparam name="TResult">Update result type.</typeparam>
+        /// <param name="workflowId">Target workflow ID.</param>
+        /// <param name="update">Update name.</param>
+        /// <param name="args">Update arguments.</param>
+        /// <param name="options">Update start options. <c>WaitForStage</c> must be
+        /// <see cref="WorkflowUpdateStage.Accepted"/>. If the update ID is unset, the Nexus request
+        /// ID is used.</param>
+        /// <returns>An operation result for the update.</returns>
+        Task<TemporalOperationResult<TResult>> StartUpdateWorkflowAsync<TResult>(
+            string workflowId,
+            string update,
+            IReadOnlyCollection<object?> args,
+            WorkflowUpdateStartOptions options);
     }
 }
--- a/src/Temporalio/Nexus/ITemporalNexusClient.cs
+++ b/src/Temporalio/Nexus/ITemporalNexusClient.cs
@@ -79,7 +79,7 @@ namespace Temporalio.Nexus
         /// </summary>
         /// <remarks>
         /// <para>Only <see cref="WorkflowUpdateStage.Accepted"/> is supported for
-        /// <c>WaitForStage</c>. The operation requires a callback URL to be present.</para>
+        /// <c>WaitForStage</c>. </para>
         /// <para>Returns an async result carrying an update-workflow token, unless the update has
         /// already completed (e.g. a retried request with the same update ID), in which case a sync
         /// result is returned; an update that completed with an error surfaces as a failed
--- a/src/Temporalio/Nexus/ITemporalNexusClient.cs
+++ b/src/Temporalio/Nexus/ITemporalNexusClient.cs
@@ -92,11 +92,13 @@ namespace Temporalio.Nexus
         /// <param name="options">Update start options. <c>WaitForStage</c> must be
         /// <see cref="WorkflowUpdateStage.Accepted"/>. If the update ID is unset, the Nexus request
         /// ID is used.</param>
+        /// <param name="runId">Target workflow run ID, or null for the latest run.</param>
         /// <returns>An operation result for the update.</returns>
-        Task<TemporalOperationResult<TResult>> StartUpdateWorkflowAsync<TWorkflow, TResult>(
+        Task<TemporalOperationResult<TResult>> StartWorkflowUpdateAsync<TWorkflow, TResult>(
             string workflowId,
             Expression<Func<TWorkflow, Task<TResult>>> updateCall,
-            WorkflowUpdateStartOptions options);
+            WorkflowUpdateStartOptions options,
+            string? runId = null);
 
         /// <summary>
         /// Start a workflow update with no result via a lambda invoking the update method, backing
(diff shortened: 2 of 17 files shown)


tests/Temporalio.Tests/WorkflowEnvironment.cs
```diff
@@ -59,16 +69,13 @@ public async Task InitializeAsync()
...
+                Environment.GetEnvironmentVariable("TEMPORAL_TEST_NEXUS_UPDATE_CALLBACKS") == "true";
         }
         else
         {
             // Otherwise, local server is good
-            env = await Temporalio.Testing.WorkflowEnvironment.StartLocalAsync(new()
-            {
-                DevServerOptions = new()
-                {
-                    DownloadVersion = "v1.7.2-standalone-nexus-operations",
```
Quinn-With-Two-Ns, line 70: Why did you change this? I think it will break SANO tests

tests/Temporalio.Tests/WorkflowEnvironment.cs
```diff
@@ -28,6 +28,16 @@ public WorkflowEnvironment()
 
     public string KitchenSinkWorkerTaskQueue => kitchenSinkWorker.Value.TaskQueue;
 
+    /// <summary>
+    /// Gets a value indicating whether this environment supports UpdateWorkflow-backed Nexus
+    /// operations, which require the server to deliver update-completion callbacks
+    /// (<c>history.enableUpdateCallbacks</c>) together with a Nexus callback-endpoint template.
+    /// The pinned local dev server does not yet support these; once it does, add those flags to the
```
Quinn-With-Two-Ns: The local dev server should support this, if it doesn't we can't merge this anyway so this logic seems unnecessary since the local dev server needs to support it

src/Temporalio/Nexus/TemporalOperationHandler.cs
```diff
@@ -170,6 +203,42 @@ public class CancelWorkflowRunInput
...
+            UpdateId = updateId;
+        }
+
+        /// <summary>
+        /// Gets the workflow ID extracted from the operation token.
+        /// </summary>
+        public string WorkflowId { get; }
+
+        /// <summary>
+        /// Gets the workflow run ID extracted from the operation token. May be empty.
```
Quinn-With-Two-Ns: This should never be empty here. We should read the value the server gives back on the update response.

src/Temporalio/Nexus/ITemporalNexusClient.cs
```diff
@@ -72,5 +72,75 @@ Task<TemporalOperationResult<NoValue>> StartWorkflowAsync<TWorkflow>(
...
         Task<TemporalOperationResult<TResult>> StartWorkflowAsync<TResult>(
             string workflow, IReadOnlyCollection<object?> args, WorkflowOptions options);
+
+        /// <summary>
+        /// Start a workflow update via a lambda invoking the update method, backing this Nexus
+        /// operation with the update.
+        /// </summary>
+        /// <remarks>
+        /// <para>Only <see cref="WorkflowUpdateStage.Accepted"/> is supported for
+        /// <c>WaitForStage</c>. The operation requires a callback URL to be present.</para>
```
Quinn-With-Two-Ns:
```suggestion
        /// <c>WaitForStage</c>. </para>
```
The callback should not be exposed to users directly right?

src/Temporalio/Nexus/TemporalOperationHandler.cs
Evanthx: Comment removed. I was reading the spec and that said the [RunID](https://app.notion.com/p/temporalio/Nexus-Operation-Token-Format-3738fc56773880d483dcd02931d6f90a) might be empty in a Workflow Update Token - so that is why I had accounted for that use case.

tests/Temporalio.Tests/WorkflowEnvironment.cs
Evanthx: It was still there but moved a few lines lower. But I tested without these changes and things still do work so I don't think these were needed after all and removed them

tests/Temporalio.Tests/Worker/NexusUpdateOperationTests.cs
```diff
@@ -0,0 +1,312 @@
...
+                taskQueue, endpoint, new(pending.Id, Amount: 5));
+
+            // The Nexus operation should remain pending (started, not failed) for a short window.
+            await Task.Delay(TimeSpan.FromSeconds(3));
+            var desc = await caller.DescribeAsync();
+            Assert.NotEqual(WorkflowExecutionStatus.Failed, desc.Status);
+        });
+    }
+
+    private void SkipIfUpdateCallbacksUnsupported()
```
Quinn-With-Two-Ns: Looking at the other dotnet SDK, we don't skip tests like this and I think it is a dangerous to have logic like this in the tests since we might accidentally skip these tests and loose there coverage. I'll defer to @jmaeagle99 but my preference is to not have this logic

tests/Temporalio.Tests/Worker/NexusUpdateOperationTests.cs
```diff
@@ -0,0 +1,312 @@
...
+/// <para>
+/// Every test here is gated on the <c>TEMPORAL_TEST_NEXUS_UPDATE_CALLBACKS</c> environment variable
+/// and is skipped unless the environment can deliver update-completion callbacks
+/// (<c>history.enableUpdateCallbacks</c> plus a Nexus callback-endpoint template). The pinned local
+/// dev server does not support these yet, so these tests are ALL currently blocked purely on that
+/// server-capability gap. Set <c>TEMPORAL_TEST_NEXUS_UPDATE_CALLBACKS=true</c> when pointing at a
+/// capable server to run them — no code change here required.
+/// </para>
+/// </summary>
+public class NexusUpdateOperationTests : WorkflowEnvironmentTestBase
```
Quinn-With-Two-Ns, line 18: I think we are missing some tests for if the update ID is already completed? Two things I think we should test are sending two Nexus Operations, one after the other has completed, that use the same update ID.

Quinn-With-Two-Ns: If we already have these tests then please feel free to ignore

tests/Temporalio.Tests/WorkflowEnvironment.cs
Evanthx: Removed

Quinn-With-Two-Ns:
The dev server is missing the DC for this feature https://github.com/temporalio/sdk-go/pull/2417/files#diff-265865b1cb4e06ecf08b23b205f709fb4281ba4d4804cbe46f8b91a725073fd5R171 and then we can remove all the test skipping

src/Temporalio/Nexus/ITemporalNexusClient.cs
```diff
@@ -72,5 +72,75 @@ Task<TemporalOperationResult<NoValue>> StartWorkflowAsync<TWorkflow>(
...
+        /// </remarks>
+        /// <typeparam name="TWorkflow">Workflow class type.</typeparam>
+        /// <typeparam name="TResult">Update result type.</typeparam>
+        /// <param name="workflowId">Target workflow ID.</param>
+        /// <param name="updateCall">Invocation of the workflow update method with a result.</param>
+        /// <param name="options">Update start options. <c>WaitForStage</c> must be
+        /// <see cref="WorkflowUpdateStage.Accepted"/>. If the update ID is unset, the Nexus request
+        /// ID is used.</param>
+        /// <returns>An operation result for the update.</returns>
+        Task<TemporalOperationResult<TResult>> StartUpdateWorkflowAsync<TWorkflow, TResult>(
```
jmaeagle99: "StartUpdateWorkflowAsync" sounds like I'm starting a workflow that performs updates. I think it should be "StartWorkflowUpdateAsync" which better matches all of the options and comments.

jmaeagle99: None of these new APIs allow specifying the run ID like the non-nexus temporal client allows. Intentional? I think the OperationToken's RunId is always null because of this.

src/Temporalio/Nexus/NexusWorkflowStartHelper.cs
```diff
@@ -118,5 +120,151 @@ internal static async Task<NexusWorkflowRunHandle> StartWorkflowAsync(
...
+            // state (ID, links, callbacks, request ID) back to a caller that reuses the instance.
+            options = (WorkflowUpdateStartOptions)options.Clone();
+
+            // If an update ID is not provided, use the Nexus request ID. This protects against
+            // retried requests (e.g. due to a network failure) spawning duplicate updates. If the
+            // request ID is also empty, generate a fallback so the update ID (and the operation
+            // token derived from it) is never empty.
+            if (string.IsNullOrEmpty(options.Id))
+            {
+                options.Id = string.IsNullOrEmpty(nexusStartContext.RequestId)
```
jmaeagle99: What if the request ID is whitespace? https://learn.microsoft.com/en-us/dotnet/api/system.string.isnullorwhitespace?view=net-10.0

src/Temporalio/Nexus/NexusWorkflowStartHelper.cs
```diff
@@ -118,5 +120,151 @@ internal static async Task<NexusWorkflowRunHandle> StartWorkflowAsync(
...
+
+            if (nexusStartContext.CallbackUrl is not { } callbackUrl)
+            {
+                throw new HandlerException(
+                    HandlerErrorType.BadRequest,
+                    "callback URL required for async UpdateWorkflow operation invocations");
+            }
+
+            var client = temporalContext.TemporalClient;
+            var namespace_ = client.Options.Namespace;
```
jmaeagle99: Use `@namespace` instead. The `@` is used to tell the compiler that it is an identifier rather than a keyword.

src/Temporalio/Nexus/NexusWorkflowStartHelper.cs
```diff
@@ -118,5 +120,151 @@ internal static async Task<NexusWorkflowRunHandle> StartWorkflowAsync(
...
+
+            // Generate the operation token before starting the update; it is needed for the callback
+            // header.
+            var handle = new NexusWorkflowUpdateHandle(
+                namespace_, workflowId, runId: string.Empty, updateId: options.Id!, version: 0);
```
jmaeagle99: I would capture the ID as a local non-nullable string to avoid refactors causing the value to be null and being missed due to the assertion.

src/Temporalio/Nexus/NexusWorkflowUpdateHandle.cs
```diff
@@ -0,0 +1,127 @@
...
+using System.Text.Json;
+
+namespace Temporalio.Nexus
+{
+    /// <summary>
+    /// Update handle representing an update-workflow operation started from a Nexus operation. It
+    /// carries the identifiers needed to encode/decode an update-workflow operation token.
+    /// </summary>
+    /// <remarks>WARNING: Nexus support is experimental.</remarks>
+    public class NexusWorkflowUpdateHandle
```
jmaeagle99: This isn't used in any public API; is that an oversight? If not, make it `internal` and change `internal` to `public` on each of its members.

src/Temporalio/Nexus/TemporalOperationHandler.cs
```diff
@@ -125,6 +125,25 @@ public Task CancelAsync(OperationCancelContext context)
             {
                 throw new HandlerException(HandlerErrorType.BadRequest, "Invalid namespace");
             }
+            if (token.Type == NexusWorkflowRunHandle.UpdateWorkflowOperationTokenType)
```
jmaeagle99: This should be part of the switch below.

src/Temporalio/Nexus/ProtoLinkExtensions.cs
```diff
@@ -132,6 +138,25 @@ public static Api.Common.V1.Link.Types.NexusOperation ToNexusOperation(this Nexu
...
+        /// <summary>
+        /// Convert a workflow link to a Nexus link. Unlike a workflow-event link, this points at a
+        /// workflow execution without referencing a particular history event, which is used when
+        /// there is no event to link to (e.g. a rejected update).
+        /// </summary>
+        /// <param name="workflow">Workflow link to convert.</param>
+        /// <returns>Nexus link.</returns>
+        public static NexusLink ToNexusLink(this Api.Common.V1.Link.Types.Workflow workflow)
+        {
+            var builder = new UriBuilder
```
jmaeagle99: I think this had some issue with erasing the empty authority segment. Please write just like the other links.

src/Temporalio/Nexus/TemporalNexusClient.cs
```diff
@@ -75,5 +77,70 @@ public async Task<TemporalOperationResult<TResult>> StartWorkflowAsync<TResult>(
...
+        /// <inheritdoc/>
+        public async Task<TemporalOperationResult<TResult>> StartUpdateWorkflowAsync<TResult>(
+            string workflowId,
+            string update,
+            IReadOnlyCollection<object?> args,
+            WorkflowUpdateStartOptions options)
+        {
+            // Reserve the single async-operation slot for this operation invocation. Only a genuine
+            // async result consumes it; a sync result or a failure releases it.
+            if (Interlocked.CompareExchange(ref asyncStarted, 1, 0) != 0)
```
jmaeagle99: Should this "only one async operation" guard be added to `StartWorkflowAsync` as well?

src/Temporalio/Client/WorkflowUpdateStartOptions.cs
```diff
@@ -32,5 +34,41 @@ public WorkflowUpdateStartOptions(string id, WorkflowUpdateStage waitForStage)
...
+        /// Gets or sets the request ID for server de-duplication. Only settable by the SDK, e.g.
+        /// when starting an update-workflow-backed Nexus operation.
+        /// </summary>
+        internal string? RequestId { get; set; }
+
+        /// <summary>
+        /// Gets or sets the completion callbacks. Only settable by the SDK, e.g. when starting an
+        /// update-workflow-backed Nexus operation.
+        /// </summary>
+        internal System.Collections.Generic.IReadOnlyCollection<Api.Common.V1.Callback>? CompletionCallbacks { get; set; }
```
jmaeagle99: nit: use a file-level using for `System.Collections.Generic` to make this less and the next property verbose.

src/Temporalio/Nexus/TemporalNexusClient.cs
jmaeagle99: Is Interlocked necessary? Do we really expect that callers to this API are multithreaded?

src/Temporalio/Nexus/NexusWorkflowStartHelper.cs
jmaeagle99: I think this is only used once; I would just inline it at the consumption site.

src/Temporalio/Client/WorkflowUpdateStartOptions.cs
```diff
@@ -32,5 +34,41 @@ public WorkflowUpdateStartOptions(string id, WorkflowUpdateStage waitForStage)
...
+        /// Gets or sets the links. Only settable by the SDK, e.g. when starting an
+        /// update-workflow-backed Nexus operation.
+        /// </summary>
+        internal System.Collections.Generic.IReadOnlyCollection<Api.Common.V1.Link>? Links { get; set; }
+
+        /// <summary>
+        /// Gets or sets a holder that captures the link returned on the start-update response. Only
+        /// set by the SDK, e.g. when starting an update-workflow-backed Nexus operation.
+        /// </summary>
+        internal UpdateResponseInfo? ResponseInfo { get; set; }
```
jmaeagle99: I think this is the only property that is not really an input via options but is an output. Would it make more sense to put this on WorkflowUpdateHandle instead, like the KnownOutcome internal property?

tests/Temporalio.Tests/Worker/NexusUpdateOperationTests.cs
jmaeagle99: Please remove this guard, add the dynamic config to the test environment, and update the server version if necessary. Seems like this dynamic config already exists in some dev server verison as demonstrated in sdk-go: https://github.com/temporalio/sdk-go/pull/2417/changes#diff-265865b1cb4e06ecf08b23b205f709fb4281ba4d4804cbe46f8b91a725073fd5R171

src/Temporalio/Client/WorkflowUpdateStartOptions.cs
```diff
@@ -32,5 +34,41 @@ public WorkflowUpdateStartOptions(string id, WorkflowUpdateStage waitForStage)
...
+    }
+
+    /// <summary>
+    /// Mutable holder used to capture the link returned on a start-update response, so callers such
+    /// as the Nexus update-workflow operation can build outbound links.
+    /// </summary>
+    internal sealed class UpdateResponseInfo
```
jmaeagle99: If this type remains after addressing feedback, I would move it to is own file rather than suppressing the diagnostic that says to do the same.

tests/Temporalio.Tests/Worker/NexusUpdateOperationTests.cs
Evanthx: Just adding a comment as I go - this mostly works, but exposed an issue that I fixed in NexusWorkflowStartHelper - the cancellation token needed to get forwarded. So that is why there is now a change there!

src/Temporalio/Nexus/NexusWorkflowStartHelper.cs
Evanthx: Changed, thank you for the language tip!

src/Temporalio/Nexus/NexusWorkflowStartHelper.cs
Evanthx: Changed

src/Temporalio/Nexus/NexusWorkflowStartHelper.cs
Evanthx: Done

src/Temporalio/Client/WorkflowUpdateStartOptions.cs
Evanthx: Done

src/Temporalio/Nexus/NexusWorkflowStartHelper.cs
Evanthx: Done

src/Temporalio/Nexus/TemporalNexusClient.cs
Evanthx: Fair - fixed

src/Temporalio/Client/WorkflowUpdateStartOptions.cs
Evanthx: Moved

src/Temporalio/Client/WorkflowUpdateStartOptions.cs
Evanthx: This did NOT remain after addressing feedback!

tests/Temporalio.Tests/Worker/NexusUpdateOperationTests.cs
Evanthx: Added!

src/Temporalio/Nexus/ITemporalNexusClient.cs
Evanthx: Changed!

Evanthx: Good catch, fixed!

src/Temporalio/Nexus/NexusWorkflowStartHelper.cs
jmaeagle99: Maybe something more like the following so we don't have to use the null-forgiving operator.

```csharp
string updateId = !string.IsNullOrEmpty(options.Id)
    ? options.Id
    : string.IsNullOrWhiteSpace(nexusStartContext.RequestId)
        ? Guid.NewGuid().ToString()
        : nexusStartContext.RequestId;
options.Id = updateId;
```

And call out that the write-back into `options.Id` is important so that the caller gets the resolved update ID.

src/Temporalio/Nexus/NexusWorkflowStartHelper.cs
```diff
@@ -118,5 +121,173 @@ internal static async Task<NexusWorkflowRunHandle> StartWorkflowAsync(
...
+            options = (WorkflowUpdateStartOptions)options.Clone();
+
+            // If an update ID is not provided, use the Nexus request ID. This protects against
+            // retried requests (e.g. due to a network failure) spawning duplicate updates. If the
+            // request ID is also empty, generate a fallback so the update ID (and the operation
+            // token derived from it) is never empty.
+            if (string.IsNullOrEmpty(options.Id))
+            {
+                options.Id = string.IsNullOrWhiteSpace(nexusStartContext.RequestId)
+                    ? Guid.NewGuid().ToString()
```
jmaeagle99: Not certain, but is the fallback to `Guid.NewGuid()` correct? Thinking about if user didn't supply an ID and if this is the initial run of this call, it calls `Guid.NewGuid()` to generate the ID. But what if the Nexus task gets redelivered for some reason? Would that not generate a different update ID because it would call `Guid.NewGuid()` again?

Quinn-With-Two-Ns: If a user doesn't supply an ID, then well use the runID, which really should always be set by the server

src/Temporalio/Nexus/NexusWorkflowStartHelper.cs
Evanthx: Did my best but evidently "IsNullOrWhiteSpace" is not marked by the compiler as guaranteeing that the resulting value isn't null, so I had to do a little weirdness. (Claude says that it is actually netstandard2.0/net462  that doesn't mark the guarantee, but either way ... )

src/Temporalio/Nexus/NexusWorkflowStartHelper.cs
Evanthx: Fixed

jmaeagle99 (commented):
There are still a few outstanding comments:
- https://github.com/temporalio/sdk-dotnet/pull/783#discussion_r3588856344
- https://github.com/temporalio/sdk-dotnet/pull/783#discussion_r3588867236
- https://github.com/temporalio/sdk-dotnet/pull/783#discussion_r3588893953
- https://github.com/temporalio/sdk-dotnet/pull/783#discussion_r3588910532

With those addressed, I think this PR can be approved.

src/Temporalio/Nexus/NexusWorkflowUpdateHandle.cs
Evanthx: Good catch, changed to internal

src/Temporalio/Nexus/TemporalOperationHandler.cs
Evanthx: Done

src/Temporalio/Nexus/ProtoLinkExtensions.cs
Evanthx: Done

src/Temporalio/Nexus/TemporalNexusClient.cs
Evanthx: Done

jmaeagle99 (approved):
I think you need to update the CompatibilitySuppressions.xml again due to the rename of the public API you've introduced in this same PR, but otherwise looks good to me!

PR #787, opened 2026-07-13
:boom: feature: payloads/memo size validation

### LINKED ISSUE #638: SDK should fail workflow task if payloads size it known to be too large - .NET
Parent issue: https://github.com/temporalio/features/issues/723

jmaeagle99:
## What was changed

- Added `PayloadLimitsOptions` (`PayloadsWarnSize`, `MemoWarnSize`) and wired it onto
  `TemporalConnectionOptions.PayloadLimits`; over-threshold values are logged (`[TMPRL1103]`) but
  still sent, `0` disables.
- Added `TemporalWorkerOptions.DisablePayloadErrorLimit` to opt out of worker error-limit
  enforcement and defer to the server.
- Update `sdk-core` submodule to bring in payloads/memo size enforcement.

## :boom: Breaking Change

Payload size-limit enforcement is now on by default; existing code changes behavior without opting in: an oversized payload is now failed proactively by the worker (as a retryable task failure) before it's sent, rather than being sent and hard-failed by the server as before.

If you use a proxy between the worker and server that alters the size of payloads (e.g.
compression, encryption, external storage), it is advised that you disable size enforcement by
setting `DisablePayloadErrorLimit` to `true` on the worker.

## Why?

Prevent server from hard-failing workflows, activities, and operations due to payloads/memo size limit violations. Allows customers to potentially update their workflows, activities, and operations to reduce the oversized payloads/memos.

## Checklist

1. Closes #638 
1. How was this tested: Core library tests and minimal .NET tests

## Todo

1. [x] Wait for https://github.com/temporalio/sdk-rust/pull/1363 to be merged
1. [x] Updated sdk-core submodule after https://github.com/temporalio/sdk-rust/pull/1363 is merged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,8 +21,8 @@ to docs, or any other relevant information.
 
 ### Breaking Changes
 
-- Payload/memo size-limit enforcement (experimental) is now on by default. Workers proactively validate outbound
-  payload/memo sizes before sending: a field over the warn threshold is logged (`[TMPRL1103]` at
+- By default, workers now proactively validate outbound payload/memo sizes before sending: a field
+  over the warn threshold is logged (`[TMPRL1103]` at
   `WARN`) but still sent, while a task completion over the error limit is failed retryably
   (`[TMPRL1103]` at `ERROR`) instead of sent. Previously these reached the server, which terminated
   the workflow or failed the activity non-retryably; failing retryably instead lets a corrected
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,18 @@ to docs, or any other relevant information.
 
 ## [Unreleased]
 
+### Breaking Changes
+
+- Payload/memo size-limit enforcement (experimental) is now on by default. Workers proactively validate outbound
+  payload/memo sizes before sending: a field over the warn threshold is logged (`[TMPRL1103]` at
+  `WARN`) but still sent, while a task completion over the error limit is failed retryably
+  (`[TMPRL1103]` at `ERROR`) instead of sent. Previously these reached the server, which terminated
+  the workflow or failed the activity non-retryably; failing retryably instead lets a corrected
+  workflow or activity be redeployed and recover. Tune warn thresholds via
+  `TemporalConnectionOptions.PayloadLimits`. If you use a proxy between the worker and server that
+  alters the size of payloads (e.g. compression, encryption, external storage), it is advised that
+  you disable size enforcement by setting `DisablePayloadErrorLimit` to `true` on the worker.
+
 ### [1.17.0] - 2026-07-13
 
 ### Added
--- a/src/Temporalio/Bridge/OptionsExtensions.cs
+++ b/src/Temporalio/Bridge/OptionsExtensions.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
 using Microsoft.Extensions.Logging;
+using Temporalio.Client;
 using Temporalio.Exceptions;
 using Temporalio.Worker.Tuning;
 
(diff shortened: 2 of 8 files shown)


src/Temporalio/Client/PayloadLimitsOptions.cs
```diff
@@ -0,0 +1,33 @@
...
+    /// WARNING: This API is experimental and may change in the future.
+    /// </remarks>
+    public class PayloadLimitsOptions : ICloneable
+    {
+        /// <summary>
+        /// Gets or sets the warning threshold, in bytes, for the size of an outbound payload-bearing
+        /// field. Over-threshold fields are logged but still sent to the server. Defaults to 512KiB;
+        /// set to <c>0</c> to disable.
+        /// </summary>
+        public int PayloadsWarnSize { get; set; } = 512 * 1024;
```
maciejdudko: `PayloadsWarnSize` and `MemoWarnSize` are unsigned 64-bit in Core. Any specific reason they're signed 32-bit here?

CHANGELOG.md
```diff
@@ -19,6 +19,18 @@ to docs, or any other relevant information.
 
 ## [Unreleased]
 
+### Breaking Changes
+
+- Payload/memo size-limit enforcement (experimental) is now on by default. Workers proactively validate outbound
```
maciejdudko: We shouldn't have experimental features enabled by default. Either there should be grace period of 1 release where the default is old behavior, or it should be instant-stable.

maciejdudko (changes requested):

src/Temporalio/Client/PayloadLimitsOptions.cs
jmaeagle99: - Unsigned size/length properties are not typical for non-interop public APIs in .NET.
- Customers are extremely unlikely to want to set their thresholds for this to be higher than 2 GB.

CHANGELOG.md
jmaeagle99: No longer describing the full feature as experimental.

src/Temporalio/Bridge/OptionsExtensions.cs
```diff
@@ -292,6 +295,10 @@ public static unsafe Interop.TemporalCoreConnectionOptions ToInteropOptions(
                     _ => throw new ArgumentException(
                         $"Unsupported gRPC compression: {options.GrpcCompression.GetType()}"),
                 },
+                payloads_warn_size =
+                    (ulong)Math.Max(0, payloadLimits.PayloadsWarnSize),
+                memo_warn_size =
+                    (ulong)Math.Max(0, payloadLimits.MemoWarnSize),
```
eamsden: This silently converts any negative value to 0, which [disables warnings](https://github.com/temporalio/sdk-rust/blob/3dac9013b9031e5ffd51d7335838585b2db42efb/crates/client/src/options_structs.rs#L254-L273). Negative sizes should throw `ArgumentOutOfRangeException` prior to casting.

tests/Temporalio.Tests/Worker/PayloadSizeLimitsTests.cs
```diff
@@ -0,0 +1,242 @@
...
+        {
+            // With the opt-out, core does not pre-fail; the oversized activity input reaches the
+            // server.
+            var handle = await env.Client.StartWorkflowAsync(
+                (LargePayloadWorkflow wf) => wf.RunAsync(new(PayloadErrorLimit + 1024, 0)),
+                new($"wf-{Guid.NewGuid()}", workerOpts.TaskQueue!)
+                {
+                    ExecutionTimeout = TimeSpan.FromSeconds(5),
+                });
+            await Assert.ThrowsAsync<WorkflowFailedException>(() => handle.GetResultAsync());
```
eamsden: Ideally this would prove that the rejection came from the server, analogous to the [assertion in Python](https://github.com/temporalio/sdk-python/blob/b9168c1d8c68a84ab709633f39c26dbaa4997f05/tests/worker/test_payload_size_limits.py#L144-L176).

eamsden (approved):
Please fix the payload limits normalization prior to merging.

maciejdudko (approved):

PR #794, opened 2026-07-20
feat(testing): support envconfig in integration test harness

THardy98:
## What was changed
Add opt-in envconfig support to .NET integration-test harness.

## Why?
This enables integration tests to target arbitrary Temporal server environments, including local, staging, release, and Temporal Cloud environments.

## Checklist

1. Part of https://github.com/temporalio/features/issues/851

2. How was this tested:
- An integration test runs a Workflow against an envconfig-selected server using the standard shared test harness and a non-default namespace.

3. Any docs updates needed?
No


tests/Temporalio.Tests/WorkflowEnvironment.cs
```diff
@@ -30,8 +38,16 @@ public WorkflowEnvironment()
 
     public async Task InitializeAsync()
     {
-        // If an existing target is given via environment variable, use that
-        if (Environment.GetEnvironmentVariable("TEMPORAL_TEST_CLIENT_TARGET_HOST") is string host)
+        if (string.Equals(
```
jmaeagle99: It looks a little strange to have to (1) set envConfigOptions in the ctor, (2) set an environment variable, and then (3) call InitializeAsync, which then decides whether the envConfigOptions is used based on the environment variable.

How about we have the default constructor pass an empty ProfileLoadOptions if the environment variable is set and simplify the check in InitializeAsync?

```
public WorkflowEnvironment()
    : this(UseEnvConfigFromEnvironment() ? new ClientEnvConfig.ProfileLoadOptions() : null)
{
}

internal WorkflowEnvironment(ClientEnvConfig.ProfileLoadOptions? envConfigOptions)
{
    this.envConfigOptions = envConfigOptions;
    this.kitchenSinkWorker = new(StartKitchenSinkWorker);
}

public async Task InitializeAsync()
{
    if (null != envConfigOptions)
    {
        var options = ClientEnvConfig.LoadClientConnectOptions(envConfigOptions);
        env = new(await TemporalClient.ConnectAsync(options));
    }
    else if (Environment.GetEnvironmentVariable("TEMPORAL_TEST_CLIENT_TARGET_HOST") is string host)
    ...
}

private static bool UseEnvConfigFromEnvironment() => string.Equals(
    Environment.GetEnvironmentVariable("TEMPORAL_TEST_ENV_CONFIG_SERVER"),
    "true", StringComparison.OrdinalIgnoreCase);
```

src/Temporalio/Testing/WorkflowEnvironment.cs
```diff
@@ -55,6 +56,21 @@ public WorkflowEnvironment(ITemporalClient client)
...
 
+        /// <summary>
+        /// Create a workflow environment from client environment configuration.
+        /// </summary>
+        /// <param name="profileLoadOptions">Options for loading the client configuration profile.</param>
+        /// <returns>The created environment.</returns>
+        /// <remarks>
+        /// This connects to an existing Temporal environment and does not manage server lifecycle.
+        /// </remarks>
+        public static async Task<WorkflowEnvironment> CreateFromEnvConfigAsync(
```
jmaeagle99, line 67: Does this need to be in the public API? It seems all of these are easily accessible by public consumers that they could do it themselves, if necessary.

THardy98: I don't mind adding this as a convenience

tests/Temporalio.Tests/WorkflowEnvironment.cs
THardy98: removed in favour of just the helper in the public API

src/Temporalio/Testing/WorkflowEnvironment.cs
jmaeagle99: Sorry. I think I read this as being in the public API of the shipped library, but I was mistaken. Either way, it's fine. Thanks for documenting it.

jmaeagle99 (approved):

PR #795, opened 2026-07-21
Add transfer type payload conversion hooks

tconley1428:
Adds experimental SDK payload-converter support for values and target types marked with `TemporalTransferTypeAttribute`.

The attribute points at an `ITemporalTransferTypeConverter` implementation. The converter supplies the `TransferType` handed to the configured payload converter, converts user values into that representation for encoding, and reconstructs user values from it for decoding. General transfer type hooks intentionally do not receive a payload converter.

This keeps normal SDK payload-converter behavior in one place, including codec handling and serialization-context propagation.

This change:

- wraps payload converters when they are stored on `DataConverter`, including `with { PayloadConverter = ... }`
- adds experimental public `TemporalTransferTypeAttribute` and `ITemporalTransferTypeConverter` APIs
- keeps the actual payload-converter adapter internal
- caches validated transfer type converter instances per marked type
- applies hooks before delegating to the configured payload converter
- adds converter tests proving the transformed transfer type is encoded with the active workflow serialization context


src/Temporalio/Converters/TemporalTransferTypePayloadConverter.cs
```diff
@@ -0,0 +1,102 @@
...
+                    $"{attr.ConverterType} that does not implement {nameof(ITemporalTransferTypeConverter)}.");
+            }
+            if (attr.ConverterType.ContainsGenericParameters)
+            {
+                throw new InvalidOperationException(
+                    $"Type {type} has an open generic Temporal transfer type converter type " +
+                    $"{attr.ConverterType}.");
+            }
+
+            if (Activator.CreateInstance(attr.ConverterType) is not ITemporalTransferTypeConverter converter)
```
jmaeagle99, line 112: While you check a few cases above, there are other ones, such as:
- Type could be abstract e.g. `attr.ConverterType.IsAbstract`
- Type constructor could have parameters e.g. `!attr.ConverterType.IsValueType && attr.ConverterType.GetConstructor(Type.EmptyTypes) == null`; all value types (e.g. `struct` have a default parameterless constructor, so you can ignore those.
- Type constructor must be public; `GetConstructor` will return `null` if not public instance constructor.

jmaeagle99, line 112: `Activator.CreateInstance` can throw exceptions, especially for all of the cases you are checking. But it can also throw exceptions due to execution problems, such as the target type constructor raises an exception, the static constructor for the type raises an exception, a dependent library cannot be loaded, etc. All different types of exceptions you cannot anticipate the types of (or you can, there's just a lot of them). Not sure if you want to let those bubble out as-is, which would probably be fine to do. Just bringing awareness.

src/Temporalio/Converters/TemporalTransferTypeAttribute.cs
```diff
@@ -0,0 +1,29 @@
...
+    /// Marks a type as converting to and from a transfer type before payload conversion.
+    /// </summary>
+    /// <remarks>
+    /// This is used by the SDK payload converter to delegate hook-aware values to the configured
+    /// payload converter using the transfer type representation provided by
+    /// <see cref="ITemporalTransferTypeConverter"/>.
+    /// This API is experimental and may change in a future release.
+    /// </remarks>
+    [AttributeUsage(AttributeTargets.Class | AttributeTargets.Struct, Inherited = true)]
+    public sealed class TemporalTransferTypeAttribute : Attribute
```
jmaeagle99: Rename to `TemporalTransferTypeConverterAttribute` to match .NET conventions.

src/Temporalio/Converters/TemporalTransferTypeAttribute.cs
```diff
@@ -0,0 +1,29 @@
...
+    /// <summary>
+    /// Marks a type as converting to and from a transfer type before payload conversion.
+    /// </summary>
+    /// <remarks>
+    /// This is used by the SDK payload converter to delegate hook-aware values to the configured
+    /// payload converter using the transfer type representation provided by
+    /// <see cref="ITemporalTransferTypeConverter"/>.
+    /// This API is experimental and may change in a future release.
+    /// </remarks>
+    [AttributeUsage(AttributeTargets.Class | AttributeTargets.Struct, Inherited = true)]
```
jmaeagle99: I think we should disallow inheritance (set `Inherited = false`). The reason being that if someone gets clever to create a base type with this attribute and a derived type without the attribute, then the converter on the base type better know how to dynamically inspect the transfer type to turn it back into the model type; seems like something easy to stumble into. With `Inherited = false`, each derivation needs to apply the attribute and declare which converter understands it; if the converter is polymorphic aware, it can be explicitly applied to each derivation.

I'd add a bunch of tests around this.

src/Temporalio/Converters/TemporalTransferTypePayloadConverter.cs
```diff
@@ -0,0 +1,102 @@
...
+                return this;
+            }
+
+            var contextInner = withContext.WithSerializationContext(context);
+            return ReferenceEquals(contextInner, inner) ? this : new TemporalTransferTypePayloadConverter(contextInner);
+        }
+
+        private static ITemporalTransferTypeConverter? CreateConverter(Type type)
+        {
+            var attr = type.GetCustomAttribute<TemporalTransferTypeAttribute>(inherit: true);
```
jmaeagle99: Flip to `false` to match https://github.com/temporalio/sdk-dotnet/pull/795/changes#r3641267623. I think we want the runtime type to actually declare it's converter rather than getting something via inheritance.

src/Temporalio/Converters/DataConverter.cs
```diff
@@ -12,6 +12,18 @@ public record DataConverter(
         IPayloadCodec? PayloadCodec = null)
     : IWithSerializationContext<DataConverter>
     {
+        private readonly IPayloadConverter payloadConverter =
+            TemporalTransferTypePayloadConverter.Wrap(PayloadConverter);
+
+        /// <summary>
+        /// Gets the payload converter.
+        /// </summary>
+        public IPayloadConverter PayloadConverter
```
jmaeagle99: I would not do this since it breaks the record semantics. I think if you construct two separate `DataConverter` instances with the same payload converter, failure converter, and payload codec, before this change they would appear to be equal but after this change they would appear to be unequal.

This becomes acute for plugins, which could try to wrap a payload converter from the DataConverter given to it with it's own. That'll cause an old `TemporalTransferTypePayloadConverter` to be sandwiched between a user-defined and a plugin-defined payload converter, and all wrapped by another `TemporalTransferTypePayloadConverter` payload converter. Maybe not detrimental, but definitely unexpected.

I would apply this at `TemporalClient.ConnectAsync` just after plugins do their adjustments and before we actually consume the options within the client infrastructure. The other place that `DataConverter` is provided is via `HeartbeatAsyncActivityInput`, but I'm not sure we really need to wrap that there. If just for system Nexus operations, probably not, but if as a general transfer type mechanism, then probably yes.

src/Temporalio/Converters/DataConverter.cs
```diff
@@ -12,6 +12,18 @@ public record DataConverter(
...
+        {
+            get => payloadConverter;
+            init => payloadConverter = TemporalTransferTypePayloadConverter.Wrap(value);
```
jmaeagle99: I don't think the init is necessary; everyone is already constructing the `DataConverter` by passing the payload converter via the constructor.

jmaeagle99 (changes requested):

src/Temporalio/Client/TemporalClient.cs
```diff
@@ -29,6 +30,7 @@ public TemporalClient(ITemporalConnection connection, TemporalClientOptions opti
                 }
             }
 
+            options.DataConverter = TemporalTransferTypePayloadConverter.Wrap(options.DataConverter);
```
jmaeagle99, line 34: I think we should clone the options and replace the data converter on the clone because if the caller looks at the data converter after constructing the client, they will see a different value on their options instance. We should probably make the clone before plugins operate on the options.

src/Temporalio/Client/TemporalClient.cs
jmaeagle99, line 23: I think this is looking better. One thing I'm thinking through is if the customer constructs another client based on the options that the first client has AND a plugin mutates the PayloadConverter, we'll have nested special internal payload converter within the chain and one as the very outermost instance. For example:

```cs
var client1 = new TemporalClient(connection, new TemporalClientOptions
{
    DataConverter = new DataConverter(inner, new DefaultFailureConverter()),
});

var client2 = new TemporalClient(connection, new TemporalClientOptions
{
    DataConverter = client1.Options.DataConverter,
    Plugins = new[] { plugin },
});
```

If `plugin` wraps the PayloadConverter during ConfigureClient, we end up with a PayloadConverter on client2 that looks like `W(P(W(inner)))` where `W` is the `TemporalTransferTypePayloadConverter` and `P` is the plugin payload converter. This is probably benign and not likely to happen all that often.

tests/Temporalio.Tests/Converters/PayloadConverterTests.cs
```diff
@@ -160,7 +228,7 @@ public class NoJsonProtoPayloadConverter : DefaultPayloadConverter
     {
         public NoJsonProtoPayloadConverter()
             : base(
-                ((DefaultPayloadConverter)DataConverter.Default.PayloadConverter).
+                new DefaultPayloadConverter().
```
jmaeagle99: This can probably be reverted.

tests/Temporalio.Tests/Worker/WorkflowWorkerTests.cs
```diff
@@ -7876,7 +7876,7 @@ public async Task ExecuteWorkflowAsync_ConverterContext_ProperlyAvailable()
         // Context-aware data converter and client
         var dataConverter = new DataConverter(
             PayloadConverter: new DefaultPayloadConverter(
-                ((DefaultPayloadConverter)Client.Options.DataConverter.PayloadConverter).EncodingConverters.Select(e =>
```
jmaeagle99, line 7879: This demonstrates part of the problem I wrote up in https://github.com/temporalio/sdk-dotnet/pull/795/changes#r3677825054. If we keep this behavior, maybe just use `DataConverter.Default.PayloadConverter` here.

src/Temporalio/Client/TemporalClient.cs
jmaeagle99: _If_ we want to completely fix this, might need to have an internal `ResolvedDataConverter` property on the options that is used internally throughout the library.

tconley1428: I think it seems marginal enough of a scenario that the cost of a wrapping object that doesn't do anything isn't prohibitive. We can return to it if we see need.

jmaeagle99 (approved):

PR #799, opened 2026-07-28
fix: mark system Nexus envelope payloads

tconley1428:
## What changed

- Mark System Nexus envelopes when they are created.
- Detect marked envelopes anywhere workflow codecs visit a payload, so nested user payloads are encoded and decoded after an envelope is stored in a generic payload field.
- Keep unmarked payloads on the regular codec path and preserve batch codec behavior for ordinary payload collections.

## Blocked on server behavior

This stacked draft depends on #846. It must not be completed until the server marks System Nexus
operation result payloads as System envelopes too. The server currently returns an unmarked
binary-protobuf completion payload, which a workflow codec correctly treats as an ordinary payload
and may transform, making the outer response no longer parseable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -115,11 +115,6 @@ to docs, or any other relevant information.
 - Try-cancel child workflows no longer cause nondeterminism when they complete or fail after their
   cancellation was requested.
 
-### Fixed
-
-- Marked system Nexus envelope payloads so nested payloads continue to be encoded and decoded after
-  the envelope is stored in a generic payload field.
-
 ### [1.17.0] - 2026-07-13
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -115,6 +115,11 @@ to docs, or any other relevant information.
 - Try-cancel child workflows no longer cause nondeterminism when they complete or fail after their
   cancellation was requested.
 
+### Fixed
+
+- Marked system Nexus envelope payloads so nested payloads continue to be encoded and decoded after
+  the envelope is stored in a generic payload field.
+
 ### [1.17.0] - 2026-07-13
 
 ### Added
--- a/src/Temporalio/Nexus/SystemNexusPayloadConverter.cs
+++ b/src/Temporalio/Nexus/SystemNexusPayloadConverter.cs
@@ -1,6 +1,7 @@
 using System;
 using Temporalio.Api.Common.V1;
 using Temporalio.Converters;
+using Temporalio.Worker;
 
 namespace Temporalio.Nexus
 {
@@ -38,9 +39,9 @@ namespace Temporalio.Nexus
         /// <inheritdoc />
         public Payload ToPayload(object? value)
         {
-            // TODO: Scope userPayloadConverter and userFailureConverter in the generated System
-            // Nexus support converter context once that support file is ingested into the SDK.
-            return OuterPayloadConverter.ToPayload(value);
+            var payload = OuterPayloadConverter.ToPayload(value);
+            SystemNexusPayloadVisitor.MarkSystemPayload(payload);
+            return payload;
         }
 
         /// <inheritdoc />
--- a/src/Temporalio/Worker/SystemNexusPayloadVisitor.cs
+++ b/src/Temporalio/Worker/SystemNexusPayloadVisitor.cs
@@ -16,6 +16,7 @@ namespace Temporalio.Worker
         private const string SystemPayloadMetadataKey = "__temporal_system_payload";
 
         private static readonly ByteString SystemPayloadMetadataValue = ByteString.CopyFromUtf8("true");
+
         internal delegate Task PayloadVisitor(Payload payload);
 
         internal delegate Task PayloadsVisitor(RepeatedField<Payload> payloads);
@@ -50,10 +51,8 @@ namespace Temporalio.Worker
             }
         }
 
-        private static bool IsSystemNexusPayload(Payload payload) =>
+        private static bool IsSystemPayload(Payload payload) =>
             payload.Metadata.TryGetValue(SystemPayloadMetadataKey, out var value) &&
-            value == SystemPayloadMetadataValue;
-
-        private static bool IsSystemNexusEndpoint(string? endpoint) => endpoint == TemporalSystemEndpoint;
+            value.Equals(SystemPayloadMetadataValue);
     }
 }
--- a/src/Temporalio/Worker/SystemNexusPayloadVisitor.cs
+++ b/src/Temporalio/Worker/SystemNexusPayloadVisitor.cs
@@ -21,9 +21,7 @@ namespace Temporalio.Worker
 
         internal delegate Task PayloadsVisitor(RepeatedField<Payload> payloads);
 
-        internal delegate Task EnvelopeVisitor(Payload payload);
-
-        internal static bool IsSystemNexusEndpoint(string? endpoint) => endpoint == TemporalSystemEndpoint;
+        internal static bool IsSystemEndpoint(string? endpoint) => endpoint == TemporalSystemEndpoint;
 
         internal static void MarkSystemPayload(Payload payload) =>
             payload.Metadata[SystemPayloadMetadataKey] = SystemPayloadMetadataValue;
@@ -32,8 +30,7 @@ namespace Temporalio.Worker
             Payload payload,
             Func<T, PayloadVisitor, PayloadsVisitor, Task> visitMessage,
             PayloadVisitor visitPayload,
-            PayloadsVisitor visitPayloads,
-            EnvelopeVisitor? visitEnvelope)
+            PayloadsVisitor visitPayloads)
             where T : IMessage<T>, new()
         {
             BinaryProtoConverter.AssertProtoPayload(payload, typeof(T));
@@ -45,10 +42,6 @@ namespace Temporalio.Worker
             payload.Metadata["messageType"] = ByteString.CopyFromUtf8(message.Descriptor.FullName);
             MarkSystemPayload(payload);
             payload.Data = message.ToByteString();
-            if (visitEnvelope != null)
-            {
-                await visitEnvelope(payload).ConfigureAwait(false);
-            }
         }
 
         private static bool IsSystemPayload(Payload payload) =>
--- a/src/Temporalio/Worker/SystemNexusPayloadVisitor.cs
+++ b/src/Temporalio/Worker/SystemNexusPayloadVisitor.cs
@@ -11,27 +11,21 @@ namespace Temporalio.Worker
 {
     internal static partial class SystemNexusPayloadVisitor
     {
+        private const string TemporalSystemEndpoint = "__temporal_system";
+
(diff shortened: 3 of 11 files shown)


src/Temporalio/Worker/SystemNexusPayloadVisitor.cs
```diff
@@ -70,11 +61,18 @@ private static async Task VisitEnvelopeAsync<T>(
...
             payload.Metadata["messageType"] = ByteString.CopyFromUtf8(message.Descriptor.FullName);
+            payload.Metadata[SystemPayloadMetadataKey] = SystemPayloadMetadataValue;
             payload.Data = message.ToByteString();
             if (visitEnvelope != null)
             {
                 await visitEnvelope(payload).ConfigureAwait(false);
             }
         }
+
+        private static bool IsSystemNexusPayload(Payload payload) =>
```
jmaeagle99: nit: Looks like these are being more generically named given the concept rename. So maybe update the method names:
- `IsSystemNexusPayload` -> `IsSystemPayload`
- `IsSystemNexusEndpoint` -> `IsSystemEndpoint`

src/Temporalio/Worker/WorkflowCodecHelper.cs
```diff
@@ -429,15 +410,43 @@ private static async Task EncodeAsync(
...
+            }
+            if (codecPayloadIndexes.Count == 0)
+            {
+                return;
+            }
+            // We have to convert to a list here just in case codec results are based on the
+            // underlying list and we clear it out below.
+            var codecPayloads = codecPayloadIndexes.Select(index => allPayloads[index]).ToList();
+            var newPayloads = (await codec.EncodeAsync(codecPayloads).ConfigureAwait(false)).ToList();
+            for (var i = 0; i < codecPayloadIndexes.Count; i++)
```
jmaeagle99: This breaks when a codec encodes N payloads into 1 payload.

src/Temporalio/Worker/WorkflowCodecHelper.cs
```diff
@@ -429,15 +410,43 @@ private static async Task EncodeAsync(
             {
                 return;
             }
-            // We have to convert to list here just in case they are based on the underlying list
-            // and we clear it out (which can happen with Linq selectors)
-            var newPayloads = (await codec.EncodeAsync(payloads).ConfigureAwait(false)).ToList();
+            var allPayloads = payloads.ToList();
+            var codecPayloadIndexes = new List<int>();
+            for (var i = 0; i < allPayloads.Count; i++)
```
jmaeagle99: Are we actually ever to expect that there would be a mix of system payloads and user payloads within a single `Payloads` proto or `Payload` sequence?

src/Temporalio/Worker/WorkflowCodecHelper.cs
```diff
@@ -125,20 +126,9 @@ await childCodec2.DecodeFailureAsync(
                         }
                         if (job.ResolveNexusOperation.Result.Completed != null)
                         {
-                            var operationInfo = context.Instance?.GetPendingNexusOperationInfo(
```
jmaeagle99, line 128: `GetPendingNexusOperationInfo` and `NexusOperationInfo` are likely dead code.

CHANGELOG.md
```diff
@@ -50,6 +50,11 @@ to docs, or any other relevant information.
   callbacks cannot mutate handlers or workflow details, invoke patches, or schedule workflow work.
   Patch activation callbacks also cannot use workflow randomness or issue workflow commands.
 
+### Fixed
+
+- Marked system Nexus envelope payloads so nested payloads continue to be encoded and decoded after
```
jmaeagle99: Is this really a user-facing change that they need to know in the release?

src/Temporalio/Worker/WorkflowCodecHelper.cs
```diff
@@ -125,20 +126,9 @@ await childCodec2.DecodeFailureAsync(
...
-                                    payload => DecodeAsync(nexusCodec, payload),
-                                    payloads => DecodeAsync(nexusCodec, payloads)).
-                                    ConfigureAwait(false))
-                            {
-                                await DecodeAsync(
-                                    nexusCodec, job.ResolveNexusOperation.Result.Completed).
-                                    ConfigureAwait(false);
-                            }
+                            await DecodeAsync(
+                                nexusCodec, job.ResolveNexusOperation.Result.Completed).
```
jmaeagle99, line 130: This payload possibly does not have the system payload marker on it even if it is a system payload. Which means that it'll go through the normal user codec decode path and then fail deserialization causing the workflow to continuously fail the task and timeout the ExecuteWorkflowAsync_SignalWithStartFromWorkflow_Succeeds test. Might need to check for the marker or see if it's from the system nexus endpoint; or maybe somehow stamp the payload with the marker if it comes from the system nexus endpoint before we attempt the decode here.

tests/Temporalio.Tests/Worker/WorkflowCodecHelperTests.cs
```diff
@@ -104,6 +106,98 @@ await CreateAndVisitPayload(new(), comp, async (ctx, payload) =>
         });
     }
 
+    [Fact]
```
jmaeagle99, line 109: Probably need some more tests for encode and decode mixed batch of user payloads and system payloads, a test for when they are all system payloads, and a server response decoding test (which would catch the problem described in https://github.com/temporalio/sdk-dotnet/pull/799/changes#r3733469302).

src/Temporalio/Worker/WorkflowCodecHelper.cs
tconley1428: In what scenario would it be a system payload without the system payload marker? That should be tautologically equivalent as far as I know. (At least once server is updated sufficiently) Are you talking backwards compat or something else?

tests/Temporalio.Tests/Worker/WorkflowCodecHelperTests.cs
tconley1428: I don't think a mixed batch is realistic, but also worth covering just in case.

src/Temporalio/Worker/WorkflowCodecHelper.cs
tconley1428: I don't think so, but it would be good to be resilient anyway.

src/Temporalio/Worker/WorkflowCodecHelper.cs
jmaeagle99: Back compat. Today, system payloads are identified by the endpoint, but this change completely drops that (I think). So in-flight workflows that do not have the new marker now going through the path of decoding the envelop. But if you don't care about those, then I think this is fine.

But I see that `ExecuteWorkflowAsync_SignalWithStartFromWorkflow_Succeeds` hung in every job. It's likely retrying a workflow task indefinitely for some reason. Probably due to the missing marker.

tconley1428: Yeah, this was in draft for a reason. All you say is true, but it's also not enabled at all yet, so we're good to do the breaking change. Nice catch though.

tconley1428: Change has been made to server now.

src/Temporalio/Nexus/SystemNexusPayloadConverter.cs
```diff
@@ -38,9 +39,9 @@ internal SystemNexusPayloadConverter(
         /// <inheritdoc />
         public Payload ToPayload(object? value)
         {
-            // TODO: Scope userPayloadConverter and userFailureConverter in the generated System
-            // Nexus support converter context once that support file is ingested into the SDK.
-            return OuterPayloadConverter.ToPayload(value);
+            var payload = OuterPayloadConverter.ToPayload(value);
+            SystemNexusPayloadVisitor.MarkSystemPayload(payload);
```
jmaeagle99, line 43: Do we need to skip `RawValue`?

src/Temporalio/Worker/SystemNexusPayloadVisitor.cs
jmaeagle99: Is there any purpose for the EnvelopeVisitor any more?

src/Temporalio/Worker/WorkflowCodecHelper.cs
```diff
@@ -429,15 +410,43 @@ private static async Task EncodeAsync(
...
                 return;
             }
-            // We have to convert to list here just in case they are based on the underlying list
-            // and we clear it out (which can happen with Linq selectors)
-            var newPayloads = (await codec.EncodeAsync(payloads).ConfigureAwait(false)).ToList();
+            var newPayloads = new List<Payload>();
+            var codecPayloads = new List<Payload>();
+            foreach (var payload in payloads)
+            {
+                if (!await SystemNexusPayloadVisitor.TryVisitAsync(
```
jmaeagle99, line 417: This is mutating the payload instance in-place. The original payload instances are modified. This could be problematic if anything inspects the original payloads after this is applied. Might want to consider updating this visitor to pass the payload by `ref` and have it construct a new payload instance _if_ it is a system nexus payload.

src/Temporalio/Worker/SystemNexusPayloadVisitor.cs
```diff
@@ -11,27 +11,22 @@ namespace Temporalio.Worker
...
         internal delegate Task EnvelopeVisitor(Payload payload);
 
-        internal static Task<bool> TryVisitInputAsync(
-            string? endpoint,
-            Payload payload,
-            PayloadVisitor visitPayload,
-            PayloadsVisitor visitPayloads,
-            EnvelopeVisitor? visitEnvelope = null) =>
-            TryVisitAsync(endpoint, payload, visitPayload, visitPayloads, visitEnvelope);
+        internal static bool IsSystemNexusEndpoint(string? endpoint) => endpoint == TemporalSystemEndpoint;
```
jmaeagle99: `IsSystemEndpoint`

jmaeagle99 (approved):

src/Temporalio/Nexus/SystemNexusPayloadConverter.cs
tconley1428: I don't _think_ so. I suppose if we really did that we would expect the marking to already be there and do nothing but it seems super unlikely. I'd rather not introduce the logic unless it really will end up doing something.

src/Temporalio/Worker/SystemNexusPayloadVisitor.cs
tconley1428: Yes agreed

src/Temporalio/Worker/SystemNexusPayloadVisitor.cs
tconley1428: Maybe not now that payload sizing is moved into core

src/Temporalio/Worker/WorkflowCodecHelper.cs
tconley1428: Not a new concern, and not one which has any current ramifications I believe.

PR #804, opened 2026-08-03
Add Workflow Streams extension library

### LINKED ISSUE #803: [Feature Request] Support for Workflow Streaming in the .NET SDK

The .NET SDK currently does not appear to support **workflow streaming**, making it difficult to build responsive applications that can surface workflow progress, intermediate events, or partial results to users in real time.

This creates a gap compared to other SDKs and limits the ability to provide a rich streaming experience in .NET applications.

I'd like the .NET SDK to support **workflow streaming**, allowing developers to subscribe to workflow execution events and receive incremental updates as the workflow progresses.

Ideally, the API would provide a streaming interface (e.g., `IAsyncEnumerable`, async streams, or another .NET-native abstraction) that exposes workflow events and partial outputs while the workflow is running.

Additionally, it would be helpful to know whether this feature is already planned and, if so, if there is an estimated timeline for its availability.

### Additional context

Our team is building applications on top of the .NET SDK and would benefit greatly from streaming workflow updates to the frontend, improving user feedback during long-running workflow executions for agentic.

If this feature is already on the roadmap, we'd appreciate any information you can share regarding its planned availability.

gabrielschmith:
Introducing the `Temporalio.Extensions.WorkflowStreams` project, which provides a durable publish/subscribe mechanism hosted inside Temporal workflows. Includes comprehensive setup instructions, publishing/subscribing capabilities, and cross-language compatibility.

## What was changed

Introduced the new `Temporalio.Extensions.WorkflowStreams` project, providing a durable publish/subscribe abstraction built on top of Temporal Workflows.

This implementation includes:

- Durable Workflow Stream implementation hosted entirely within Temporal.
- Publish/Subscribe APIs for producers and consumers.
- Support for multiple subscribers per stream.
- Durable message delivery leveraging Temporal workflow state.
- Cross-language compatibility by exposing stream interactions through Temporal primitives.
- Dependency injection extensions and configuration helpers.
- Documentation and usage examples covering setup, publishing, and subscribing.

## Why?

Many applications built on Temporal require long-lived event distribution between workflows and external clients. While Temporal provides durable execution, there is currently no built-in abstraction for durable publish/subscribe communication.

`WorkflowStreams` fills this gap by providing a reusable and reliable streaming mechanism that:
- Enables workflows to communicate through durable event streams.
- Simplifies event-driven workflow architectures.
- Removes the need for external messaging infrastructure in scenarios where Temporal durability is sufficient.
- Provides a consistent developer experience across supported SDKs.

## Checklist

1. Closes #803 

2. How was this tested:

- Added unit tests covering the core stream implementation.
- Validated publish and subscribe scenarios.
- Verified durable delivery across workflow execution.
- Tested multiple subscribers receiving published events.
- Validated end-to-end examples using the .NET SDK.

3. Any docs updates needed?

- Installation instructions.
- Creating a Workflow Stream.
- Publishing events.
- Subscribing to events.
- Complete end-to-end examples.
- Best practices and limitations.
- Cross-language interoperability notes.


brianstrauch:
Thanks for the PR! Updated it with current upstream main and a hardening pass.

Changes:
- Added release artifact, API documentation, README, changelog, and first-release package-validation integration.
- Added AI SDK ownership for the Workflow Streams package and tests.
- Made subscription cancellation interrupt in-flight update RPCs and honored async-enumeration cancellation.
- Rejected publishing after client closure and preserved fresh publisher sequence numbers after retry timeout.
- Regenerated lockfiles and added canonical cross-SDK payload, response paging, parked-poll truncation, cancellation, lifecycle, and reserved-handler tests.

Validation:
- dotnet format --verify-no-changes: passed
- Workflow Streams tests: 49/49 passed
- Reserved-handler regression: passed
- dotnet pack -c Debug: passed across all target frameworks
- DocFX with warnings as errors: passed with 0 warnings
- Full suite: 783 passed and 13 skipped; one unrelated shared-server failure passed when rerun on a fresh server

Commits: 40464d4, 93a70c7.

brianstrauch:
While this PR is going through review, can you attribute your first commit to your GitHub account and sign the CLA? @gabrielschmith

src/Temporalio/Workflows/WorkflowQueryDefinition.cs
```diff
@@ -24,7 +24,7 @@ public class WorkflowQueryDefinition
 
         private WorkflowQueryDefinition(string? name, string? description, MethodInfo? method, Delegate? del)
         {
-            if (name != null)
+            if (name != null && !name.StartsWith(TemporalRuntime.ReservedWorkflowStreamNamePrefix))
```
moedash, line 27: This wraps the whole prefix loop, so a `__temporal_workflow_stream_` name also skips the `__stack_trace` check. Java keeps them apart though: the exemption covers only `__temporal_`.

IMO, let's match that:

```csharp
if (name != null)
{
    var reservedQ = ReservedQueryHandlerPrefixes.FirstOrDefault(p => name.StartsWith(p));
    if (!string.IsNullOrEmpty(reservedQ) &&
        !(reservedQ == TemporalRuntime.ReservedNamePrefix &&
          name.StartsWith(TemporalRuntime.ReservedWorkflowStreamNamePrefix)))
    {
        throw new ArgumentException($"Query handler name {name} cannot start with {reservedQ}");
    }
}
```

A wider point is that this opens a hole in core SDK validation from an extension. Go and Java use a prefix, TypeScript an exact-name allowlist keyed by handler kind. Currently, as written, anyone can register `__temporal_workflow_stream_anything` as any kind. We need a review from a SDK team member on this.

src/Temporalio.Extensions.WorkflowStreams/Internal/SubscriptionDriver.cs
```diff
@@ -0,0 +1,325 @@
...
+            if (status == WorkflowExecutionStatus.Completed ||
+                status == WorkflowExecutionStatus.Failed ||
+                status == WorkflowExecutionStatus.Canceled ||
+                status == WorkflowExecutionStatus.Terminated ||
+                status == WorkflowExecutionStatus.TimedOut)
+            {
+                FinishCompleted();
+                return false;
+            }
+            FinishError(e);
```
moedash, line 296: Any poll failure where `DescribeAsync` says `Running` ends the subscription. There won't be a retry.

Java's `handleError` and Go's `Subscribe` do the same, so I'm not asking you to diverge here.

But .NET looks more exposed. `WorkflowUpdateHandle.PollUntilOutcomeAsync` rethrows `DeadlineExceeded` as `WorkflowUpdateRpcTimeoutOrCanceledException`, which is in this branch. Java's update future and Go's `handle.Get` re-poll internally instead.

Can you check what deadline `PollWorkflowExecutionUpdate` gets through core? If it can surface, I'd rather handle it here, then file the same against `sdk-java` and `sdk-go`. Please add a test that idles past the deadline.

src/Temporalio.Extensions.WorkflowStreams/Internal/SubscriptionDriver.cs
```diff
@@ -0,0 +1,325 @@
...
+                    Rpc = new RpcOptions { CancellationToken = closeCts.Token },
+                }).ConfigureAwait(false)).Status;
+            }
+            catch (Exception)
+            {
+                status = WorkflowExecutionStatus.Unspecified;
+            }
+            if (status == WorkflowExecutionStatus.ContinuedAsNew)
+            {
+                return true;
```
moedash, line 285: `StreamDraining` above backs off by `pollCooldown`. This returns immediately. Java does the same, so it's doing the same.

Flagging it because every poll spends one of the successor's `history.maxTotalUpdates` (default 2000 for that execution). A spin here consumes the budget of the run.

Do you know if the Java side did that on purpose? If not, the same `pollCooldown` backoff here, plus an issue on `sdk-java` and `sdk-go`, seems worth doing.

src/Temporalio.Extensions.WorkflowStreams/Internal/SubscriptionDriver.cs
```diff
@@ -0,0 +1,325 @@
...
+                }
+            }
+            if (appFailure != null)
+            {
+                if (appFailure.ErrorType == WorkflowStreamConstants.ErrorTypeTruncatedOffset)
+                {
+                    // Fell behind truncation; restart from the beginning of whatever still
+                    // exists, with an immediate repoll.
+                    offset = 0;
+                    return true;
```
moedash, line 247: `offset = 0` and carry on. The listener never learns it lost items, it just sees offsets jump.

Matches Java and Go, so I'm not asking you to change the behaviour. Still hard to debug in production. Worth a log line here, or a README note that offsets can jump after truncation?

src/Temporalio.Extensions.WorkflowStreams/Internal/StreamPublisher.cs
```diff
@@ -0,0 +1,346 @@
...
+        private async Task BackgroundLoopAsync()
+        {
+            while (true)
+            {
+                Task wakeTask;
+                lock (stateLock)
+                {
+                    wakeTask = wakeTcs.Task;
+                }
+                var delayTask = Task.Delay(batchInterval, backgroundCts.Token);
```
moedash: When `wakeTask` wins, `delayTask` is dropped without cancelling. Its timer and its token registration live on for the rest of `batchInterval`. At the 2s default with frequent `forceFlush`, those pile up on a `CancellationTokenSource` that lasts as long as the client.

Go uses one `time.Ticker`, Java one `scheduleWithFixedDelay`. This is the only package rebuilding the timer per turn.

A single reusable timer instead? Or a per-iteration linked CTS you cancel when the wake wins.

Also `backgroundCts` is never disposed. The `CA1001` suppression says disposal adds nothing, but that assumes nothing registers on the token.

src/Temporalio.Extensions.WorkflowStreams/WorkflowStream.cs
```diff
@@ -0,0 +1,362 @@
...
+        /// <paramref name="publisherTtl" />.
+        /// </summary>
+        /// <param name="publisherTtl">How long per-publisher dedup state is retained since the
+        /// publisher's last accepted batch.</param>
+        /// <returns>Stream state snapshot.</returns>
+        public WorkflowStreamState GetState(TimeSpan publisherTtl)
+        {
+            var now = NowUnixSeconds();
+            var state = new WorkflowStreamState { BaseOffset = baseOffset };
+            foreach (var pair in publisherSequences)
```
moedash, line 184: `GetState` prunes the snapshot, but `publisherSequences` and `publisherLastSeen` never shrink. Only a rollover clears them.

Each `WorkflowStreamClient` mints a fresh publisher ID, so one short-lived client per activity attempt means one entry per attempt, forever. Then we try to fit all of it into the continue-as-new args.

Is it worth pruning in place on publish, rather than only at snapshot time?

src/Temporalio.Extensions.WorkflowStreams/WorkflowStream.cs
```diff
@@ -0,0 +1,362 @@
...
+                if (now - lastSeen < publisherTtl.TotalSeconds)
+                {
+                    state.PublisherSequences[pair.Key] = pair.Value;
+                    state.PublisherLastSeen[pair.Key] = lastSeen;
+                }
+            }
+            foreach (var entry in log)
+            {
+                // Per-item offset is re-derivable from baseOffset + index on reload.
+                state.Log.Add(new WireItem
```
moedash, line 196: The whole retained log goes into the continue-as-new args. That's one payload, so the 2 MB blob limit applies with the codec on top. Increases the chance of rollover failure, at rollover time.

`Truncate` can avoid it, but it's not in README.

Every poll spends one of the workflow's `history.maxTotalUpdates` (2000 per execution, whole life), so read traffic forces rollovers on a schedule. We pass through here often, and each pass has to fit the log in one payload. The two limits pull against each other.

Can you add a README section with a `Truncate` call and the payload limit? A rollover test with a large log would be great (to show the limit).

src/Temporalio.Extensions.WorkflowStreams/WorkflowStream.cs
```diff
@@ -0,0 +1,362 @@
...
+            {
+                var entry = log[(int)i];
+                if (topicSet != null && (entry.Topic == null || !topicSet.Contains(entry.Topic)))
+                {
+                    continue;
+                }
+                var globalOffset = baseOffset + i;
+                var encoded = PayloadWire.Encode(entry.Payload);
+                var itemSize = PayloadWire.WireSize(encoded, entry.Topic ?? string.Empty);
+                if (size + itemSize > WorkflowStreamConstants.MaxPollResponseBytes && items.Count > 0)
```
moedash, line 350: `items.Count > 0` means the first item always goes in, whatever its size.

`WorkflowTopicHandle.Publish` appends straight to the log with no size check. So a workflow can publish an item bigger than the response limit, and then every poll reaching that offset builds a response the server rejects. The subscriber retries forever. Only a truncate past it gets out.

The signal path is covered by the request limit already. Should workflow-side publish reject an oversized value, or should the poll return a typed error so the subscriber can skip the offset?

src/Temporalio.Extensions.WorkflowStreams/WorkflowStreamClient.cs
```diff
@@ -0,0 +1,271 @@
...
+        /// <summary>
+        /// Closes the client synchronously; see <see cref="CloseAsync" />, which is preferred
+        /// over this blocking form.
+        /// </summary>
+        public void Dispose()
+        {
+            // Safe to block: every await inside CloseAsync uses ConfigureAwait(false), so no
+            // synchronization context can be required to finish it.
+#pragma warning disable VSTHRD002 // Avoid problematic synchronous waits
+            CloseAsync().GetAwaiter().GetResult();
```
moedash, line 244: `Dispose` blocks on `CloseAsync`, which can throw `FlushTimeoutException`. So `using var client = ...` can throw from the closing brace, and can replace an exception already propagating from the body.

The README activity example pairs `using` with an explicit `await client.CloseAsync()`, which seems like the `using` is what flushes.

Can the examples use `await using`, or `CloseAsync` in a `finally`, with the docs saying `Dispose` is the last resort?

src/Temporalio.Extensions.WorkflowStreams/WorkflowStreamClient.cs
```diff
@@ -0,0 +1,271 @@
...
+        /// continue-as-new chains in its subscriptions.
+        /// </summary>
+        /// <param name="client">Temporal client to communicate through.</param>
+        /// <param name="workflowId">Workflow ID hosting the stream.</param>
+        /// <param name="options">Client options, or null for defaults.</param>
+        public WorkflowStreamClient(
+            ITemporalClient client, string workflowId, WorkflowStreamClientOptions? options = null)
+        {
+            this.client = client;
+            this.workflowId = workflowId;
```
moedash, line 61: `Subscribe` null-checks its arguments, the constructor doesn't. A null `client` here surfaces later as a `NullReferenceException` inside the publisher's background loop. Add the same guards?

src/Temporalio.Extensions.WorkflowStreams/WorkflowStreamSubscription.cs
```diff
@@ -0,0 +1,254 @@
...
+        /// <paramref name="cancellationToken" /> is canceled.
+        /// </summary>
+        /// <param name="cancellationToken">Token that stops the subscription and any in-flight
+        /// poll.</param>
+        /// <returns>True if <see cref="Current" /> holds a new item; false at the end of the
+        /// stream or when canceled.</returns>
+        public async Task<bool> MoveNextAsync(CancellationToken cancellationToken)
+        {
+            cancellationToken.ThrowIfCancellationRequested();
+            using (cancellationToken.Register(Dispose))
```
moedash, line 142: Cancelling disposes the subscription, so `MoveNextAsync` returns `false` and `await foreach` exits normally. Everywhere else in .NET, including `ListWorkflowsAsync` in this SDK, `WithCancellation` throws `OperationCanceledException`.

It also makes a cancelled subscription look identical to a completed workflow. Java has no cancellation on its pull path, so this is a .NET choice.

Is it deliberate? If so the XML doc should say it, since it's the reverse of the platform default.

src/Temporalio.Extensions.WorkflowStreams/WireItem.cs
```diff
@@ -0,0 +1,37 @@
...
+    /// with exactly these names.
+    /// </para>
+    /// </remarks>
+    public class WireItem
+    {
+        /// <summary>
+        /// Gets or sets the topic this item was published on. Null means no topic.
+        /// </summary>
+        [JsonPropertyName("topic")]
+        public string? Topic { get; set; }
```
moedash: `string?` with "null means no topic". None of the other four SDKs do it that way:

- Go: `Topic string`
- Java: `public String topic`, no null contract
- Python: `topic: str`
- TypeScript: `topic: string`

`""` is the no-topic value everywhere else.

It leaks into behaviour too. `HandlePollAsync` drops a null topic whenever a filter is set, and `SubscribeOptions.Topics` is non-nullable, so a .NET subscriber can never get a no-topic item while a Go one can by asking for `""`.

Make it a plain `string` with `""`? If it stays nullable, `WireProtocolTests` should handle what an empty topic serializes to.

tests/Temporalio.Tests/Extensions/WorkflowStreams/ListenerSubscribeTests.cs
```diff
@@ -0,0 +1,471 @@
...
+                var subscriptionHandle = streamClient.Subscribe(FastPoll(), listener);
+                await AwaitItemsAsync(listener, 1);
+
+                subscriptionHandle.Dispose();
+                await subscriptionHandle.Completion.WaitAsync(Timeout);
+
+                streamClient.Topic("evt").Publish("b", forceFlush: true);
+                await streamClient.FlushAsync();
+                // Bounded grace window to catch any delivery that would wrongly follow Dispose.
+                await Task.Delay(300);
```
moedash: Four fixed `Task.Delay(300)` windows in this file, plus the 200/300 pacing in `StreamPublisherTests`. 300ms isn't much on a loaded runner, and this repo has already had to deflake timing tests.

This one is deterministic though. After `Dispose` the driver is finished and `subscriptionHandle.Completion` has already completed here. Publish "b", wait for `GetOffsetAsync()` to report 2, then assert `Assert.Single(listener.Items)`?

src/Temporalio.Extensions.WorkflowStreams/README.md
```diff
@@ -0,0 +1,206 @@
...
+Items are buffered and flushed automatically every batch interval (default 2s),
+when the buffer reaches the max batch size, on `forceFlush`, on an explicit
+`FlushAsync()`, or on `CloseAsync()`.
+
+## Subscribing
+
+There are two subscriber APIs over one shared poll engine: a non-blocking
+listener and an async iterator. Polling is fully async — no thread is occupied
+while a poll is blocked on the server — so many concurrent subscriptions do not
+mean many threads. Either way, the subscription ends cleanly when the workflow
```
moedash, line 123: Threads aren't what caps subscriber count. The server is.

Each subscriber holds one long-poll update open. `history.maxInFlightUpdates` defaults to 10 per execution, so a stream serves at most 10 concurrent subscribers. Past that it starves rather than degrades.

`history.maxTotalUpdates` is 2000 for an execution's whole life, and every poll spends one whether it returns an item or not. At the default 100ms `PollCooldown` one active subscriber burns about 500 a minute.

So continue-as-new isn't optional here. The README calls it something "for workflows that use continue-as-new". It's the only way to reset the poll counter, and read traffic forces it.

Can both limits go in this section? `Workflow.ContinueAsNewSuggested` looks like the right trigger, since the server flips it at 90% of the budget (`history.maxTotalUpdates.suggestContinueAsNewThreshold`).

src/Temporalio.Extensions.WorkflowStreams/SubscribeOptions.cs
```diff
@@ -0,0 +1,32 @@
...
+        /// Gets or sets the global offset to start from. Zero (the default) means the beginning
+        /// of whatever still exists.
+        /// </summary>
+        public long FromOffset { get; set; }
+
+        /// <summary>
+        /// Gets or sets the minimum interval between polls when no more items are immediately
+        /// ready. Default: 100 milliseconds.
+        /// </summary>
+        public TimeSpan PollCooldown { get; set; } = WorkflowStreamConstants.DefaultPollCooldown;
```
moedash, line 30: `PollCooldown` is a spend rate, not just a latency knob. Every poll consumes one of the workflow's `history.maxTotalUpdates` (2000 per execution, whole life). At 100ms a busy subscriber burns about 500 a minute, so this default sets how often the host has to continue-as-new.

Worth saying in the XML doc, so someone tuning it down to 10ms knows what they're spending. Is a higher default worth considering?

src/Temporalio.Extensions.WorkflowStreams/WorkflowStreamConstants.cs
```diff
@@ -0,0 +1,77 @@
...
+        /// Application failure type the poll update's validator returns while the stream is
+        /// detaching for continue-as-new. It tells a subscriber the rollover is in progress so it
+        /// retries (rather than surfacing an error) until the poll lands on the successor run.
+        /// </summary>
+        public const string ErrorTypeStreamDraining = "StreamDraining";
+
+        /// <summary>
+        /// Default interval between automatic client-side publisher flushes.
+        /// </summary>
+        public static readonly TimeSpan DefaultBatchInterval = TimeSpan.FromSeconds(2);
```
moedash: These four are `public`. Java keeps them package-private, Go unexported. Only handler names and error types are public there, which is right: those are the contract, tuning defaults aren't.

`EnablePackageValidation` is on, so once the first baseline publishes, `ApiCompat` locks these and removing one breaks. `MaxPollResponseBytes` is already `internal`, so the file is inconsistent with itself.

Make them `internal`?

src/Temporalio.Extensions.WorkflowStreams/Internal/StreamPublisher.cs
```diff
@@ -0,0 +1,346 @@
...
+                {
+                    // The pending batch was dropped and can't be recovered. Stash the error so
+                    // FlushAsync/CloseAsync surface it and stop the loop. Clearing started lets
+                    // the next Publish restart the loop; without it the publisher would silently
+                    // buffer forever after a single timeout. EnsureStartedLocked still refuses to
+                    // restart once closed.
+                    lock (stateLock)
+                    {
+                        deferredError = e;
+                        started = false;
```
moedash, line 346: Two deliberate differences from Java and Go live in this file. Both look right to me.

The restart is a correct fix. Java stashes the error and calls `scheduler.shutdown()` without clearing `started`, so the loop is dead for the client's life and items ship only on an explicit `flush()`. Go does the same minus the shutdown. `FlushTimeout_InBackgroundLoop_DoesNotStopLaterPublishes` tests this.

`Publish` after close throws. Java and Go buffer and drop. Better, but observable: a program ported from Java starts throwing from a `void` call in a loop.

Can both go in the README's cross-language section? The restart is also worth reporting to `sdk-java` and `sdk-go`, since both have the gap.

moedash (commented):
Thanks! It's a port of the Java module. Where .NET differs from it:

- Things that should change:

  - `Topic` is `string?` with "null means no topic". Java uses a plain string with `""`. .Net would emit `"topic": null`, and it leaks into the topic filter.
  - The four default `TimeSpan`s on `WorkflowStreamConstants` are public. Java keeps them package-private, Go unexported. `EnablePackageValidation` locks our choice after the first release.
  - The publisher rebuilds a `Task.Delay` every turn. Go uses one `time.Ticker`, Java one `scheduleWithFixedDelay`. The `CancellationTokenSource` those register on is never disposed.

- It's OK to keep, but should be documented (if intentional):

  - Restarting the flush loop after `FlushTimeoutException`. Java and Go both stop publishing for the client's life instead.
  - `Publish` throwing after close.

  Both belong in the README's cross-language section.

- Shared gaps (not .Net (this PR) alone):

  - Ending a subscription on any poll failure where the workflow reports `Running`, no backoff on the continue-as-new retry, truncation silently rewinding. All three match Java and Go. The first has a .NET difference: `WorkflowUpdateRpcTimeoutOrCanceledException` may surface where Java and Go re-poll internally.

Also, the reserved-name carve-out in `WorkflowQueryDefinition` needs an SDK team review. And the README should state two limits: `history.maxInFlightUpdates` caps a stream at 10 concurrent subscribers, and `history.maxTotalUpdates` gives an execution 2000 polls for its whole life, which makes continue-as-new mandatory.

src/Temporalio/Workflows/WorkflowQueryDefinition.cs
brianstrauch: Will standardize this fix across all SDK languages in v2

src/Temporalio.Extensions.WorkflowStreams/Internal/SubscriptionDriver.cs
brianstrauch: Will standardize this fix across all SDK languages in v2

src/Temporalio.Extensions.WorkflowStreams/Internal/SubscriptionDriver.cs
brianstrauch: Will standardize this fix across all SDK languages in v2

src/Temporalio.Extensions.WorkflowStreams/WorkflowStream.cs
brianstrauch: Will standardize this fix across all SDK languages in v2

src/Temporalio.Extensions.WorkflowStreams/WorkflowStream.cs
brianstrauch: Will standardize this fix across all SDK languages in v2

src/Temporalio.Extensions.WorkflowStreams/README.md
brianstrauch: Will standardize this fix across all SDK languages in v2

src/Temporalio.Extensions.WorkflowStreams/SubscribeOptions.cs
brianstrauch: Will standardize this fix across all SDK languages in v2

moedash (approved):
LGTM.

src/Temporalio/Workflows/WorkflowSignalDefinition.cs
```diff
@@ -20,7 +20,9 @@ private WorkflowSignalDefinition(
             Delegate? del,
             HandlerUnfinishedPolicy unfinishedPolicy)
         {
-            if (name != null && name.StartsWith(TemporalRuntime.ReservedNamePrefix))
+            if (name != null &&
```
jmaeagle99, line 23: These are punching exceptions through the deny list for the named prefix. But that also means anyone can now bypass if they use the same prefix e.g. `__temporal_workflow_stream_hahaha_i_win`. I think other SDKs (correct me if I'm wrong) check for exact names.

src/Temporalio/Workflows/WorkflowQueryDefinition.cs
jmaeagle99: I do not think this should be deferred and is not fixable in a "v2" of the extension since this is in the `Temporalio` library.

src/Temporalio.Extensions.WorkflowStreams/WorkflowStreamClient.cs
```diff
@@ -0,0 +1,291 @@
...
+            driver.Start();
+            return new WorkflowStreamSubscriptionHandle(driver);
+        }
+
+        /// <summary>
+        /// Stops the background publisher and drains any remaining items, guaranteeing a final
+        /// flush. It surfaces any deferred <see cref="FlushTimeoutException" /> from a prior
+        /// background flush failure. Also stops this client's live subscriptions (their
+        /// <see cref="WorkflowStreamSubscriptionHandle.Completion" /> tasks complete normally,
+        /// without <see cref="WorkflowStreamListener.OnCompleted" />). Idempotent.
```
jmaeagle99, line 204: Claims idempotency when clearly not. Maybe it _is_ if all methods of all of the implementations are idempotent, but that is extremely fragile.

src/Temporalio.Extensions.WorkflowStreams/Internal/StreamPublisher.cs
```diff
@@ -0,0 +1,372 @@
...
+            }
+            ThrowDeferred();
+        }
+
+        /// <summary>
+        /// Stops the background flush loop and drains any remaining items, surfacing a deferred
+        /// <see cref="FlushTimeoutException" /> from a prior background failure. Idempotent.
+        /// </summary>
+        /// <returns>Task completing when the drain finished.</returns>
+        public async Task CloseAsync()
```
jmaeagle99, line 257: All of the `CloseAsync` methods feel like a transliteration from Java/Go instead of doing what's natural in C#. Many of these should be removed and split into more idiomatic naming and concepts in C#, like `FlushAsync` and `DisposeAsync`. `FlushAsync` can take a `TimeSpan` or a `CancellationToken` (two separate overloads) if it needs to be time bound or cancellable. `DisposeAsync` should be more about terminal stoppage.

src/Temporalio.Extensions.WorkflowStreams/TopicHandle.cs
```diff
@@ -0,0 +1,80 @@
...
+
+        /// <summary>
+        /// Returns a subscription over items on this topic, starting at
+        /// <paramref name="fromOffset" />. See
+        /// <see cref="WorkflowStreamClient.Subscribe(SubscribeOptions?)" />.
+        /// </summary>
+        /// <param name="fromOffset">Global offset to start from. Zero means the beginning of
+        /// whatever still exists.</param>
+        /// <returns>The subscription.</returns>
+#pragma warning disable VSTHRD200 // Name matches the other SDKs' workflow streams packages; subscribing is not itself async work
```
jmaeagle99, line 54: I highly disagree with this rationale for suppressing this. APIs should be designed to be idiomatic in the language and not merely for "it's named the same". This probably should be named `SubscribeAsync`.

src/Temporalio.Extensions.WorkflowStreams/TopicHandle.cs
```diff
@@ -0,0 +1,80 @@
...
+        public WorkflowStreamSubscription Subscribe(long fromOffset = 0) =>
```
jmaeagle99, line 55: This should return `IAsyncEnumerable<WorkflowStreamItem>`.

src/Temporalio.Extensions.WorkflowStreams/WorkflowStreamSubscription.cs
```diff
@@ -0,0 +1,259 @@
...
+        /// Returns this subscription as its own enumerator. The subscription is single-use, so
+        /// this may only be called once (typically via <c>await foreach</c>).
+        /// </summary>
+        /// <param name="cancellationToken">Token that stops iteration and any in-flight poll.
+        /// </param>
+        /// <returns>This subscription.</returns>
+        public IAsyncEnumerator<WorkflowStreamItem> GetAsyncEnumerator(
+            CancellationToken cancellationToken = default)
+        {
+            if (Interlocked.Exchange(ref enumeratorTaken, 1) != 0)
```
jmaeagle99, line 177: This prevents multiple iterations of the same enumerable. That's quite a contract violation and very unexpected.

src/Temporalio.Extensions.WorkflowStreams/WorkflowStreamSubscription.cs
```diff
@@ -0,0 +1,259 @@
...
+    /// <remarks>
+    /// WARNING: This API is experimental and may change.
+    /// <para>
+    /// <see cref="Dispose" /> stops the subscription and interrupts a poll blocked on the server.
+    /// Items already buffered still drain after <see cref="Dispose" />.
+    /// </para>
+    /// </remarks>
+#if NETCOREAPP3_0_OR_GREATER
+    public sealed class WorkflowStreamSubscription :
+        IDisposable, IAsyncEnumerable<WorkflowStreamItem>, IAsyncEnumerator<WorkflowStreamItem>
```
jmaeagle99, line 27: If this is going to remain implementing `IAsyncEnumerable<WorkflowStreamItem>` and `IAsyncEnumerator<WorkflowStreamItem>`, then a lot more care needs to be taken here. Problems are:
- `Current` is shared by all instances of returned by `GetAsyncEnumerator`.
- `Current` is initially `null` on a non-nullable property.
- `enumerationCancellationToken` is mutably updated by `GetAsyncEnumerator`, stomping on any other call to `GetAsyncEnumerator`.
- Unsure why `public async Task<bool> MoveNextAsync()` and `public async Task<bool> MoveNextAsync(CancellationToken cancellationToken)` exist at all. Should be done in `ValueTask<bool> IAsyncEnumerator<WorkflowStreamItem>.MoveNextAsync()` instead of forward to those others.

Probably should just adopt the typical `async IAsyncEnumerable<T>` method pattern and not implement any of this ourselves.

jmaeagle99, line 27: Let's think about adopting a dependency on `Microsoft.Bcl.AsyncInterfaces` for `net462` and `netstandard2.0` instead of conditionally carving this out.

src/Temporalio.Extensions.WorkflowStreams/Temporalio.Extensions.WorkflowStreams.csproj
```diff
@@ -0,0 +1,34 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <Description>Temporal SDK .NET Workflow Streams Extension (Experimental)</Description>
+    <IncludeSymbols>true</IncludeSymbols>
+    <LangVersion>9.0</LangVersion>
+    <PackageReadmeFile>README.md</PackageReadmeFile>
+    <PublishRepositoryUrl>true</PublishRepositoryUrl>
+    <SymbolPackageFormat>snupkg</SymbolPackageFormat>
+    <TargetFrameworks>netcoreapp3.1;netstandard2.0;net462</TargetFrameworks>
```
jmaeagle99, line 10: Any particular reason why `netstandard2.0` itself isn't sufficient?

src/Temporalio.Extensions.WorkflowStreams/README.md
```diff
@@ -0,0 +1,218 @@
+# Workflow Streams
```
jmaeagle99, line 1: This library should clearly call out that it's experimental closer to the top, if not in the header itself.

src/Temporalio.Extensions.WorkflowStreams/SubscribeOptions.cs
```diff
@@ -0,0 +1,32 @@
+using System;
+using System.Collections.Generic;
+
+namespace Temporalio.Extensions.WorkflowStreams
+{
+    /// <summary>
+    /// Options for <see cref="WorkflowStreamClient.Subscribe(SubscribeOptions?)" />.
+    /// </summary>
+    /// <remarks>WARNING: This API is experimental and may change.</remarks>
+    public class SubscribeOptions
```
jmaeagle99, line 10: All of the new options classes should be `ICloneable`.

src/Temporalio.Extensions.WorkflowStreams/WorkflowStreamListener.cs
```diff
@@ -0,0 +1,54 @@
...
+    /// </summary>
+    /// <remarks>
+    /// WARNING: This API is experimental and may change.
+    /// <para>
+    /// Callbacks are serialized (never invoked concurrently) and run on the thread pool, so they
+    /// must not block; to defer further delivery, return a pending task from
+    /// <see cref="OnNextAsync" />.
+    /// </para>
+    /// </remarks>
+    public abstract class WorkflowStreamListener
```
jmaeagle99, line 19: This is probably another transliteration from Java. It implements the `IObservable<T>` pattern without literally implementing it. Modern .NET uses `IAsyncEnumerable<T>` instead. This is largely redundant with the `Subscribe`/`SubscribeAsync` APIs.

src/Temporalio.Extensions.WorkflowStreams/Internal/PayloadWire.cs
```diff
@@ -0,0 +1,60 @@
...
+        /// </exception>
+        public static Payload Decode(string wire)
+        {
+            try
+            {
+                return Payload.Parser.ParseFrom(Convert.FromBase64String(wire));
+            }
+            catch (Exception e) when (e is FormatException || e is InvalidProtocolBufferException)
+            {
+                throw new ArgumentException("workflowstreams: unmarshal payload", e);
```
jmaeagle99, line 37: Exceptions should be full sentences in idiomatic .NET; this looks like a Go transliteration.

src/Temporalio.Extensions.WorkflowStreams/WorkflowStreamItem.cs
```diff
@@ -0,0 +1,42 @@
...
+namespace Temporalio.Extensions.WorkflowStreams
+{
+    /// <summary>
+    /// A single decoded item yielded by a subscription. <see cref="Payload" /> is the raw
+    /// <see cref="Api.Common.V1.Payload" />; decode it at the call site with a payload converter,
+    /// e.g. <c>converter.ToValue&lt;string&gt;(item.Payload)</c> (see
+    /// <see cref="Converters.ConverterExtensions.ToValue{T}(Converters.IPayloadConverter, Payload)" />).
+    /// </summary>
+    /// <remarks>WARNING: This API is experimental and may change.</remarks>
+    public sealed class WorkflowStreamItem
```
jmaeagle99, line 12: nit: just make this a record: `public sealed record WorkflowStreamItem(string Topic, Payload Payload, long Offset)`

jmaeagle99 (changes requested):
I've mostly review .NET idioms rather than correctness of the library.

Also, from the PR description:

> Dependency injection extensions and configuration helpers.

There is no such support that I saw.

brianstrauch:
While this PR is going through review, can you attribute your first commit to your GitHub account and sign the CLA? @gabrielschmith

brianstrauch:
Superseded by the rewrite in #884.

PR #812, opened 2026-08-04
Updating Temporal API protos

Evanthx:
Updating SDK core to get functionality needed for a requested task (query backlinks)


JoshuaFrenchwood (approved):

CHANGELOG.md
```diff
@@ -47,6 +47,7 @@ to docs, or any other relevant information.
 
 ### Changed
 
+- Updated Core to `65b25ada` (`temporal-core` 0.6.0)
```
jmaeagle99: sdk-core submodule updates typically don't rise to the level of needing a release note entry since it is not meaningful for customers. I think this should be removed. Will probably need to diff the sdk-rust changes between the last release and the pending release to figure out what really should be put into release notes.

Evanthx: I got a linter error without updating the release note entry - otherwise I agree with you!

Evanthx: I pulled the change out to double check -

Run if echo "$LABELS" | grep -q '"skip-changelog"'; then
Error: This PR does not update a CHANGELOG.md. If it includes a user-facing change, add an entry under [Unreleased] in CHANGELOG.md. If no changelog entry is needed, a maintainer can apply the 'skip-changelog' label.
Error: Process completed with exit code 1.

jmaeagle99: The error states what the problem was. The 'skip-changelog' label wasn't applied to the PR. It is now. Please remove the change log entry.

jmaeagle99 (approved):

CHANGELOG.md
jmaeagle99: I've made the change to remove the changelog and will merge once checks pass.

PR #822, opened 2026-08-11
Test Cloud with ephemeral namespaces

THardy98:
## What was changed

**TLDR**: port of https://github.com/temporalio/sdk-python/pull/1731

Cloud CI now generates a short-lived certificate authority and client certificate, provisions an
ephemeral Temporal Cloud namespace that trusts that authority, and runs the existing mTLS workflow
smoke test through the envconfig-backed test harness. Namespace deletion runs even if provisioning
or the smoke test fails.

The namespace lifecycle command is hosted in the existing test executable and waits for Cloud
Operations API requests to reach a terminal state. The focused API-key and Cloud Operations client
tests remain unchanged.

## Why?

Enables isolation for parallel cloud CI runs

## Checklist

1. Closes N/A

2. How was this tested:

   - `dotnet format --verify-no-changes --no-restore`
   - `actionlint .github/workflows/ci.yml`
   - Certificate-chain verification
   - Ephemeral namespace creation, envconfig mTLS smoke execution, and deletion in GitHub Actions

3. Any docs updates needed?

   No. This is an internal test/CI change, so no changelog entry is needed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -87,15 +87,52 @@ jobs:
       - name: Confirm bench works
         run: dotnet run --project tests/Temporalio.SimpleBench/Temporalio.SimpleBench.csproj -- --workflow-count 5 --max-cached-workflows 100 --max-concurrent 100
 
+      - name: Generate Cloud test certificates
+        # Only supported in non-fork, non-Dependabot runs since secrets are not available in those contexts.
+        if: ${{ matrix.cloudTestTarget && github.actor != 'dependabot[bot]' && (github.event.pull_request.head.repo.full_name == '' || github.event.pull_request.head.repo.full_name == 'temporalio/sdk-dotnet') }}
+        shell: bash
+        run: |
+          cert_dir="$RUNNER_TEMP/cloud-test-certs"
+          mkdir "$cert_dir"
+          openssl req -x509 -newkey rsa:2048 -nodes -days 1 \
+            -keyout "$cert_dir/ca.key" -out "$cert_dir/ca.pem" \
+            -subj '/CN=Temporal .NET SDK Cloud CI CA'
+          openssl req -newkey rsa:2048 -nodes \
+            -keyout "$cert_dir/client.key" -out "$cert_dir/client.csr" \
+            -subj '/CN=Temporal .NET SDK Cloud CI'
+          openssl x509 -req -days 1 -in "$cert_dir/client.csr" \
+            -CA "$cert_dir/ca.pem" -CAkey "$cert_dir/ca.key" -CAcreateserial \
+            -out "$cert_dir/client.pem" -extfile <(printf 'extendedKeyUsage=clientAuth')
+          {
+            echo "TEMPORAL_CLOUD_CLIENT_CA_PATH=$cert_dir/ca.pem"
+            echo "TEMPORAL_TLS_CLIENT_CERT_PATH=$cert_dir/client.pem"
+            echo "TEMPORAL_TLS_CLIENT_KEY_PATH=$cert_dir/client.key"
+          } >> "$GITHUB_ENV"
+
+      - name: Create Cloud namespace
+        id: create-cloud-namespace
+        # Only supported in non-fork, non-Dependabot runs since secrets are not available in those contexts.
+        if: ${{ matrix.cloudTestTarget && github.actor != 'dependabot[bot]' && (github.event.pull_request.head.repo.full_name == '' || github.event.pull_request.head.repo.full_name == 'temporalio/sdk-dotnet') }}
+        env:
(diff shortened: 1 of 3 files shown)


tests/Temporalio.Tests/CloudNamespaceCommand.cs
```diff
@@ -0,0 +1,141 @@
...
+
+            var remaining = OperationTimeout - elapsed.Elapsed;
+            if (remaining <= TimeSpan.Zero)
+            {
+                throw new TimeoutException($"Timed out waiting for Cloud operation {operationId}");
+            }
+
+            // Honor the server hint so the control plane can throttle polling when needed.
+            var delay = operation.CheckDuration?.ToTimeSpan() ?? TimeSpan.Zero;
+            if (delay < TimeSpan.FromSeconds(1))
```
jmaeagle99, line 134: How likely is it that the service response doesn't have a check duration field? Is a fallback of 1 second too aggressive?

.github/workflows/ci.yml
```diff
@@ -87,16 +87,53 @@ jobs:
...
+            echo "TEMPORAL_TLS_CLIENT_KEY_PATH=$cert_dir/client.key"
+          } >> "$GITHUB_ENV"
+
+      - name: Create Cloud namespace
+        id: create-cloud-namespace
+        # Only supported in non-fork, non-Dependabot runs since secrets are not available in those contexts.
+        if: ${{ matrix.cloudTestTarget && github.actor != 'dependabot[bot]' && (github.event.pull_request.head.repo.full_name == '' || github.event.pull_request.head.repo.full_name == 'temporalio/sdk-dotnet') }}
+        env:
+          TEMPORAL_CLIENT_CLOUD_API_KEY: ${{ secrets.TEMPORAL_CLIENT_CLOUD_API_KEY }}
+          TEMPORAL_CLIENT_CLOUD_API_VERSION: v0.19.1
```
jmaeagle99: Can we make this a job level env so it's not duplicated everywhere?

jmaeagle99: This version disagrees with the one that comes later in this file under the "Test cloud operations client" step.

.github/workflows/ci.yml
```diff
@@ -87,16 +87,53 @@ jobs:
       - name: Confirm bench works
         run: dotnet run --project tests/Temporalio.SimpleBench/Temporalio.SimpleBench.csproj -- --workflow-count 5 --max-cached-workflows 100 --max-concurrent 100
 
+      - name: Generate Cloud test certificates
+        # Only supported in non-fork, non-Dependabot runs since secrets are not available in those contexts.
+        if: ${{ matrix.cloudTestTarget && github.actor != 'dependabot[bot]' && (github.event.pull_request.head.repo.full_name == '' || github.event.pull_request.head.repo.full_name == 'temporalio/sdk-dotnet') }}
```
jmaeagle99, line 94: This condition is now replicated many times. I'm wondering if this should be evaluated once in the first step where it's needed and just output one value that the remaining steps then check. That way we have less potential drift if the condition needs to change.

.github/workflows/ci.yml
```diff
@@ -87,16 +87,53 @@ jobs:
...
-          TEMPORAL_TEST_CLIENT_TARGET_HOST: ${{ vars.TEMPORAL_CLIENT_NAMESPACE }}.tmprl.cloud:7233
-          TEMPORAL_TEST_CLIENT_NAMESPACE: ${{ vars.TEMPORAL_CLIENT_NAMESPACE }}
-          TEMPORAL_TEST_CLIENT_CERT: ${{ secrets.TEMPORAL_CLIENT_CERT }}
-          TEMPORAL_TEST_CLIENT_KEY: ${{ secrets.TEMPORAL_CLIENT_KEY }}
+          TEMPORAL_TEST_ENV_CONFIG_SERVER: true
+          TEMPORAL_ADDRESS: ${{ steps.create-cloud-namespace.outputs.namespace }}.tmprl.cloud:7233
+          TEMPORAL_NAMESPACE: ${{ steps.create-cloud-namespace.outputs.namespace }}
         run: dotnet run --project tests/Temporalio.Tests -- -verbose -method "*.ExecuteWorkflowAsync_Simple_Succeeds"
 
+      - name: Delete Cloud namespace
```
jmaeagle99, line 134: I think we need to ignore failures from this step so that the job doesn't fail just because we couldn't tear down the namespace.

tests/Temporalio.Tests/CloudNamespaceCommand.cs
THardy98: not certain, but I think unlikely. I've used this API as part of the load testing stuff and it's always been present from what I can tell.

In any case, I don't think a fallback is harmful, I'll bump it to 10s.

jmaeagle99 (approved):

PR #823, opened 2026-08-11
Classify and separate .NET tests for Cloud execution

THardy98:
## What was changed

This change makes Cloud test eligibility part of xUnit discovery. Tests are eligible by default and
can be excluded at the class or method level with a typed `CloudTestExclusion` trait that records
whether they require a local server, depend on functionality unavailable in Cloud CI, or still need
Cloud adaptation.

The ephemeral mTLS lane introduced by #822 now runs the eligible suite with a standard `dotnet test`
filter and uploads TRX results after cleanup. Tests that create custom-runtime clients clone the
active connection options so envconfig-provided TLS and authentication are retained. The README
documents commands for executing and inventorying both eligible and excluded tests.

Added:
```
dotnet test tests/Temporalio.Tests --list-tests --filter "CloudTestExclusionReason=NeedsCloudAdaptation"
```
to inventory tests that have been skipped against cloud (can filter by any cloud exclusion reason).

## Why?

Running one hand-picked smoke test cannot show which parts of the SDK suite work against Cloud.
Runner-native exclusion metadata makes that segregation automated, queryable, and reviewable
without maintaining a separate file whose entries can drift from xUnit discovery.

The ephemeral namespace lifecycle required by this lane was added in #822 and is now available on
`main`. This PR therefore contains only the suite-segregation and Cloud-execution changes.

## Checklist

1. Closes N/A

2. How was this tested:

   - `dotnet format --verify-no-changes --no-restore`
   - `dotnet test --no-restore` (683 passed, 13 skipped)
   - Targeted exclusion metadata tests (3 passed)
   - xUnit inventory filters for all three exclusion reasons
   - `actionlint .github/workflows/ci.yml`
   - Ephemeral mTLS Cloud suite in GitHub Actions (pending)

3. Any docs updates needed?

   The README documents Cloud eligibility, execution, inventory, and exclusion reasons. This is an
   internal test/CI change, so no changelog entry is needed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,6 @@ jobs:
   build-lint-test:
     env:
       RestoreLockedMode: true
-      TEMPORAL_CLIENT_CLOUD_API_VERSION: v0.19.1
     strategy:
       fail-fast: false
       matrix:
@@ -23,7 +22,6 @@ jobs:
             checkTarget: true
           - os: ubuntu-latest
             docsTarget: true
-            cloudTestTarget: true
           - os: ubuntu-arm
             runsOn: ubuntu-24.04-arm64-2-core
           - os: macos-arm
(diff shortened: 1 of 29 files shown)


tests/Temporalio.Tests/CloudTestExclusionAttribute.cs
```diff
@@ -0,0 +1,14 @@
+namespace Temporalio.Tests;
+
+using Xunit.Sdk;
+
+[AttributeUsage(AttributeTargets.Class | AttributeTargets.Method, AllowMultiple = false)]
+[TraitDiscoverer(
+    "Temporalio.Tests.CloudTestExclusionDiscoverer",
+    "Temporalio.Tests")]
+public sealed class CloudTestExclusionAttribute : Attribute, ITraitAttribute
```
jmaeagle99, line 9: I would consider adding a `string note`/`string comment` constructor parameter to force anyone who is applying to attribute to describe why the test is being excluded.

tests/Temporalio.Tests/Worker/NexusWorkerTests.cs
```diff
@@ -95,6 +95,7 @@ public Task CancelAsync(OperationCancelContext context) =>
     }
 
     [Fact]
+    [CloudTestExclusion(CloudTestExclusionReason.NeedsCloudAdaptation)]
```
jmaeagle99: Nearly all of the tests in this class have been excluded except for the `AddNexusService_*` tests and the `ExecuteNexusOperationAsync_BadService_FailsRegistration` test. These tests don't actually connect to any server. So I'm thinking if these tests should be moved to a new `NexusWorkerRegistrationTests` class and then the entire `NexusWorkerTests` class is excluded instead of every individual test.

tests/Temporalio.Tests/Extensions/Hosting/NexusWorkerServiceTests.cs
```diff
@@ -83,6 +83,7 @@ public Task CancelAsync(OperationCancelContext context) =>
     }
 
     [Fact]
+    [CloudTestExclusion(CloudTestExclusionReason.NeedsCloudAdaptation)]
```
jmaeagle99: Similar separation as described in https://github.com/temporalio/sdk-dotnet/pull/823/changes#r3805858081

tests/Temporalio.Tests/Common/EnvConfig/ClientConfigTests.cs
```diff
@@ -58,6 +60,7 @@ public ClientConfigTests(ITestOutputHelper output)
 
         // === PROFILE LOADING TESTS (7 tests) ===
         [Fact]
+        [CloudTestExclusion(CloudTestExclusionReason.NeedsCloudAdaptation)]
```
jmaeagle99: It's possible that all of these annotated tests in this test file can be fixed by setting `OverrideEnvVars = new()`, then the annotations and comment can be undone.

tests/Temporalio.Tests/CloudTestExclusionReason.cs
```diff
@@ -0,0 +1,8 @@
+namespace Temporalio.Tests;
+
+public enum CloudTestExclusionReason
```
jmaeagle99, line 6: - I think each of the enum fields should have some comment describing what their purpose is. For example, `CloudUnavailable` is a big nebulous because it reads like "the cloud is not available", which isn't something that the tests can statically declare.
- I think `CloudUnavailable` currently means "cloud provisioning not available (yet)". Maybe rename to `RequiresCloudProvisioning` to better convey the classification.

README.md
```diff
@@ -1374,6 +1374,19 @@ Extra args can be added after `--`, e.g. `-- -verbose` would show verbose logs a
...
+Tests are eligible to run against Temporal Cloud unless they have a `CloudTestExclusion` attribute. Run or list the
+Cloud-eligible tests with:
+
+    dotnet test tests/Temporalio.Tests --filter 'CloudTest!=Excluded'
+    dotnet test tests/Temporalio.Tests --list-tests --filter 'CloudTest!=Excluded'
+
+To inventory excluded tests, filter on `CloudTest=Excluded`, or filter on `CloudTestExclusionReason` to inspect a
+particular category:
+
+* `RequiresLocalServer` for tests of local server behavior
```
jmaeagle99: I would move these explanations to the `CloudTestExclusionReason` enum and not duplicate the documentation here. Developers can go look at the enum if they need to know what the values are and their definitions.

tests/Temporalio.Tests/CloudTestExclusionDiscoverer.cs
```diff
@@ -0,0 +1,23 @@
...
+
+using Xunit.Abstractions;
+using Xunit.Sdk;
+
+public sealed class CloudTestExclusionDiscoverer : ITraitDiscoverer
+{
+    public IEnumerable<KeyValuePair<string, string>> GetTraits(IAttributeInfo traitAttribute)
+    {
+        var reason = (CloudTestExclusionReason)traitAttribute.GetConstructorArguments().Single();
+        var reasonName = reason switch
```
jmaeagle99: Use something like the following instead so there is no drift if a new enum field is added:

```
if (!Enum.IsDefined(typeof(CloudTestExclusionReason), reason))
{
    throw new InvalidOperationException($"Unknown exclusion reason: {reason}");
}
yield return new("CloudTest", "Excluded");
yield return new("CloudTestExclusionReason", reason.ToString());
```

.github/workflows/ci.yml
```diff
@@ -129,7 +129,13 @@ jobs:
           TEMPORAL_TEST_ENV_CONFIG_SERVER: true
           TEMPORAL_ADDRESS: ${{ steps.create-cloud-namespace.outputs.namespace }}.tmprl.cloud:7233
           TEMPORAL_NAMESPACE: ${{ steps.create-cloud-namespace.outputs.namespace }}
-        run: dotnet run --project tests/Temporalio.Tests -- -verbose -method "*.ExecuteWorkflowAsync_Simple_Succeeds"
+        run: |
+          dotnet test tests/Temporalio.Tests/Temporalio.Tests.csproj --no-build \
+            --filter "CloudTest!=Excluded" \
+            --logger "console;verbosity=detailed" \
+            --logger "trx;LogFilePrefix=cloud-mtls-results" \
+            --results-directory tests/Temporalio.Tests/CloudTestResults \
```
jmaeagle99, line 137: Add this directory to the .gitignore file

.github/workflows/ci.yml
```diff
@@ -129,7 +129,13 @@ jobs:
           TEMPORAL_TEST_ENV_CONFIG_SERVER: true
           TEMPORAL_ADDRESS: ${{ steps.create-cloud-namespace.outputs.namespace }}.tmprl.cloud:7233
           TEMPORAL_NAMESPACE: ${{ steps.create-cloud-namespace.outputs.namespace }}
-        run: dotnet run --project tests/Temporalio.Tests -- -verbose -method "*.ExecuteWorkflowAsync_Simple_Succeeds"
+        run: |
```
jmaeagle99, line 132: We should consider moving this to a separate job. This is going to rerun all of the tests again cloud, which is likely to be longer than running against the local server and will add more time to the current job. This is a possible source of job timeout. This can also cause later actions to fail to run (e.g. doc deployment).

README.md
```diff
@@ -1374,6 +1374,19 @@ Extra args can be added after `--`, e.g. `-- -verbose` would show verbose logs a
 options. If the arguments are anything but `--help`, the current assembly is prepended to the args before sending to the
 xUnit runner.
 
+Tests are eligible to run against Temporal Cloud unless they have a `CloudTestExclusion` attribute. Run or list the
+Cloud-eligible tests with:
+
+    dotnet test tests/Temporalio.Tests --filter 'CloudTest!=Excluded'
```
jmaeagle99: nit: Let's use double-quotes so it that it works in more environments.

jmaeagle99 (commented):
Some of the test failures need exclusion as well:
- `ListActivitiesAsync_SimpleList_IsAccurate` is probably NeedsCloudAdaptation
- `ExecuteWorkflowAsync_LastResult_ProperlyPresent` just needs to be fixed to not assume there are no existing schedules, so maybe NeedsCloudAdaptation? Something like #842 should fix this.

jmaeagle99 (commented):
- `ListActivitiesAsync_SimpleList_IsAccurate` is failing because it has the same issue as the `ListWorkflowsAsync_ManualPaging_IsAccurate` test; it's not that visibility is eventually consistent, it's that Cloud's Elastic Search is returning a next page token for the what we assume is the last page (that token will produce an empty page with an empty token) whereas the SQLite backed Temporal CLI server doesn't do this. I would exclude this for now and determine how to fix separately.
- `ExecuteWorkflowAsync_SignalWithStartFromWorkflow_Succeeds` is hanging because it relies on `history.enableSignalWithStartFromWorkflow=true` dynamic config, which may not be enabled in Cloud. Exclude and classify as `RequiresLocalServer`.

tests/Temporalio.Tests/Worker/WorkflowWorkerTests.cs
```diff
@@ -843,6 +846,9 @@ public enum Scenario
     }
 
     [Fact]
+    [CloudTestExclusion(
+        CloudTestExclusionReason.NeedsCloudAdaptation,
+        "Activity cancellation can complete after worker shutdown begins, exposing a bridge finalization race.")]
```
jmaeagle99, line 851: Can you log an issue for this and `ExecuteWorkflowAsync_StdlibSemaphore_NonAsyncDeadlocks`?

jmaeagle99 (approved):

tests/Temporalio.Tests/Worker/WorkflowWorkerTests.cs
THardy98: Tracked both tests in https://github.com/temporalio/sdk-dotnet/issues/856.

THardy98: https://github.com/temporalio/sdk-dotnet/issues/857
general tracking issue for tests that need cloud adaptation

PR #825, opened 2026-08-12
Adding SDK Ergonomics Query link

Evanthx:
Queries now have a backlink.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,14 @@ to docs, or any other relevant information.
   Temporal Server 1.32.0 or later.
 - 
 ### Changed                 — changes in existing functionality
+- Support Workflow Queries as Nexus operations. A query issued from inside a Nexus operation handler
+  now propagates the link the server returns for the workflow that processed it, so the caller's
+  Nexus operation event points back at the queried workflow. Requires a server that populates
+  `QueryWorkflowResponse.link` and understands the `Link.Workflow` variant, added in server
+  1.32.0-162.0 (temporalio/temporal#11274); older servers leave it unset and nothing is propagated.
+
+### Changed
+
 - A `common.v1.Link.Workflow` now serializes to the workflow path
   `temporal:///namespaces/{ns}/workflows/{wid}/{rid}` with the optional `reason` as a query param,
   rather than reusing the workflow-event path with a `/history` suffix and dropping `reason`. The
@@ -52,9 +60,7 @@ to docs, or any other relevant information.
   match the other SDKs. Inbound workflow links are now parsed as well, and a link with a trailing
   path segment is rejected.
 - A Nexus operation backed by a workflow query now fails when the query fails or is rejected, rather
-  than being retried until the operation times out. `WorkflowQueryFailedException` and
-  `WorkflowQueryRejectedException` map to a non-retryable `BadRequest` handler error, since neither
-  outcome can change on a retry.
+  than being retried until the operation times out.
 - A non-retryable `ApplicationFailureException` with error type `PayloadValidationError` thrown by a
   payload codec or payload converter while decoding Nexus operation input is now reported as a
   non-retryable `BadRequest` handler exception (with the application failure as its cause) instead of
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@ to docs, or any other relevant information.
 # Changelog
 
 ## [Unreleased]
+
+### Added                   — new features
 - Support Workflow Queries as Nexus operations. A query issued from inside a Nexus operation handler
   now propagates the link the server returns for the workflow that processed it, so the caller's
   Nexus operation event points back at the queried workflow. Requires a server that populates
@@ -41,6 +43,17 @@ to docs, or any other relevant information.
 - Workflow task completions larger than the gRPC request size limit are now paginated
   automatically when the namespace supports it. Paginated workflow task completions require
   Temporal Server 1.32.0 or later.
+### Changed                 — changes in existing functionality
+
+### Deprecated              — soon-to-be-removed features
+
+### :boom: Breaking Changes — removed or backwards-incompatible features
+
+### Fixed                   — notable bug fixes
+
+### Security                — notable security fixes
+
+
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,9 +54,6 @@ to docs, or any other relevant information.
   than being retried until the operation times out. `WorkflowQueryFailedException` and
   `WorkflowQueryRejectedException` map to a non-retryable `BadRequest` handler error, since neither
   outcome can change on a retry.
-
-### Changed
-
 - A non-retryable `ApplicationFailureException` with error type `PayloadValidationError` thrown by a
   payload codec or payload converter while decoding Nexus operation input is now reported as a
   non-retryable `BadRequest` handler exception (with the application failure as its cause) instead of
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,16 +55,6 @@ to docs, or any other relevant information.
   `WorkflowQueryRejectedException` map to a non-retryable `BadRequest` handler error, since neither
   outcome can change on a retry.
 
-### Deprecated              — soon-to-be-removed features
-
-### :boom: Breaking Changes — removed or backwards-incompatible features
-
-### Fixed                   — notable bug fixes
-
-### Security                — notable security fixes
-
-
-
 ### Changed
 
 - A non-retryable `ApplicationFailureException` with error type `PayloadValidationError` thrown by a
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,10 @@ to docs, or any other relevant information.
 # Changelog
 
 ## [Unreleased]
+- Support Workflow Queries as Nexus operations. A query issued from inside a Nexus operation handler
+  now propagates the link the server returns for the workflow that processed it, so the caller's
+  Nexus operation event points back at the queried workflow. Requires a server that populates
+  `QueryWorkflowResponse.link`; older servers leave it unset and nothing is propagated.
 
 ### :boom: Breaking Changes
 
@@ -131,10 +135,6 @@ to docs, or any other relevant information.
   pollers are left unchanged.
 - Workers now log a [TMPRL1104] warning when a workflow task takes longer than 5 seconds. Set
   `TEMPORAL_WORKFLOW_TASK_DURATION_WARN_SECONDS` to change the threshold.
-- Support Workflow Queries as Nexus operations. A query issued from inside a Nexus operation handler
-  now propagates the link the server returns for the workflow that processed it, so the caller's
-  Nexus operation event points back at the queried workflow. Requires a server that populates
-  `QueryWorkflowResponse.link`; older servers leave it unset and nothing is propagated.
 
 ### Changed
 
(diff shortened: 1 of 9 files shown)


src/Temporalio/Worker/NexusWorker.cs
```diff
@@ -387,6 +387,19 @@ private HandlerException ConvertToHandlerException(Exception exc)
             {
                 return new(HandlerErrorType.BadRequest, "Workflow failed", exc);
             }
+            else if (exc is WorkflowQueryFailedException)
+            {
+                // A query handler that threw will throw again on every attempt, so this must be a
+                // non-retryable type. Falling through to the Internal default below would make the
+                // server retry the operation until it times out instead of failing it.
+                return new(HandlerErrorType.BadRequest, "Workflow query failed", exc);
```
jmaeagle99: This isn't a BadRequest condition, though. The request was well formed and accepted, it's the handler that faulted. So should this return a different error type? Maybe there isn't a direct mapping to something, but possibly use Internal but NonRetryable?

src/Temporalio/Nexus/ProtoLinkExtensions.cs
```diff
@@ -271,9 +299,23 @@ public static Api.Common.V1.Link.Types.WorkflowEvent ToWorkflowEvent(this NexusL
...
-        // Validate a Temporal-shaped link URI and return its path segments. Expected path shape
-        // is /namespaces/{namespace}/{kind}/{id}/{run}/{tail}.
-        private static string[] ParseTemporalLinkPath(NexusLink link, string expectedKind, string expectedTail)
+        // Simple query param parser because .NET stdlib doesn't have one in all versions.
+        private static Dictionary<string, string> ParseQueryParams(Uri uri) =>
+            uri.Query.
+                TrimStart('?').
+                Split(QuerySeparator, StringSplitOptions.RemoveEmptyEntries).
+                Select(v => v.Split(QueryValueSeparator, 2)).
+                ToDictionary(
```
jmaeagle99, line 322: I had to look this up, but maybe leave a comment stating that the dictionary using ordinal case comparison by default, so that a lookup of `"reason"` and `"Reason"` can yield different result. Possibly worth a XML doc comment on the `ParseQueryParams` function to describe the return value behavior.

tests/Temporalio.Tests/Nexus/ProtoLinkExtensionsTests.cs
```diff
@@ -224,4 +224,305 @@ public void ToActivity_RejectsBadPath()
             Api.Common.V1.Link.Types.Activity.Descriptor.FullName);
         Assert.Throws<ArgumentException>(() => link.ToActivity());
     }
+
+    [Fact]
+    public void WorkflowLink_ToNexusLink_BuildsExpectedUri()
```
jmaeagle99, line 229: This is pretty much the same as `WorkflowLink_ToNexusLink_BuildsWorkflowUri`. I would consolidate.

tests/Temporalio.Tests/Nexus/ProtoLinkExtensionsTests.cs
```diff
@@ -224,4 +224,305 @@ public void ToActivity_RejectsBadPath()
...
+            WorkflowId = "wf",
+            RunId = "run-id",
+            Reason = "Query processed",
+        };
+        var protoLink = workflow.ToNexusLink().ToProtoLink();
+        Assert.Equal(workflow, protoLink.Workflow);
+    }
+
+    [Fact]
+    public void ProtoToNexusLink_WorkflowVariant_Dispatches()
```
jmaeagle99, line 456: Nearly the same as `CommonLink_ToNexusLink_FallsBackToWorkflow`. I would consolidate.

tests/Temporalio.Tests/Nexus/ProtoLinkExtensionsTests.cs
```diff
@@ -224,4 +224,305 @@ public void ToActivity_RejectsBadPath()
...
+                "&eventType=WorkflowExecutionStarted"),
+            Api.Common.V1.Link.Types.WorkflowEvent.Descriptor.FullName);
+
+        Assert.Equal("a b", link.ToWorkflowEvent().RequestIdRef.RequestId);
+    }
+
+    [Fact]
+    public void ToWorkflow_AcceptsEmptyRunId()
+    {
+        // Characterization, not a statement of intent. A trailing slash still yields the expected
```
jmaeagle99: Not sure I understand what "Characterization, not a statement of intent." is meant to describe. Passing whatever is in the test is "a statement of intent".

jmaeagle99 (commented):
The PR description says "Queries now have a backlink.", but that's just one of three changes in this PR. There's also the URL format change and the error mapping change. Not sure they really are tied to the support for query backlinks. Might have made it faster to review if they were separate (if they truly are disconnected) so I didn't have to figure out that delineation. This will also make it harder to look back into history to discover which change was responsible for one of these three aspects.

src/Temporalio/Worker/NexusWorker.cs
Evanthx: Good catch! Changed it as suggested.

src/Temporalio/Nexus/ProtoLinkExtensions.cs
Evanthx: Added XML doc as suggested

tests/Temporalio.Tests/Nexus/ProtoLinkExtensionsTests.cs
Evanthx: Valid, improved wording

tests/Temporalio.Tests/Nexus/ProtoLinkExtensionsTests.cs
Evanthx: Found four link conversion tests in there that were pretty much the same. I consolidated them but made the new location ProtoLinkExtensionsTests as that kept them in a test suite for links instead of a more generally named one.

tests/Temporalio.Tests/Nexus/ProtoLinkExtensionsTests.cs
Evanthx: As above - found four link conversion tests in there that were pretty much the same. I consolidated them but made the new location ProtoLinkExtensionsTests as that kept them in a test suite for links instead of a more generally named one.

src/Temporalio/Worker/NexusWorker.cs
Evanthx: Follow up question - when reviewing this, right above it is another BadRequest return -
            if (exc is WorkflowFailedException)
            {
                return new(HandlerErrorType.BadRequest, "Workflow failed", exc);
            }

Should this be changed as well? Going off your initial comment I'm not sure it should be done in this PR if so, but ... well, wanted to ask.

jmaeagle99: I agree that should not be BadRequest either. I'm also talking from a conceptual perspective, using HTTP 400 vs HTTP 500 as an example. But there doesn't seem to be a separate HandlerErrorType for this situation. So I leave it to the Nexus team to decide what to do for this. Probably want to be consistent across the SDKs though.

jmaeagle99 (dismissed):

CHANGELOG.md
```diff
@@ -77,6 +77,10 @@ to docs, or any other relevant information.
   pollers are left unchanged.
 - Workers now log a [TMPRL1104] warning when a workflow task takes longer than 5 seconds. Set
   `TEMPORAL_WORKFLOW_TASK_DURATION_WARN_SECONDS` to change the threshold.
+- Support Workflow Queries as Nexus operations. A query issued from inside a Nexus operation handler
```
jmaeagle99: Please merge `main` into this branch and move these updates to the unreleased section.

tests/Temporalio.Tests/Nexus/QueryResponseLinkTests.cs
```diff
@@ -0,0 +1,245 @@
...
+    public async Task QueryAsync_OutsideNexusContextIgnoresResponseLink()
+    {
+        // A query issued outside a Nexus operation handler must not touch the operation context at
+        // all. Guards against the propagation being reached without a context, which would throw.
+        var client = NewClient(new QueryWorkflowResponse
+        {
+            Link = WorkflowLink("wf-target", "target-run", "Query processed"),
+            QueryResult = await ToPayloadsAsync("answer"),
+        });
+        var context = NewContext();
```
jmaeagle99, line 77: This isn't used and probably should be remove.

Evanthx: Removed

Evanthx: --Side note: Added an end to end test as we talked about. That test will fail, waiting for the server to get released then I can update that in here and it should work

tests/Temporalio.Tests/Worker/NexusWorkerTests.cs
```diff
@@ -1426,7 +1426,12 @@ public async Task ExecuteNexusOperationAsync_CancelWaitRequested_ProperlyFails()
         var exc2 = Assert.IsType<NexusOperationFailureException>(exc.InnerException);
         var exc3 = Assert.IsType<HandlerException>(exc2.InnerException);
         Assert.Equal(HandlerErrorType.NotImplemented, exc3.ErrorType);
-        Assert.Equal("Intentional failure", exc3.Message);
+        var chainMessages = new List<string>();
+        for (Exception? cause = exc; cause != null; cause = cause.InnerException)
```
jmaeagle99: Looked into this more. The top level exception of exc3 used to be a `HandlerException` with message `"Intentional failure"`. But now it's a `HandlerException` with message `""` with an inner exception of type `HandlerException` with message `"Intentional failure"`. I think this needs to be investigated as to why this is happening because that's a behavior change to an existing feature that I believe is now GA.

Evanthx: Thank you for investigating!

tests/Temporalio.Tests/Worker/NexusQueryOperationTests.cs
```diff
@@ -0,0 +1,279 @@
...
+using Temporalio.Workflows;
+using Xunit;
+using Xunit.Abstractions;
+
+/// <summary>
+/// End-to-end tests for Query-backed Nexus operations. A Query is always synchronous and writes
+/// nothing to history, so the handler simply queries and returns the result; there is no operation
+/// token and no completion callback.
+/// </summary>
+public class NexusQueryOperationTests : WorkflowEnvironmentTestBase
```
jmaeagle99, line 22: This needs the `[CloudTestExclusion]` attribute.

CHANGELOG.md
```diff
@@ -19,20 +19,30 @@ to docs, or any other relevant information.
 
 ## [Unreleased]
 
-### Added
-
+### Added                   — new features
```
jmaeagle99: Please remove the `- new features` and `- changes in existing functionality`. That's explanation for the author, not meant to be in the release notes.

tests/Temporalio.Tests/Worker/NexusWorkerTests.cs
Evanthx: I got pulled off this for a bit and it looks like this issue got addressed in https://github.com/temporalio/sdk-dotnet/pull/799 - so that fix is now merged into this PR

tests/Temporalio.Tests/Worker/NexusQueryOperationTests.cs
Evanthx: Added

jmaeagle99 (approved):

PR #829, opened 2026-08-12
Pass through Nexus handler exceptions in payload serializer

bergundy:
The Nexus payload serializer already let ApplicationFailureException from a payload codec or converter propagate untouched. Do the same for HandlerException so users can control the resulting Nexus error type and retry behavior instead of having it rewrapped as INTERNAL (codec) or BAD_REQUEST (converter).

Matches the behavior of Python and TypeScript SDKs, and to some degree Java and Go with the recent changes in https://github.com/temporalio/sdk-java/pull/3002 and https://github.com/temporalio/sdk-go/pull/2549.


Quinn-With-Two-Ns (approved):

Quinn-With-Two-Ns:
Please flag the changelog as well

PR #830, opened 2026-08-12
Automate NuGet and GitHub releases

eamsden:
## What changed?

- Start release publication automatically after a successful `NuGet Package` run on `main` or a release branch when the SDK version changed.
- Reuse the exact package artifact and source commit from that run instead of rebuilding packages during publication.
- Keep the existing integration-gallery publish and smoke test followed by nuget.org publish and smoke test.
- Wait for all seven published package IDs and verify both each runtime package and symbol package against the originating workflow artifact.
- Create the matching Git tag and GitHub release only after production packages pass their smoke test.
- Build release notes from the curated `CHANGELOG.md` Unreleased section and GitHub-generated contributor/change notes.
- Isolate concurrency by immutable release candidate so one pending version cannot displace another, while making retries accept only matching published artifacts and GitHub state.
- Use GitHub's server-side legacy latest-release selection so concurrent or delayed stable releases cannot leave an older version marked latest.

## Why?

The current release workflow must be dispatched manually and stops after NuGet publication. This makes the version-bump merge the single release trigger and completes both NuGet and GitHub publication from the same immutable package build.

This is intentionally much smaller than the Temporal-backed Java release automation in temporalio/sdk-java#2995. NuGet Trusted Publishing, immutable Actions artifacts, and the existing integration/production smoke gates provide the required sequencing and retry behavior without another orchestration runtime.

## Verification

- Full repository CI matrix, feature tests, security scan, changelog check, and dependency submission are green.
- Actionlint, ShellCheck, Bash syntax validation, and `git diff --check` pass.
- Production and integration symbol CDN endpoints were checked against an existing Temporalio release.
- No workflow, package publication, tag, or GitHub release was dispatched while developing this change.

The release scripts intentionally do not have a mocked test harness or a CI-for-CI job. This is an internal SDK-team release utility; the chosen maintenance bar is readable production scripts, static validation, and the live repository CI/release gates rather than maintaining mocks of GitHub and NuGet behavior.


.github/workflows/release-publish.yml
```diff
@@ -1,105 +1,187 @@
...
+# A merged version change is the release authorization. The workflow publishes
+# NuGet first, verifies the packages, and only then makes the GitHub release
+# public. Rerunning a partially completed release is safe: NuGet pushes skip
+# existing packages and the GitHub job accepts only the exact expected release.
 
 on:
-  workflow_dispatch: {}
+  workflow_run:
+    workflows:
+      - NuGet Package
```
jmaeagle99: The "NuGet Package" workflow automatically runs for each new branch update. I don't think we want to automatically run the release workflow each time that workflow runs.

.github/workflows/release-publish.yml
```diff
@@ -1,105 +1,187 @@
...
-          if [ -z "$version" ]; then
-            echo "Could not read <Version> from Directory.Build.props" >&2
+          set -euo pipefail
+
+          read_version() {
+            sed -n 's|^[[:space:]]*<Version>\([^<]*\)</Version>[[:space:]]*$|\1|p'
+          }
+
+          version=$(read_version < Directory.Build.props)
+          if [[ ! "$version" =~ ^[0-9]+\.[0-9]+\.[0-9]+(-[0-9A-Za-z]+([.-][0-9A-Za-z]+)*)?$ ]]; then
```
jmaeagle99, line 64: This allows for things that look like `<major>.<minor>.<patch>-<label>` but not [build metadata](https://semver.org/#spec-item-10) e.g. `<major>.<minor>.<patch>-<label>+<commit_sha>`. But not entirely sure this is necessary here.

.github/workflows/release-publish.yml
```diff
@@ -1,105 +1,187 @@
...
+          persist-credentials: false
+          ref: ${{ github.event.workflow_run.head_sha }}
 
-      - name: Read version
-        id: read
+      - name: Read release candidate
+        id: candidate
+        env:
+          ARTIFACT_RUN_ID: ${{ github.event.workflow_run.id }}
+          RELEASE_COMMIT: ${{ github.event.workflow_run.head_sha }}
```
jmaeagle99: I think the commit that should be released should be different than the commit that the workflow uses to perform the release. This allows fixing the workflow outside of the code that is actually being released. Typically not a problem when release workflows are in separate repos than the thing being released, but that's not the case here.

.github/workflows/release-publish.yml
```diff
@@ -1,105 +1,187 @@
...
 
-# Automated NuGet publishing and smoke testing.
+# A merged version change is the release authorization. The workflow publishes
+# NuGet first, verifies the packages, and only then makes the GitHub release
+# public. Rerunning a partially completed release is safe: NuGet pushes skip
+# existing packages and the GitHub job accepts only the exact expected release.
 
 on:
-  workflow_dispatch: {}
+  workflow_run:
```
jmaeagle99, line 10: We'll want a way to rerun the release workflow without having to rerun "NuGet Package". Otherwise, a partially successful workflow result represents a poisoned version that can't be fixed without manually uploading everything.

PR #844, opened 2026-08-17
OpenTelemetry common-core refactor

eamsden:
## What was changed

- Added the experimental `Temporalio.Extensions.Gcp.CloudRun.OpenTelemetry` package for Temporal workers running on Google Cloud Run (by @seanbollin).
- Moved lightweight provider-neutral configuration into `Temporalio.Extensions.OpenTelemetry`, including resolved options, Temporal runtime metrics, interceptor insertion, and bounded force-flush/task orchestration.
- Kept the full OpenTelemetry SDK and OTLP exporter dependencies provider-local. AWS Lambda owns the shared tracer-provider factory source, and Cloud Run links that source into its own assembly.
- Kept AWS X-Ray trace IDs and Lambda shutdown-hook behavior in AWS, and Cloud Run environment fallbacks, defaults, and its returned shutdown handle in GCP.
- Consolidated duplicated test support and serialized environment-variable tests across the AWS and GCP test classes.

## Why?

The AWS Lambda and GCP Cloud Run extensions had near-identical OpenTelemetry setup and shutdown implementations. Sharing the provider-neutral behavior reduces maintenance and drift without making the full OpenTelemetry SDK/exporter dependency graph transitive for users of the lightweight `Temporalio.Extensions.OpenTelemetry` package.

## Checklist

1. How was this tested:
   - 19 focused AWS Lambda, Cloud Run, and shared OpenTelemetry configuration tests passed.
   - The shared package built for `netstandard2.0` and `net462`; both provider packages and the test assembly built with 0 warnings and 0 errors.
   - Formatting verification passed for all four affected projects.
   - Debug packing passed for the shared, AWS, and GCP OpenTelemetry packages. The generated shared-package manifest contains only `Temporalio`, `OpenTelemetry.Api`, and `System.Diagnostics.DiagnosticSource`; the full SDK and OTLP exporter are dependencies only of the provider packages.
2. Any docs updates needed?
   - The Cloud Run package README includes shutdown guidance, the public GCP assembly is included in generated API documentation, and the changelog entry mentions the shared configuration layer.


src/Temporalio.Extensions.OpenTelemetry/OpenTelemetryConfiguration.cs
```diff
@@ -0,0 +1,147 @@
...
+            {
+                await flushTask.WaitAsync(cancellationToken).ConfigureAwait(false);
+            }
+            catch (OperationCanceledException) when (cancellationToken.IsCancellationRequested)
+            {
+                flushTask.Forget();
+            }
+        }
+
+        private static string FirstNonEmpty(IEnumerable<string?> values) =>
```
jmaeagle99: I didn't think about this in the Lambda review, but this has a hidden problem in that if the caller forgets to provide a fixed default value for the last value, then it will throw an exception. I think that we should make a couple of changes here to avoid that:
- Refactor to `static string? FirstNonWhitespaceOrDefault(IEnumerable<string?> values) =>`
- Change `.First(` to `.FirstOrDefault(`
- Change `IsNullOrEmpty` to `IsNullOrWhitespace`

Then callers are forced to decide how to handle not resolving a value. They can use null-coalescing with the call to fallback to a default e.g. `FirstNonWhitespaceOrDefault(...) ?? "default-value"`.

Secondly, this fixes a latent bug where if one of the provided values is just whitespace (e.g. `" "`), then it would return that. You probably don't want a whitespace value. This change ignores those.

src/Temporalio.Extensions.Aws.Lambda.OpenTelemetry/TemporalLambdaWorkerOptionsExtensions.cs
```diff
@@ -22,12 +15,7 @@ public static class TemporalLambdaWorkerOptionsExtensions
     {
         private const string DefaultCollectorEndpoint = "http://localhost:4317";
```
jmaeagle99: Should this be "hoisted" to a common location or do you think we should define this per-platform like it currently is? I _may_ just happen to have the same value for both platforms. Or if this value is just the standard OTEL fallback value, then it should probably be hoisted.

src/Temporalio.Extensions.OpenTelemetry/OpenTelemetryConfiguration.cs
```diff
@@ -0,0 +1,147 @@
...
+
+        /// <summary>
+        /// Force-flushes a tracer provider without blocking the caller's thread.
+        /// </summary>
+        /// <param name="forceFlush">Function that force-flushes a provider.</param>
+        /// <param name="flushTimeout">Maximum time for the provider flush.</param>
+        /// <param name="cancellationToken">Cancellation token for waiting on the flush.</param>
+        /// <returns>A task for the flush.</returns>
+        internal static async Task ForceFlushAsync(
+            Func<int, bool> forceFlush,
```
jmaeagle99, line 93: nit: might want to document what `int` and `bool` are supposed to be within this function.

tests/Temporalio.Tests/Extensions/Gcp/CloudRun/OpenTelemetry/TemporalClientConnectOptionsExtensionsTests.cs
```diff
@@ -0,0 +1,184 @@
...
+using Temporalio.Client.Interceptors;
+using Temporalio.Extensions.Gcp.CloudRun.OpenTelemetry;
+using Temporalio.Tests.Extensions.OpenTelemetry;
+using Xunit;
+using TemporalOpenTelemetry = Temporalio.Extensions.OpenTelemetry;
+
+[Collection(OpenTelemetryEnvironmentDefinition.Name)]
+public class TemporalClientConnectOptionsExtensionsTests
+{
+    private const string OTelExporterOtlpEndpointEnvironmentVariable =
```
jmaeagle99, line 13: Should this be replaced by `OpenTelemetryConfiguration.OtlpEndpointEnvironmentVariable` instead of embedding in the test? Similarly for `ServiceNameEnvironmentVariable`.

src/Temporalio.Extensions.Aws.Lambda.OpenTelemetry/OpenTelemetryTracerProviderFactory.cs
```diff
@@ -0,0 +1,44 @@
...
+        /// </summary>
+        /// <param name="options">Resolved OpenTelemetry options.</param>
+        /// <param name="configureBuilder">Optional platform-specific builder customization.</param>
+        /// <returns>The created tracer provider.</returns>
+        internal static TracerProvider CreateTracerProvider(
+            ResolvedOpenTelemetryOptions options,
+            Action<TracerProviderBuilder>? configureBuilder = null)
+        {
+            var builder = Sdk.CreateTracerProviderBuilder();
+            configureBuilder?.Invoke(builder);
```
jmaeagle99: Configure callbacks typically have a last-word-wins semantic. Probably should move to just before the `Build` call.

src/Temporalio.Extensions.OpenTelemetry/OpenTelemetryConfiguration.cs
```diff
@@ -0,0 +1,147 @@
...
+        {
+            if (metricsExportInterval <= TimeSpan.Zero)
+            {
+                throw new ArgumentOutOfRangeException(
+                    optionsParameterName,
+                    "MetricsExportInterval must be greater than zero");
+            }
+
+            return new ResolvedOpenTelemetryOptions(
+                new Uri(FirstNonEmpty(collectorEndpointCandidates)),
```
jmaeagle99: I know this is another existing this from the Lambda code... This can either result in an exception being thrown if the chosen candidate is not well formatted OR could silently fail to export due to underspecified URI components.

Could use `Uri.TryCreate` (throw something more meaningful if this returns false) + scheme validation if we require HTTP/S. The URI parsing validation with more meaningful error message might be worth an attempt since it won't be obvious what URI was not parseable.

I don't feel strongly about this, just pointing it out. Maybe log an issue that we can tackle later.

src/Temporalio.Extensions.Gcp.CloudRun.OpenTelemetry/Temporalio.Extensions.Gcp.CloudRun.OpenTelemetry.csproj
```diff
@@ -0,0 +1,45 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
jmaeagle99, line 1: Please add to relevant Temporalio.ApiDoc files for API doc generation.

eamsden: Caught this during my own review, am fixing.

src/Temporalio.Extensions.Aws.Lambda.OpenTelemetry/TemporalLambdaWorkerOptionsExtensions.cs
eamsden: Good catch, yes this is the standard OTel endpoint

src/Temporalio.Extensions.OpenTelemetry/OpenTelemetryConfiguration.cs
eamsden: Updated in 28dbf07: the helper is now nullable , uses  with , and each provider explicitly coalesces to its default. I also added whitespace/null coverage.

src/Temporalio.Extensions.Aws.Lambda.OpenTelemetry/TemporalLambdaWorkerOptionsExtensions.cs
eamsden: Hoisted in 28dbf07 as the shared internal ; both providers use it as their explicit final fallback.

src/Temporalio.Extensions.OpenTelemetry/OpenTelemetryConfiguration.cs
eamsden: Documented in 28dbf07: the  is the timeout in milliseconds and the  reports whether the flush completed within that timeout.

tests/Temporalio.Tests/Extensions/Gcp/CloudRun/OpenTelemetry/TemporalClientConnectOptionsExtensionsTests.cs
eamsden: Updated in 28dbf07 for both the GCP and AWS tests to reference the shared internal environment-variable constants.

src/Temporalio.Extensions.Aws.Lambda.OpenTelemetry/OpenTelemetryTracerProviderFactory.cs
eamsden: Moved in 28dbf07 so the provider callback runs immediately before , preserving last-word-wins behavior.

src/Temporalio.Extensions.OpenTelemetry/OpenTelemetryConfiguration.cs
eamsden: Good point. I left URI parsing behavior unchanged in this refactor because stricter absolute URI and scheme validation could reject endpoint forms the existing Lambda API accepted. I think that validation and its error contract are best handled as a focused follow-up.

src/Temporalio.Extensions.Gcp.CloudRun.OpenTelemetry/Temporalio.Extensions.Gcp.CloudRun.OpenTelemetry.csproj
eamsden: Addressed in 45a9922: the GCP project is included in , its namespace is listed in the API index, and a warnings-as-errors docs build generated the three intended public GCP types without exposing the internal shared implementation types.

src/Temporalio.Extensions.Gcp.CloudRun.OpenTelemetry/README.md
```diff
@@ -39,9 +40,32 @@ using var worker = new TemporalWorker(
...
+{
+    await worker.ExecuteAsync(shutdownToken);
+}
+catch (OperationCanceledException) when (shutdownToken.IsCancellationRequested)
+{
+    // SIGTERM cancellation is the expected shutdown path.
+}
+finally
+{
+    var elapsed = shutdownToken.IsCancellationRequested ?
```
jmaeagle99: nit: I would probably write this as:

```
shutdownStarted.TrySetResult(TimeSpan.Zero);
var elapsed = await shutdownStarted.Task;
```

instead of checking the cancellation token.

jmaeagle99 (approved):

PR #846, opened 2026-08-17
Add System Nexus transfer payload converter

tconley1428:
## Summary

- Add a dedicated System Nexus outer payload converter.
- Apply transfer conversion to the binary-proto outer envelope.
- Keep System Nexus payload conversion limited to the operation envelope, not command metadata.

Generated System Nexus support will own the nested converter context in a follow-up once the
nex-gen transfer-type changes are ingested.

This mirrors the Python System Nexus converter: it owns transfer conversion for the outer envelope
and scopes the user converters while generated System Nexus transfer types serialize nested values.

## Validation

- `dotnet build tests/Temporalio.Tests/Temporalio.Tests.csproj --no-restore`
- `dotnet test tests/Temporalio.Tests/Temporalio.Tests.csproj --no-build --filter FullyQualifiedName~SystemNexusPayloadConverterTests --verbosity normal`
- `dotnet test tests/Temporalio.Tests/Temporalio.Tests.csproj --no-build --filter FullyQualifiedName~SystemNexusTests --verbosity minimal`


src/Temporalio/Nexus/SystemNexusPayloadConverter.cs
```diff
@@ -0,0 +1,50 @@
...
+            this.userPayloadConverter = userPayloadConverter;
+            this.userFailureConverter = userFailureConverter;
+            outerPayloadConverter = TemporalTransferTypePayloadConverter.Wrap(
+                new DefaultPayloadConverter(new BinaryProtoConverter()));
+        }
+
+        /// <inheritdoc />
+        public Temporalio.Api.Common.V1.Payload ToPayload(object? value)
+        {
+            using var context = SystemNexusConverterContext.Push(
```
jmaeagle99:
```suggestion
            using var _ = SystemNexusConverterContext.Push(
```

src/Temporalio/Nexus/SystemNexusPayloadConverter.cs
```diff
@@ -0,0 +1,50 @@
...
+                userPayloadConverter, userFailureConverter);
+            return outerPayloadConverter.ToPayload(value);
+        }
+
+        /// <inheritdoc />
+        public object? ToValue(Temporalio.Api.Common.V1.Payload payload, Type type)
+        {
+            using var context = SystemNexusConverterContext.Push(
```
jmaeagle99:
```suggestion
            using var _ = SystemNexusConverterContext.Push(
```

src/Temporalio/Nexus/SystemNexusPayloadConverter.cs
```diff
@@ -0,0 +1,50 @@
...
+        /// </summary>
+        /// <param name="userPayloadConverter">The application's payload converter.</param>
+        /// <param name="userFailureConverter">The application's failure converter.</param>
+        internal SystemNexusPayloadConverter(
+            IPayloadConverter userPayloadConverter,
+            IFailureConverter userFailureConverter)
+        {
+            this.userPayloadConverter = userPayloadConverter;
+            this.userFailureConverter = userFailureConverter;
+            outerPayloadConverter = TemporalTransferTypePayloadConverter.Wrap(
```
jmaeagle99: Could this be a static readonly field on this class instead of instantiating it every time?

tests/Temporalio.Tests/Nexus/SystemNexusPayloadConverterTests.cs
```diff
@@ -0,0 +1,94 @@
...
+        var value = new SystemNexusRequest("embedded-value");
+
+        var payload = converter.ToPayload(value);
+
+        Assert.Equal("binary/protobuf", payload.Metadata["encoding"].ToStringUtf8());
+        var transferType = ApiWorkflowService.SignalWithStartWorkflowExecutionRequest.Parser.ParseFrom(
+            payload.Data);
+        Assert.Equal("test/string", transferType.Input.Payloads_[0].Metadata["encoding"].ToStringUtf8());
+        Assert.Equal(value, converter.ToValue(payload, typeof(SystemNexusRequest)));
+        Assert.Throws<InvalidOperationException>(() =>
```
jmaeagle99: This should be in a separate test since it is not related to round trip testing.

jmaeagle99: Also test the same condition for the FailureConverter static property.

jmaeagle99: And maybe some other unit tests for checking the push/pop behavior.

These all might warrant their own unit test class for the type.

src/Temporalio/Worker/WorkflowInstance.cs
```diff
@@ -2688,15 +2689,15 @@ public override Task<NexusWorkflowOperationHandle<TResult>> ScheduleNexusOperati
                 }
 
                 // TODO(cretz): Support Nexus serialization context
-                var payloadConverter = instance.payloadConverterNoContext;
+                var payloadConverter = SystemNexusPayloadVisitor.IsSystemNexusEndpoint(
```
jmaeagle99: `payloadConverter` is also used non-envelope conversion, for example, the summary conversion. I would suggest creating a separate local that holds this instance and use it at the necessary locations.

src/Temporalio/Nexus/SystemNexusConverterContext.cs
```diff
@@ -0,0 +1,73 @@
...
+namespace Temporalio.Nexus
+{
+    /// <summary>
+    /// Provides the user configured converters while System Nexus transfer conversion runs.
+    /// </summary>
+    /// <remarks>
+    /// This context is only available while a System Nexus workflow operation converts its input
+    /// or result through a Temporal transfer type converter.
+    /// </remarks>
+    public static class SystemNexusConverterContext
```
jmaeagle99: `internal`?

tests/Temporalio.Tests/Worker/SystemNexusTests.cs
```diff
@@ -0,0 +1,114 @@
...
+using Temporalio.Client;
+using Temporalio.Converters;
+using Temporalio.Tests.Converters;
+using Temporalio.Worker;
+using Temporalio.Worker.Interceptors;
+using Temporalio.Workflows;
+using Xunit;
+using Xunit.Abstractions;
+
+public class SystemNexusTests : WorkflowEnvironmentTestBase
```
jmaeagle99, line 12: All of this looks to be just a move from WorkflowWorkerTests to SystemNexusTests, which is fine. But I don't see any net-new tests that validate the integration into `ScheduleNexusOperationAsync` method.

src/Temporalio/Worker/WorkflowInstance.cs
```diff
@@ -2688,15 +2689,15 @@ public override Task<NexusWorkflowOperationHandle<TResult>> ScheduleNexusOperati
...
+                    input.ClientOptions.Endpoint) ?
+                    new SystemNexusPayloadConverter(
+                        instance.payloadConverterNoContext,
```
jmaeagle99: Not a regression in this PR, but if `SignalWithStartWorkflowExecutionRequest` switches over to using transfer types, this will be a regression. The [existing code path](https://github.com/temporalio/sdk-dotnet/blob/1.18.0/src/Temporalio/Workflows/Generated/Models.cs#L113) uses `Workflow.PayloadConverter` to [convert the input args](https://github.com/temporalio/sdk-dotnet/blob/1.18.0/src/Temporalio/Workflows/Generated/TemporalSupport.cs#L64).

src/Temporalio/Nexus/SystemNexusPayloadConverter.cs
```diff
@@ -0,0 +1,50 @@
...
+                new DefaultPayloadConverter(new BinaryProtoConverter()));
+        }
+
+        /// <inheritdoc />
+        public Temporalio.Api.Common.V1.Payload ToPayload(object? value)
```
jmaeagle99: Add a using for `Temporalio.Api.Common.V1` and don't fully qualify throughout the file.

src/Temporalio/Worker/WorkflowInstance.cs
jmaeagle99: Have you considered the impact on the workflow replayer? I think similar support might be needed there.

src/Temporalio/Nexus/SystemNexusConverterContext.cs
jmaeagle99: And if not... if this is meant for use by public, then it should probably be named after where it is intended to be consumed. I think that might be `ITemporalTransferTypeConverter`. So it should probably be named for that since there is nothing public facing named "SystemNexus" in the API. And maybe should be plumbed through the `ITemporalTransferTypeConverter` API rather than being an async local while these APIs are still early in their life.

src/Temporalio/Worker/WorkflowInstance.cs
tconley1428: Kinda not really though. The one provided by the generated stuff will be the workflow context.

tests/Temporalio.Tests/Worker/SystemNexusTests.cs
tconley1428: Integration with it is currently not available since as you pointed out elsewhere, it doesn't use transfer type converter yet.

src/Temporalio/Nexus/SystemNexusConverterContext.cs
tconley1428: It's meant for use by the generated code, which since it is generated into this project should be able to use internal, but that will mean the equivalent generated code will not be able to compile or run as an example in nexgen, though we could fork it slightly there, it would be a little unfortunate to have to deviate.

src/Temporalio/Worker/WorkflowInstance.cs
tconley1428: I haven't, here or in Python. That's a very good point.

tconley1428: I'm not sure I follow this concern. Is the idea that we would use this same thing for summaries/memos, etc.? That's not currently something that we are making possible with system nexus. All relevant payloads are embedded in the envelope.

src/Temporalio/Nexus/SystemNexusPayloadConverter.cs
tconley1428: Yeah I think so.

src/Temporalio/Worker/WorkflowInstance.cs
tconley1428: Ah, you mean it _is_ used elsewhere. Agreed that's not the cleanest thing.

src/Temporalio/Nexus/SystemNexusPayloadConverter.cs
tconley1428: "_ is a declared local disposable, not a discard, because using must retain it until scope exit so it can call Dispose. StyleCop therefore validates its name and rejects it."

tconley1428: Seems to be disallowed

src/Temporalio/Worker/WorkflowInstance.cs
tconley1428: Added some coverage, but it appears to be the same path.

src/Temporalio/Worker/WorkflowInstance.cs
jmaeagle99: How not? Shouldn't the transfer type converters use this context to do the conversion, which this context is supplying `payloadConverterNoContext` and `failureConverterNoContext`? If they are not using those, then what's the point of this context?

src/Temporalio/Nexus/SystemNexusConverterContext.cs
tconley1428: Going to live in the support file entirely, with a call to use it from the converter.

src/Temporalio/Worker/WorkflowInstance.cs
tconley1428: Discussed offline. The context will be defined by the operation itself, which allows different system operations to have different contexts.

src/Temporalio/Worker/WorkflowInstance.cs
```diff
@@ -2761,7 +2765,7 @@ public override Task<NexusWorkflowOperationHandle<TResult>> ScheduleNexusOperati
                             // If there is a start sync fail, we have to fail the handle task and
                             // there's nothing more we can do here
                             var handle = new NexusWorkflowOperationHandleImpl<TResult>(
```
jmaeagle99, line 2769: This needs to be passed the `instance.payloadConverterNoContext` so it can be used at https://github.com/temporalio/sdk-dotnet/blob/dcf7ca3f4fc230315c6f518e3f3c1a1f52670963/src/Temporalio/Worker/WorkflowInstance.cs#L3187C53-L3187C60

src/Temporalio/Worker/WorkflowInstance.cs
```diff
@@ -2771,7 +2775,7 @@ public override Task<NexusWorkflowOperationHandle<TResult>> ScheduleNexusOperati
                                 // TODO(cretz): Support Nexus serialization context
                                 handleSource.SetException(
                                     instance.failureConverterNoContext.ToException(
-                                        syncStartFail, payloadConverter));
+                                        syncStartFail, operationPayloadConverter));
```
jmaeagle99, line 2780: This needs to use `instance.payloadConverterNoContext`.

src/Temporalio/Worker/WorkflowInstance.cs
jmaeagle99: I think we need to apply `TemporalTransferTypePayloadConverter.Wrap` at https://github.com/temporalio/sdk-dotnet/blob/dcf7ca3f4fc230315c6f518e3f3c1a1f52670963/src/Temporalio/Worker/WorkflowReplayer.cs#L221 just like is done within the `TemporalClient` at https://github.com/temporalio/sdk-dotnet/blob/dcf7ca3f4fc230315c6f518e3f3c1a1f52670963/src/Temporalio/Client/TemporalClient.cs#L34

tests/Temporalio.Tests/Worker/SystemNexusTests.cs
jmaeagle99: I think we should add some kind of integration testing real soon. At least something that would have checked for the correct application of the converters to the different aspects of the `ScheduleNexusOperationAsync` invocation.

src/Temporalio/Worker/WorkflowInstance.cs
tconley1428: Yeah. Missed adding a transfer type convertible to the replay test. Added coverage for that in the next commit.

src/Temporalio/Worker/WorkflowInstance.cs
tconley1428: I don't _think_ these are correct, but I might want to walk through your logic with you to be sure.

src/Temporalio/Worker/WorkflowInstance.cs
tconley1428: We think not, because the nexus handler is implemented by the server. Probably a good idea to try to add a failure case.

jmaeagle99 (approved):

PR #853, opened 2026-08-20
Add payload validation exception factory

bergundy:
Payload converters and codecs can use this to report invalid Nexus operation input with structured details.


Quinn-With-Two-Ns (approved):

src/Temporalio/Converters/PayloadValidationException.cs
```diff
@@ -0,0 +1,31 @@
...
+using System;
+using Temporalio.Exceptions;
+
+namespace Temporalio.Converters
+{
+    /// <summary>
+    /// Factory for application failures that report invalid Nexus operation input from payload
+    /// conversion.
+    /// </summary>
+    public static class PayloadValidationException
```
jmaeagle99: This is not an exception class, so it should not be named *Exception. I would suggest renaming to `PayloadValidationError` and renaming the factory method to `CreateException`. Please update the change log to reflect this change too.

src/Temporalio/Converters/PayloadValidationException.cs
```diff
@@ -0,0 +1,31 @@
+#pragma warning disable CA1711 // Factory naming matches the payload validation API across SDKs
```
jmaeagle99: This pragma will not be needed once the rename occurs.

src/Temporalio/Converters/PayloadValidationException.cs
```diff
@@ -0,0 +1,31 @@
...
+        /// Creates a non-retryable application failure for invalid Nexus operation input.
+        /// </summary>
+        /// <param name="details">Validation error details to serialize with the failure.</param>
+        /// <returns>The payload validation application failure.</returns>
+        public static ApplicationFailureException Create(object? details) =>
+            new(
+                "Payload validation failed",
+                errorType: ErrorType,
+                nonRetryable: true,
+                details: details == null ? Array.Empty<object?>() : new object?[] { details });
```
jmaeagle99: Every other SDK is producing one detail object unconditionally, regardless of details being `null` / `None`.  Either this should match or the other SDKs corrected.

src/Temporalio/Worker/NexusPayloadSerializer.cs
```diff
@@ -144,6 +136,6 @@ public NexusPayloadSerializer(DataConverter dataConverter) =>
         private static bool IsPayloadValidationFailure(Exception e) =>
```
jmaeagle99: I would recommend moving this to `PayloadValidationError` as an internal method. Since it creates the validation exception, it should also decide whether something is a validation exception or not.

jmaeagle99 (changes requested):

src/Temporalio/Converters/PayloadValidationException.cs
bergundy: Renamed the helper to PayloadValidationError, renamed the factory method to CreateException, and updated the changelog and call sites in c693292.

src/Temporalio/Converters/PayloadValidationException.cs
bergundy: Removed the CA1711 pragma as part of the rename in c693292.

src/Temporalio/Converters/PayloadValidationException.cs
bergundy: The null behavior is intentional for this API: null means there are no validation details, so the failure contains zero detail values. I kept that contract and clarified it in the parameter documentation; PayloadValidationErrorTests.CreateException_WithNullDetails_CreatesApplicationFailureWithoutDetails covers it. The other SDKs should be aligned separately where their details APIs can distinguish absent details.

src/Temporalio/Worker/NexusPayloadSerializer.cs
bergundy: Moved the predicate to PayloadValidationError.IsException as an internal method and use it for both codec and converter failures in c693292.

src/Temporalio/Converters/PayloadValidationException.cs
jmaeagle99: Please follow up with changes to the other SDKs immediately or create issues in their repos.

jmaeagle99 (approved):

PR #855, opened 2026-08-21
Add Google Cloud Run CloudRunIDPlugin

seanbollin:
## What

Adds an experimental Google Cloud Run worker-identity plugin — the Cloud Run counterpart to the SDK's AWS Lambda extension (`Temporalio.Extensions.Aws.Lambda`), and a sibling to the `Temporalio.Extensions.Gcp.CloudRun.OpenTelemetry` package — for both Cloud Run worker pools and services. It ships as the `Temporalio.Extensions.Gcp.CloudRun.Id` package. Cloud Run runs a long-lived container, so this is a small client plugin rather than a worker wrapper.

> **Naming:** the plugin is `CloudRunIDPlugin` in the `Temporalio.Extensions.Gcp.CloudRun.Id` package. It sets the client identity — which the workers then inherit — and the name carries the Cloud Run context without relying on the full namespace.

## What it does

Register `CloudRunIDPlugin` on your client:

```csharp
var options = new TemporalClientConnectOptions("localhost:7233")
{
    Plugins = new[] { new CloudRunIDPlugin() },
};
```

When the client connects, the plugin reads the current Cloud Run instance's metadata — the revision (`CLOUD_RUN_REVISION` / `K_REVISION`), the worker pool or service name (`CLOUD_RUN_WORKER_POOL` / `K_SERVICE`), and the unique instance id from the metadata server (`http://metadata.google.internal/computeMetadata/v1/instance/id`) — and sets the client `Identity` to `{InstanceId}@{Revision}` when one isn't already set. Workers created from that client inherit it.

`GoogleCloudRunMetadata` / `Identity` are exposed for reading the values directly. Experimental.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


src/Temporalio.Extensions.Gcp.CloudRun.Id/GoogleCloudRunMetadata.cs
```diff
@@ -0,0 +1,181 @@
+using System;
+using System.Net.Http;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Temporalio.Extensions.Gcp.CloudRun.WorkerId
+{
+    /// <summary>
+    /// Reads Google Cloud Run instance metadata to derive a Temporal worker identity for a
+    /// long-lived worker, on both Cloud Run worker pools and services.
```
Sushisource:
```suggestion
    /// Reads Google Cloud Run instance metadata to derive a Temporal worker identity on both Cloud Run worker pools and services.
```

src/Temporalio.Extensions.Gcp.CloudRun.Id/CloudRunIdPluginOptions.cs
```diff
@@ -0,0 +1,44 @@
...
+    {
+        /// <summary>
+        /// Gets or sets pre-fetched Cloud Run metadata for the plugin to use.
+        /// </summary>
+        /// <remarks>
+        /// When set, the plugin uses this metadata directly and never contacts the metadata server,
+        /// so <see cref="MetadataUri" /> and <see cref="Timeout" /> are ignored. This is primarily
+        /// for tests and advanced scenarios that resolve the metadata themselves.
+        /// </remarks>
+        public GoogleCloudRunMetadata? Metadata { get; set; }
```
Sushisource, line 19: Same comment as on Go here (I maybe missed it in Java as well) where ideally this would just be package-private.

Sushisource (approved):

src/Temporalio.Extensions.Gcp.CloudRun.Id/GoogleCloudRunMetadata.cs
```diff
@@ -0,0 +1,181 @@
...
+                ServiceRevisionEnvironmentVariable);
+
+            using var httpClient = new HttpClient { Timeout = timeout };
+            using var request = new HttpRequestMessage(HttpMethod.Get, metadataUri);
+            request.Headers.Add(MetadataFlavorHeader, MetadataFlavorValue);
+
+            string instanceId;
+            try
+            {
+                using var response = await httpClient.SendAsync(request, cancellationToken).
```
jmaeagle99, line 125: This will throw `TaskCanceledException` when the timeout occurs. I would catch with `catch (OperationCanceledException e) when (!cancellationToken.IsCancellationRequested)` and handle / rethrow appropriately. And add a test for it.

src/Temporalio.Extensions.Gcp.CloudRun.Id/GoogleCloudRunMetadata.cs
```diff
@@ -0,0 +1,181 @@
...
+        /// <c>CLOUD_RUN_REVISION</c> then <c>K_REVISION</c> for the revision. Worker pools receive
+        /// the <c>CLOUD_RUN_*</c> variables and services receive the <c>K_*</c> variables. The
+        /// instance id is read from the metadata server, which is available on both and requires the
+        /// <c>Metadata-Flavor: Google</c> request header.
+        /// </remarks>
+        /// <exception cref="InvalidOperationException">
+        /// Thrown when the instance id cannot be read from the metadata server, which usually means
+        /// the process is not running on a Google Cloud Run worker pool or service.
+        /// </exception>
+        public static async Task<GoogleCloudRunMetadata> FetchAsync(
```
jmaeagle99, line 105: I would recommend making the `metadataUri` and `timeout` parameters nullable with default `null` value and dropping the `FetchAsync(CancellationToken cancellationToken = default)` method overload and the `FetchWithDefaultsAsync` method.

src/Temporalio.Extensions.Gcp.CloudRun.Id/GoogleCloudRunMetadata.cs
```diff
@@ -0,0 +1,181 @@
...
+            using var request = new HttpRequestMessage(HttpMethod.Get, metadataUri);
+            request.Headers.Add(MetadataFlavorHeader, MetadataFlavorValue);
+
+            string instanceId;
+            try
+            {
+                using var response = await httpClient.SendAsync(request, cancellationToken).
+                    ConfigureAwait(false);
+                response.EnsureSuccessStatusCode();
+                instanceId = (await response.Content.ReadAsStringAsync().ConfigureAwait(false)).
```
jmaeagle99, line 128: What if `instanceId` is empty? The `GoogleCloudRunMetadata` class is formatting other values that use this as an input and would have WorkerIdentity return "@revision-1".

src/Temporalio.Extensions.Gcp.CloudRun.Id/README.md
```diff
@@ -0,0 +1,101 @@
...
+
+using var worker = new TemporalWorker(
+    client,
+    new TemporalWorkerOptions("my-task-queue").
+        AddWorkflow<MyWorkflow>().
+        AddActivity(MyActivities.DoThing));
+
+// Cloud Run sends SIGTERM before stopping the instance; cancel the worker on it.
+using var shutdown = new CancellationTokenSource();
+Console.CancelKeyPress += (_, eventArgs) =>
```
jmaeagle99: `CancelKeyPress` is SIGINT, not SIGTERM. If CloudRun is sending SIGTERM and SIGKILL, then this handler will never see those; the worker will not be gracefully shut down.

jmaeagle99: Probably need another handler like:

```
using var sigterm = PosixSignalRegistration.Create(PosixSignal.SIGTERM, ctx =>
{
    ctx.Cancel = true;
    shutdown.Cancel();
});
```

This only works on .NET 6+

src/Temporalio.Extensions.Gcp.CloudRun.Id/CloudRunIDPlugin.cs
```diff
@@ -0,0 +1,101 @@
...
+    /// Cloud Run worker identity, so an explicitly configured identity always wins. Workers created
+    /// from the connected client inherit that identity.
+    /// <para>
+    /// The metadata fetch throws an <see cref="InvalidOperationException" /> at connect time when the
+    /// process is not running on a Cloud Run worker pool or service. Tests and advanced
+    /// users can bypass the real fetch with <see cref="CloudRunIDPluginOptions" />.
+    /// </para>
+    /// WARNING: Google Cloud Run support is experimental.
+    /// </remarks>
+    public class CloudRunIDPlugin : SimplePlugin
```
jmaeagle99: One shortcoming of using a plugin is that if the Temporal client is lazily created, like it is in dependency injection, then the plugin will not be applied. Please document this in the README. Haven't thought through if this is fixable or not, might be worth an issue.

jmaeagle99:
```suggestion
    public class CloudRunIdPlugin : SimplePlugin
```

The repo convention is to use lowercase `d` when using `Id` in names.

src/Temporalio.Extensions.Gcp.CloudRun.WorkerId/CloudRunIDPlugin.cs
```diff
@@ -0,0 +1,101 @@
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Temporalio.Client;
+using Temporalio.Common;
+
+namespace Temporalio.Extensions.Gcp.CloudRun.WorkerId
```
jmaeagle99: The `WorkerId` segment of the namespace and of the package name now stands out as inconsistent. But it looks like this is inconsistent across the SDKs too:
- [Go](https://github.com/temporalio/sdk-go/pull/2599/changes#diff-3456b8eb630cdb1063e9fcb377997ee97047e8a4d0e43b50f915528344c371c5R1): `go.temporal.io/sdk/contrib/gcp/cloudrun/id`
- [Python](https://github.com/temporalio/sdk-python/pull/1776/changes#diff-2fc63fbe9a3e73a56c4d428dbfaf3f43396fd5a3803a2abd1448d3991b7fbd8eR10): `temporalio.contrib.gcp.cloud_run.worker_id`
- [Java](https://github.com/temporalio/sdk-java/pull/3028/changes#diff-f41e2587ed1e8aa0894f85a89a268f067fe598ee0bf2842303db33c504467043R1): `io.temporal.gcp.cloudrun.workerid`

seanbollin: I think you likely just caught it mid-refactor, everything should be normalized on CloudRunId now

src/Temporalio.Extensions.Gcp.CloudRun.Id/README.md
```diff
@@ -0,0 +1,95 @@
...
+}
+catch (OperationCanceledException)
+{
+    // Expected shutdown path.
+}
+```
+
+> **Note:** the plugin applies the identity when the client connects, so it only runs if your client
+> connects with the plugin registered. If the client is created lazily — as some dependency-injection
+> setups do — the plugin may not be applied; register it on the connect options the client is built from.
```
jmaeagle99: The remedy is ambiguous; "it" seems to refer to the plugin, which is what was already done. I think it needs to say something like "In that case set the identity yourself before building the client:" and provide the example:

```csharp
var metadata = await GoogleCloudRunMetadata.FetchAsync();
connectOptions.Identity = metadata.Identity;
```

jmaeagle99 (approved):
The build failures are due to the stale lock files. Other than the one clarification in the readme, this now LGTM.

PR #852, opened 2026-08-19
feat: apply transfer converters to System Nexus

tconley1428:
## Summary

- scopes the System Nexus payload/failure converter context while serializing and deserializing outer System Nexus envelopes
- uses a local unreleased nexgen build to generate .NET transfer-type converters and context-backed support helpers
- generates a payload visitor for payload-bearing Workflow Service envelopes; any payload marked as a System Nexus envelope always bypasses the user codec

## Status

**Work in progress — do not merge yet.** This PR must wait for the nexgen changes in https://github.com/temporalio/nexgen/pull/128 to be released, then replace the local-generator integration with that released version.

There is also a server dependency: System Nexus result envelopes are not currently marked by the server. Until that behavior exists, codec coverage for marked result envelopes cannot be enabled; the end-to-end Signal-with-Start test intentionally uses no payload codec.

## Validation

- `dotnet test tests/Temporalio.Tests/Temporalio.Tests.csproj --no-restore --filter 'FullyQualifiedName~SystemNexusPayloadConverterTests|FullyQualifiedName~SystemNexusTests'`\n- `dotnet format --verify-no-changes --no-restore` (blocked only by pre-existing unused imports in `tests/Temporalio.SimpleBench/Program.cs`)\n- `dotnet build --no-restore` (blocked because `tests/Temporalio.SimpleBench/obj/project.assets.json` is absent)

PR #863, opened 2026-08-26
Add System Nexus signal-with-start support

tconley1428:
## Summary

- Consolidates the System Nexus transfer-converter context work formerly proposed in #852 with Signal-with-Start support.
- Generates the System Nexus API with nexgen from the API WIT vendored through SDK Core.
- Applies generated System Nexus serialization context when scheduling Signal-with-Start.

## Validation

- Local nexgen generation against the API WIT input
- `dotnet build`
- Focused System Nexus tests: 4 passed


src/Temporalio.SystemNexus.Generator/Program.cs
```diff
@@ -90,7 +83,7 @@ void PostProcessGeneratedNexusApi()
 
     WriteWorkflowGeneratedFile("Models.cs");
     WriteWorkflowGeneratedFile("Operations.cs");
-    WriteWorkflowGeneratedFile("Service.cs");
+    WriteWorkflowGeneratedFile("Services.cs", "Service.cs");
```
jmaeagle99: Why not rename the committed file?

src/Temporalio.SystemNexus.Generator/Program.cs
```diff
@@ -152,6 +128,7 @@ static void GeneratePayloadVisitor(
     builder.AppendLine("// <auto-generated />");
     builder.AppendLine("// Generated by Temporalio.SystemNexus.Generator. DO NOT EDIT!");
     builder.AppendLine("#nullable enable");
+    builder.AppendLine("#pragma warning disable CS0612");
```
jmaeagle99, line 131: What was obsoleted?

src/Temporalio/Worker/Interceptors/ScheduleSystemNexusOperationInput.cs
```diff
@@ -0,0 +1,20 @@
...
+    /// </summary>
+    /// <typeparam name="TResult">System Nexus operation result type.</typeparam>
+    /// <param name="Service">System Nexus service name.</param>
+    /// <param name="Operation">System Nexus operation definition.</param>
+    /// <param name="Arg">Generated request argument.</param>
+    /// <remarks>
+    /// System Nexus operations do not expose normal Nexus endpoint, scheduling, cancellation, or
+    /// transport-header options. Interceptors may modify the generated request argument.
+    /// </remarks>
+    public record ScheduleSystemNexusOperationInput<TResult>(
```
jmaeagle99: Is the generic parameter `TResult` necessary?

src/Temporalio/Worker/WorkflowInstance.cs
```diff
@@ -2681,15 +2681,33 @@ public override Task<NexusWorkflowOperationHandle<TResult>> ScheduleNexusOperati
...
                 }
 
-                // TODO: Scope the generated System Nexus support converter context around this
-                // operation converter once the generated support file is ingested into the SDK.
+                ISerializationContext? serializationContext = null;
+                if (SystemNexusPayloadVisitor.IsSystemEndpoint(input.ClientOptions.Endpoint) &&
+                    NexgenOperationRegistry.Operations.TryGetValue(
+                        (input.Service, input.OperationName), out var operationInfo))
+                {
+                    serializationContext = operationInfo.SerializationContext?.Invoke(input.Arg!);
```
jmaeagle99: Because the interceptor allows modifying the `Arg` property, we need to assume that it could be null, despite nullability annotations (because you can set something to null with `= null!` or deserialization often doesn't initialize non-null properties).

src/Temporalio.Extensions.OpenTelemetry/TracingInterceptor.cs
```diff
@@ -748,6 +772,21 @@ public override Task<NexusWorkflowOperationHandle<TResult>> ScheduleNexusOperati
                 return base.ScheduleNexusOperationAsync<TResult>(input);
             }
 
+            public override Task<NexusWorkflowOperationHandle<TResult>> ScheduleSystemNexusOperationAsync<TResult>(
+                ScheduleSystemNexusOperationInput<TResult> input)
+            {
+                var name = input.Operation.Name == "SignalWithStartWorkflowExecution" ?
+                    "SignalWithStartWorkflow" :
```
jmaeagle99: Should this be "SignalWithStartWorkflow:{workflowType}"? Most spans do not have a bare identifier so that way you can distinguish different calls within the same parent span.

src/Temporalio.SystemNexus.Generator/Program.cs
tconley1428: Good question, didn't catch the bot doing this.

src/Temporalio.SystemNexus.Generator/Program.cs
tconley1428: Old protos in the workflow service that are now used because we generate many more visitor functions.

src/Temporalio/Worker/Interceptors/ScheduleSystemNexusOperationInput.cs
tconley1428: Nope

src/Temporalio/Worker/WorkflowInstance.cs
tconley1428: Annoying but good to be defensive.

src/Temporalio.Extensions.OpenTelemetry/TracingInterceptor.cs
tconley1428: Yeah, it should closely match the other ones I think. I'll need to reflect for this temporarily, but can fix it when nexgen updates and makes the models public.

jmaeagle99 (approved):

PR #867, opened 2026-08-29
Support generic transfer type converters

VegetarianOrc:
## What was changed

- Generic transfer type converters can now be applied to models that have the same number of type arguments. 

## Why?

- This is needed to support transfer type conversion where the type arguments are not known until runtime.

## Checklist

1. Closes: N/A

2. How was this tested:

- Added tests covering newly supported generic models.

### DIFF
--- a/src/Temporalio/Converters/TemporalTransferTypePayloadConverter.cs
+++ b/src/Temporalio/Converters/TemporalTransferTypePayloadConverter.cs
@@ -111,12 +111,12 @@ namespace Temporalio.Converters
                 {
                     converterType = converterType.MakeGenericType(typeArguments);
                 }
-                catch (ArgumentException err)
+                catch (ArgumentException e)
                 {
                     throw new InvalidOperationException(
                         $"Type {type} cannot close Temporal transfer type converter type " +
                         $"{attr.ConverterType} because its generic arguments do not satisfy the converter constraints.",
-                        err);
+                        e);
                 }
             }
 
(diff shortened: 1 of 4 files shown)


src/Temporalio/Converters/TemporalTransferTypePayloadConverter.cs
```diff
@@ -83,43 +83,80 @@ public IPayloadConverter WithSerializationContext(ISerializationContext context)
...
-            if (attr.ConverterType.IsAbstract)
+            if (converterType.IsAbstract)
             {
                 throw new InvalidOperationException(
                     $"Type {type} has an abstract Temporal transfer type converter type " +
-                    $"{attr.ConverterType}.");
+                    $"{converterType}.");
             }
-            if (attr.ConverterType.ContainsGenericParameters)
+            if (converterType.ContainsGenericParameters)
```
jmaeagle99: This is a dead branch.

src/Temporalio/Converters/TemporalTransferTypePayloadConverter.cs
```diff
@@ -83,43 +83,80 @@ public IPayloadConverter WithSerializationContext(ISerializationContext context)
...
+                    throw new InvalidOperationException(
+                        $"Type {type} and its open generic Temporal transfer type converter type " +
+                        $"{attr.ConverterType} have different generic arities.");
+                }
+
+                try
+                {
+                    converterType = converterType.MakeGenericType(typeArguments);
+                }
+                catch (ArgumentException err)
```
jmaeagle99: nit: use `e` or `ex` in .NET

jmaeagle99 (approved):

PR #869, opened 2026-08-31
💥 Update ActivityExecutionDescription to support new API, rename StaticSummary to Summary

maciejdudko:
## 💥 **Breaking changes**
Changes to experimental `ActivityExecutionDescription` class and related classes:
- `StaticSummary` and `ScheduledTime` were renamed to `Summary` and `ScheduleTime` respectively.
- `LongPollToken` and `StateTransitionCount` were removed.

## What was changed

- Support requesting optional payload fields in `ActivityHandle.DescribeAsync` and expose them in `ActivityExecutionDescription`.
- Added, removed and renamed properties in `ActivityExecutionDescription` to match latest API.

## Why?

- Exposes latest functionality of `DescribeActivityExecution` RPC.
- Aligns exposed APIs with the latest design.

## Checklist

1. How was this tested:

New tests in `TemporalClientActivityTests`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,20 +21,27 @@ to docs, or any other relevant information.
 
 ### :boom: Breaking Changes
 
-- Changes to experimental `ActivityExecutionDescription` class and related classes:
-  - `StaticSummary` and `ScheduledTime` were renamed to `Summary` and `ScheduleTime` respectively.
-  - `LongPollToken` and `StateTransitionCount` were removed.
+- Renamed and removed several experimental methods and properties in Standalone Activities APIs:
+  - `StartActivityOptions.StaticSummary` renamed to `Summary`.
+  - `ActivityExecutionDescription.GetStaticSummaryAsync` renamed to `GetSummaryAsync`.
+  - `ActivityExecution.ScheduledTime` renamed to `ScheduleTime`.
+  - Removed `ActivityExecution.StateTransitionCount`.
+  - Removed `ActivityExecutionDescription.LongPollToken`.
+  - Removed `ActivityDescribeOptions.LongPollToken`.
 - Removed the experimental `SignalWithStartWorkflowOptions.RequestId`. Request IDs for
   workflow-side signal-with-start are now assigned internally and are no longer user-settable.
 
 ### Added
 
+- New options in `ActivityDescribeOptions` that can be used to retrieve data associated with
+  activity execution, such as input and result.
+- New properties and methods in `ActivityExecution` and `ActivityExecutionDescription`:
+  `ExecutionTime`, `LastDeploymentVersion`, `Priority`, `RetryPolicy`, `StartDelay`,
+  `TotalHeartbeatCount`, `GetLastFailureAsync`, `GetOutcomeFailureAsync`, `GetResultAsync`,
+  and access to raw Proto objects from the response.
 - Added operation-specific outbound workflow interceptors for Temporal System Nexus operations.
   `SignalWithStartWorkflowAsync` can now be intercepted before the existing
   `ScheduleSystemNexusOperationAsync` hook.
-- New options in `ActivityDescribeOptions` that can be used to retrieve data associated with
-  activity execution, such as input and result.
-- More properties in `ActivityDescribeOptions`.
 - Experimental Temporal transfer type conversion now supports constructed generic models whose
   attributes reference generic converter type definitions. Model type arguments close the converter
   directly in declaration order, with matching generic arity and compatible constraints required.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,9 @@ to docs, or any other relevant information.
 
 ### :boom: Breaking Changes
 
+- Changes to experimental `ActivityExecutionDescription` class and related classes:
+  - `StaticSummary` and `ScheduledTime` were renamed to `Summary` and `ScheduleTime` respectively.
+  - `LongPollToken` and `StateTransitionCount` were removed.
 - Removed the experimental `SignalWithStartWorkflowOptions.RequestId`. Request IDs for
   workflow-side signal-with-start are now assigned internally and are no longer user-settable.
 
@@ -29,6 +32,9 @@ to docs, or any other relevant information.
 - Added operation-specific outbound workflow interceptors for Temporal System Nexus operations.
   `SignalWithStartWorkflowAsync` can now be intercepted before the existing
   `ScheduleSystemNexusOperationAsync` hook.
+- New options in `ActivityDescribeOptions` that can be used to retrieve data associated with
+  activity execution, such as input and result.
+- More properties in `ActivityDescribeOptions`.
 - Experimental Temporal transfer type conversion now supports constructed generic models whose
   attributes reference generic converter type definitions. Model type arguments close the converter
   directly in declaration order, with matching generic arity and compatible constraints required.
--- a/src/Temporalio/Client/ActivityDescribeOptions.cs
+++ b/src/Temporalio/Client/ActivityDescribeOptions.cs
@@ -9,10 +9,32 @@ namespace Temporalio.Client
     public class ActivityDescribeOptions : ICloneable
     {
         /// <summary>
-        /// Gets or sets a long-poll token from a previous describe response. When set, the
-        /// describe call will long-poll for state changes.
+        /// Gets or sets a value indicating whether to include input in the response if available.
         /// </summary>
-        public byte[]? LongPollToken { get; set; }
+        /// <seealso cref="ActivityExecutionDescription.HasInput"/>
+        public bool IncludeInput { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value indicating whether to include outcome in the response if available.
+        /// </summary>
+        /// <seealso cref="ActivityExecutionDescription.HasResult"/>
+        /// <seealso cref="ActivityExecutionDescription.HasOutcomeFailure"/>
+        /// <seealso cref="ActivityExecutionDescription.GetResultAsync"/>
+        /// <seealso cref="ActivityExecutionDescription.GetOutcomeFailureAsync"/>
+        public bool IncludeOutcome { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value indicating whether to include heartbeat details in the response if available.
+        /// </summary>
+        /// <seealso cref="ActivityExecutionDescription.HasHeartbeatDetails"/>
+        public bool IncludeHeartbeatDetails { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value indicating whether to include last failure in the response if available.
+        /// </summary>
+        /// <seealso cref="ActivityExecutionDescription.HasLastFailure"/>
+        /// <seealso cref="ActivityExecutionDescription.GetLastFailureAsync"/>
+        public bool IncludeLastFailure { get; set; }
 
         /// <summary>
         /// Gets or sets RPC options for describing the activity.
--- a/src/Temporalio/Client/ActivityExecutionDescription.cs
+++ b/src/Temporalio/Client/ActivityExecutionDescription.cs
@@ -18,6 +18,7 @@ namespace Temporalio.Client
     /// <remarks>WARNING: Standalone activities are experimental.</remarks>
     public class ActivityExecutionDescription : ActivityExecution
     {
+        private readonly ActivityDescribeOptions requestOptions;
         private readonly DataConverter dataConverter;
         private readonly Lazy<Task<(string? Summary, string? Details)>> userMetadata;
         private readonly Lazy<Task<Exception?>> lastFailure;
@@ -26,11 +27,13 @@ namespace Temporalio.Client
         /// <summary>
         /// Initializes a new instance of the <see cref="ActivityExecutionDescription"/> class.
         /// </summary>
+        /// <param name="requestOptions">Options used to make the request.</param>
         /// <param name="resp">Raw proto response.</param>
         /// <param name="clientNamespace">Client namespace.</param>
         /// <param name="dataConverter">Data converter.</param>
         /// <remarks>WARNING: This constructor may be mutated in backwards incompatible ways.</remarks>
         protected internal ActivityExecutionDescription(
+            ActivityDescribeOptions? requestOptions,
             DescribeActivityExecutionResponse resp,
             string clientNamespace,
             DataConverter dataConverter)
@@ -49,8 +52,9 @@ namespace Temporalio.Client
                     SearchAttributeCollection.Empty :
                     SearchAttributeCollection.FromProto(resp.Info.SearchAttributes))
         {
-            RawResponse = resp;
+            this.requestOptions = requestOptions ?? new();
             this.dataConverter = dataConverter;
+            RawResponse = resp;
             var info = resp.Info;
             Attempt = info.Attempt;
             CanceledReason = string.IsNullOrEmpty(info.CanceledReason) ? null : info.CanceledReason;
@@ -73,8 +77,8 @@ namespace Temporalio.Client
             TotalHeartbeatCount = info.TotalHeartbeatCount;
 #pragma warning disable VSTHRD011 // This should not be able to deadlock
             userMetadata = new(() => dataConverter.FromUserMetadataAsync(info.UserMetadata));
-            lastFailure = new(async () => info.LastFailure == null ? null : await dataConverter.ToExceptionAsync(info.LastFailure).ConfigureAwait(false));
-            outcomeFailure = new(async () => resp.Outcome?.Failure == null ? null : await dataConverter.ToExceptionAsync(resp.Outcome.Failure).ConfigureAwait(false));
+            lastFailure = new(async () => info.LastFailure == null ? null : await dataConverter.ToExceptionAsync(info.LastFailure.Clone()).ConfigureAwait(false));
+            outcomeFailure = new(async () => resp.Outcome?.Failure == null ? null : await dataConverter.ToExceptionAsync(resp.Outcome.Failure.Clone()).ConfigureAwait(false));
 #pragma warning restore VSTHRD011
         }
 
@@ -105,33 +109,34 @@ namespace Temporalio.Client
         /// Always false if <see cref="ActivityDescribeOptions.IncludeHeartbeatDetails"/> was false.
         /// <para>Heartbeat details payloads can be accessed via <c>RawInfo.HeartbeatDetails.Payloads_</c>.</para>
         /// </remarks>
-        public bool HasHeartbeatDetails => RawInfo.HeartbeatDetails?.Payloads_?.Count > 0;
+        public bool HasHeartbeatDetails => requestOptions.IncludeHeartbeatDetails && RawInfo.HeartbeatDetails?.Payloads_?.Count > 0;
 
         /// <summary>
-        /// Gets a value indicating whether input is available.
+        /// Gets a value indicating whether the activity input is available.
         /// </summary>
         /// <remarks>
         /// Always false if <see cref="ActivityDescribeOptions.IncludeInput"/> was false.
         /// <para>Input payloads can be accessed via <see cref="RawInput"/>.</para>
         /// </remarks>
-        public bool HasInput => RawInput?.Count > 0;
+        public bool HasInput => requestOptions.IncludeInput && RawInput?.Count > 0;
 
         /// <summary>
         /// Gets a value indicating whether the last failure is available.
         /// </summary>
-        /// Always false if <see cref="ActivityDescribeOptions.IncludeLastFailure"/> was false.
+        /// <remarks>Always false if <see cref="ActivityDescribeOptions.IncludeLastFailure"/> was false.</remarks>
         /// <seealso cref="GetLastFailureAsync"/>
-        public bool HasLastFailure => RawInfo.LastFailure != null;
+        public bool HasLastFailure => requestOptions.IncludeLastFailure && RawInfo.LastFailure != null;
 
         /// <summary>
-        /// Gets a value indicating whether the last failure is available.
+        /// Gets a value indicating whether the activity result is available.
         /// </summary>
         /// <remarks>
-        /// Activity outcome failure is only available if the activity has closed with a failure.
+        /// Activity result is only available if the activity has completed successfully.
         /// <para>Always false if <see cref="ActivityDescribeOptions.IncludeOutcome"/> was false.</para>
         /// </remarks>
-        /// <seealso cref="GetOutcomeFailureAsync"/>
-        public bool HasResult => RawOutcome?.Result?.Payloads_.Count == 1; // 0 means missing, more than 1 is invalid
+        /// <seealso cref="GetResultAsync"/>
+        /// <seealso cref="HasOutcomeFailure"/>
+        public bool HasResult => requestOptions.IncludeOutcome && RawOutcome?.Result?.Payloads_.Count == 1; // 0 means missing, more than 1 is invalid
 
         /// <summary>
         /// Gets a value indicating whether the outcome failure is available.
@@ -141,7 +146,8 @@ namespace Temporalio.Client
         /// <para>Always false if <see cref="ActivityDescribeOptions.IncludeOutcome"/> was false.</para>
         /// </remarks>
         /// <seealso cref="GetOutcomeFailureAsync"/>
-        public bool HasOutcomeFailure => RawOutcome?.Failure != null;
+        /// <seealso cref="HasResult"/>
+        public bool HasOutcomeFailure => requestOptions.IncludeOutcome && RawOutcome?.Failure != null;
 
         /// <summary>
         /// Gets the heartbeat timeout.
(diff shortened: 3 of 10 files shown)


src/Temporalio/Client/ActivityExecutionDescription.cs
```diff
@@ -13,162 +18,287 @@ namespace Temporalio.Client
...
+            RetryPolicy = info.RetryPolicy == null ? null : RetryPolicy.FromProto(info.RetryPolicy);
             RunState = info.RunState;
             ScheduleToCloseTimeout = info.ScheduleToCloseTimeout?.ToTimeSpan();
             ScheduleToStartTimeout = info.ScheduleToStartTimeout?.ToTimeSpan();
+            StartDelay = info.StartDelay?.ToTimeSpan();
             StartToCloseTimeout = info.StartToCloseTimeout?.ToTimeSpan();
+            TotalHeartbeatCount = info.TotalHeartbeatCount;
 #pragma warning disable VSTHRD011 // This should not be able to deadlock
             userMetadata = new(() => dataConverter.FromUserMetadataAsync(info.UserMetadata));
+            lastFailure = new(async () => info.LastFailure == null ? null : await dataConverter.ToExceptionAsync(info.LastFailure).ConfigureAwait(false));
```
jmaeagle99: `ToExceptionAsync` can mutate the argument in-place if a payload codec is configured. Clone the objects before passing them e.g. `(Failure)info.LastFailure.Clone()`. Same for `resp.Outcome.Failure`.

src/Temporalio/Client/ActivityExecutionDescription.cs
```diff
@@ -13,162 +18,287 @@ namespace Temporalio.Client
...
+        /// <remarks>
+        /// Always false if <see cref="ActivityDescribeOptions.IncludeInput"/> was false.
+        /// <para>Input payloads can be accessed via <see cref="RawInput"/>.</para>
+        /// </remarks>
+        public bool HasInput => RawInput?.Count > 0;
+
+        /// <summary>
+        /// Gets a value indicating whether the last failure is available.
+        /// </summary>
+        /// Always false if <see cref="ActivityDescribeOptions.IncludeLastFailure"/> was false.
```
jmaeagle99: Needs to be inside a `<remark></remark>` block.

src/Temporalio/Client/ActivityExecutionDescription.cs
```diff
@@ -13,162 +18,287 @@ namespace Temporalio.Client
...
+        /// <seealso cref="GetLastFailureAsync"/>
+        public bool HasLastFailure => RawInfo.LastFailure != null;
+
+        /// <summary>
+        /// Gets a value indicating whether the last failure is available.
```
jmaeagle99: Copy-paste error from `HasLastFailure`.

src/Temporalio/Client/ActivityExecutionDescription.cs
```diff
@@ -13,162 +18,287 @@ namespace Temporalio.Client
...
+        /// <remarks>WARNING: This method is experimental.</remarks>
         /// <returns>Static details.</returns>
-#pragma warning disable VSTHRD003 // Awaiting our own lazily-created task
         public async Task<string?> GetStaticDetailsAsync() =>
             (await userMetadata.Value.ConfigureAwait(false)).Details;
+
+        /// <summary>
+        /// Gets the failure from the last failed attempt, or null if not available.
+        /// </summary>
+        /// <remarks>Always false if <see cref="ActivityDescribeOptions.IncludeLastFailure"/> was false.</remarks>
```
jmaeagle99:
```suggestion
        /// <remarks>Always null if <see cref="ActivityDescribeOptions.IncludeLastFailure"/> was false.</remarks>
```

src/Temporalio/Client/ActivityExecutionDescription.cs
```diff
@@ -13,162 +18,287 @@ namespace Temporalio.Client
...
+        /// <seealso cref="HasLastFailure"/>
+        public async Task<Exception?> GetLastFailureAsync() =>
+            await lastFailure.Value.ConfigureAwait(false);
+
+        /// <summary>
+        /// Gets the failure of the activity execution, or null if not available.
+        /// </summary>
+        /// <remarks>
+        /// Activity outcome failure is only available if the activity has closed with a failure.
+        /// <para>Always false if <see cref="ActivityDescribeOptions.IncludeOutcome"/> was false.</para>
```
jmaeagle99:
```suggestion
        /// <para>Always null if <see cref="ActivityDescribeOptions.IncludeOutcome"/> was false.</para>
```

src/Temporalio/Client/ActivityExecutionDescription.cs
```diff
@@ -13,162 +18,287 @@ namespace Temporalio.Client
...
 #pragma warning restore VSTHRD003
+
+        /// <summary>
+        /// Gets the result of the activity execution.
+        /// </summary>
+        /// <typeparam name="T">Type to convert the result into.</typeparam>
+        /// <exception cref="InvalidOperationException">If result is not available. See <see cref="HasResult"/>.</exception>
+        /// <remarks>
+        /// Activity result is only available if the activity has completed successfully.
+        /// <para>Always false if <see cref="ActivityDescribeOptions.IncludeOutcome"/> was false.</para>
```
jmaeagle99: I'd remove this line. The `<exception>` block already documents the behavior.

src/Temporalio/Client/ActivityExecutionDescription.cs
```diff
@@ -13,162 +18,287 @@ namespace Temporalio.Client
...
+        public async Task<string?> GetSummaryAsync() =>
             (await userMetadata.Value.ConfigureAwait(false)).Summary;
-#pragma warning restore VSTHRD003
 
         /// <summary>
         /// Gets the general fixed details for this activity execution that may appear in UI/CLI.
         /// This can be in Temporal markdown format and can span multiple lines.
         /// </summary>
-        /// <remarks>WARNING: Standalone activities are experimental.</remarks>
+        /// <remarks>WARNING: This method is experimental.</remarks>
```
jmaeagle99, line 263: Why is this changed? Also, the overall class still has an experimental warning but none of the other members do.

CHANGELOG.md
```diff
@@ -21,11 +21,17 @@ to docs, or any other relevant information.
...
+  - `StaticSummary` and `ScheduledTime` were renamed to `Summary` and `ScheduleTime` respectively.
+  - `LongPollToken` and `StateTransitionCount` were removed.
 - Removed the experimental `SignalWithStartWorkflowOptions.RequestId`. Request IDs for
   workflow-side signal-with-start are now assigned internally and are no longer user-settable.
 
 ### Added
 
+- New options in `ActivityDescribeOptions` that can be used to retrieve data associated with
+  activity execution, such as input and result.
+- More properties in `ActivityDescribeOptions`.
```
jmaeagle99: I think if we're going to be specific about what changed on `ActivityExecutionDescription`, then we should do the same here.

CHANGELOG.md
```diff
@@ -21,11 +21,17 @@ to docs, or any other relevant information.
 
 ### :boom: Breaking Changes
```
jmaeagle99, line 23: Also list `ActivityExecution` and `StartActivityOptions` breaking changes?

src/Temporalio/Client/StartActivityOptions.cs
```diff
@@ -89,8 +89,7 @@ public StartActivityOptions(string id, string taskQueue)
         /// Gets or sets a single-line fixed summary for this activity execution that may appear in
         /// UI/CLI. This can be in single-line Temporal markdown format.
         /// </summary>
-        /// <remarks>WARNING: This setting is experimental.</remarks>
```
jmaeagle99, line 92: Intentional dropping of the experimental warning?

src/Temporalio/Client/ActivityExecutionDescription.cs
```diff
@@ -13,162 +18,287 @@ namespace Temporalio.Client
...
+        /// <para>
+        /// Zero if the activity has not sent any heartbeats or if the server didn't report heartbeat count.
+        /// </para>
+        /// </summary>
+        public long TotalHeartbeatCount { get; }
+
+        /// <summary>
+        /// Gets the raw proto info.
+        /// </summary>
+        public new ActivityExecutionInfo RawInfo => RawResponse.Info;
```
jmaeagle99, line 233: What are we trying to enable by making select fields of the API public properties? Also, this shadowing of the existing `RawInfo` on the base class is going to lead to confusion.

src/Temporalio/Client/ActivityExecutionDescription.cs
```diff
@@ -13,162 +18,287 @@ namespace Temporalio.Client
...
         /// <summary>
         /// Gets the single-line fixed summary for this activity execution that may appear in
         /// UI/CLI. This can be in single-line Temporal markdown format.
         /// </summary>
-        /// <remarks>WARNING: Standalone activities are experimental.</remarks>
-        /// <returns>Static summary.</returns>
-#pragma warning disable VSTHRD003 // Awaiting our own lazily-created task
-        public async Task<string?> GetStaticSummaryAsync() =>
+        /// <returns>Activity summary.</returns>
+        public async Task<string?> GetSummaryAsync() =>
```
jmaeagle99, line 256: Why are we doing this rename? It's inconsistent with everything else in the SDK. And if we are sticking with the rename, why isn't `StaticDetails` also being renamed?

src/Temporalio/Client/ActivityExecutionDescription.cs
```diff
@@ -13,162 +18,287 @@ namespace Temporalio.Client
...
+        /// </summary>
+        /// <typeparam name="T">Type to convert the result into.</typeparam>
+        /// <exception cref="InvalidOperationException">If result is not available. See <see cref="HasResult"/>.</exception>
+        /// <remarks>
+        /// Activity result is only available if the activity has completed successfully.
+        /// <para>Always false if <see cref="ActivityDescribeOptions.IncludeOutcome"/> was false.</para>
+        /// </remarks>
+        /// <returns>Activity result.</returns>
+        /// <seealso cref="HasResult"/>
+        public async Task<T> GetResultAsync<T>()
```
jmaeagle99, line 311: This has divergent behavior compared to `ActivityHandle.GetResultAsync` for the failure cases. For example, I think `ActivityHandle.GetResultAsync` raises an `ActivityFailedException` when there was a failure. Is that intentional?

src/Temporalio/Client/ActivityExecutionDescription.cs
```diff
@@ -13,162 +18,287 @@ namespace Temporalio.Client
...
+
+        /// <summary>
+        /// Gets the raw proto input.
+        /// </summary>
+        public IReadOnlyCollection<Payload>? RawInput => RawResponse.Input?.Payloads_;
+
+        /// <summary>
+        /// Gets the raw proto outcome.
+        /// </summary>
+        public ActivityExecutionOutcome? RawOutcome => RawResponse.Outcome;
```
jmaeagle99, line 243: If this property stays, probably should be named `RawOutcomeFailure` or rename `GetOutcomeFailureAsync` to `GetOutcomeAsync`. As it is right now, they are asymmetric.

src/Temporalio/Client/ActivityExecutionDescription.cs
```diff
@@ -13,162 +18,287 @@ namespace Temporalio.Client
...
             (await userMetadata.Value.ConfigureAwait(false)).Details;
+
+        /// <summary>
+        /// Gets the failure from the last failed attempt, or null if not available.
+        /// </summary>
+        /// <remarks>Always false if <see cref="ActivityDescribeOptions.IncludeLastFailure"/> was false.</remarks>
+        /// <returns>Last failure, or null if not available.</returns>
+        /// <seealso cref="HasLastFailure"/>
+        public async Task<Exception?> GetLastFailureAsync() =>
+            await lastFailure.Value.ConfigureAwait(false);
```
jmaeagle99: I think this and `GetOutcomeFailureAsync` should throw if their corresponding options to include them were false. And this is not simply just checking the `Has*` properties. As they are right now, there's no way to distinguish "forgot to set `IncludeLastFailure = true`" vs "the activity doesn't have a failure". Something like (for LastFailure):

- `!options.IncludeLastFailure` -> throw `InvalidOperationException` stating that the last failure was not included and possibly how to remediate that.
- `!HasLastFailure` -> throw InvalidOperationException stating the activity not have an known failures (yet).

and then change the return type to `Task<Exception>` (to be non-nullable).

Then similarly for `GetOutcomeFailureAsync`

src/Temporalio/Client/ActivityDescribeOptions.cs
```diff
@@ -9,10 +9,32 @@ namespace Temporalio.Client
...
+        public bool IncludeInput { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value indicating whether to include outcome in the response if available.
+        /// </summary>
+        /// <seealso cref="ActivityExecutionDescription.HasResult"/>
+        /// <seealso cref="ActivityExecutionDescription.HasOutcomeFailure"/>
+        /// <seealso cref="ActivityExecutionDescription.GetResultAsync"/>
+        /// <seealso cref="ActivityExecutionDescription.GetOutcomeFailureAsync"/>
+        public bool IncludeOutcome { get; set; }
```
jmaeagle99, line 24: This controls both of the "outcome failure" and the "activity result"?

src/Temporalio/Client/ActivityExecutionDescription.cs
```diff
@@ -13,162 +18,287 @@ namespace Temporalio.Client
...
+        /// <exception cref="InvalidOperationException">If result is not available. See <see cref="HasResult"/>.</exception>
+        /// <remarks>
+        /// Activity result is only available if the activity has completed successfully.
+        /// <para>Always false if <see cref="ActivityDescribeOptions.IncludeOutcome"/> was false.</para>
+        /// </remarks>
+        /// <returns>Activity result.</returns>
+        /// <seealso cref="HasResult"/>
+        public async Task<T> GetResultAsync<T>()
+        {
+            if (!HasResult)
```
jmaeagle99, line 317: This check should be expanded and throw an error with a better message:
- `!options.IncludeOutcome` -> throw `InvalidOperationException` stating that the outcome was not included and possibly how to remediate that.
- `!HasResult` -> throw `InvalidOperationException` stating the activity hasn't closed yet.
- `HasOutcomeFailure` throw `ActivityFailedException` wrapping the converted cause.

src/Temporalio/Common/Priority.cs
```diff
@@ -33,7 +33,7 @@ public Priority(int? priorityKey = null, string? fairnessKey = null, float? fair
         /// Initializes a new instance of the <see cref="Priority"/> class from a proto.
         /// </summary>
         /// <param name="priority">The proto to initialize from.</param>
-        internal Priority(Temporalio.Api.Common.V1.Priority priority)
+        internal Priority(Temporalio.Api.Common.V1.Priority? priority)
```
jmaeagle99, line 36: Why was this widened?

jmaeagle99 (changes requested):
I left a lot of comments about the API behavior on `ActivityExecutionDescription`. It might be best to read through them all first because some of them nullify others depending on the decisions taken.

src/Temporalio/Client/ActivityDescribeOptions.cs
maciejdudko: Yes. In Proto API there's just one field Outcome.

src/Temporalio/Client/ActivityExecutionDescription.cs
maciejdudko: We want to provide access to raw Protobuf objects containing the description, but not to the whole raw response. Currently that means exposing `RawInfo` (which contains `HeartbeatDetails` and `LastFailure`), `RawInput` and `RawOutcome`.

Shadowing of `RawInfo` is intentional as the types are incompatible.

src/Temporalio/Client/ActivityExecutionDescription.cs
maciejdudko: `RawOutcome` contains both outcome result and outcome failure.

src/Temporalio/Client/ActivityExecutionDescription.cs
maciejdudko: It's consistent with Workflow Activity `ActivityOptions.Summary` which was supposed to be renamed to `StaticSummary`, but was stabilized before the rename happened and now we're stuck.

src/Temporalio/Client/ActivityExecutionDescription.cs
maciejdudko: It's possible Static Details remain experimental longer than the rest of Standalone Activities, and I want to prevent accidental stabilization.

src/Temporalio/Client/ActivityExecutionDescription.cs
maciejdudko: Yes, it's intentional. These APIs are for different use cases (waiting for execution vs. inspecting execution details).

src/Temporalio/Client/StartActivityOptions.cs
maciejdudko: Yes, user metadata is no longer experimental.

src/Temporalio/Common/Priority.cs
maciejdudko: Priority properties are non-nullable as a convention, Proto nulls are translated to default Priority instances. It's more convenient if this constructor accepts null.

CHANGELOG.md
maciejdudko: Note there was a typo here, the line is meant to say "More properties in `ActivityExecutionDescription`"

src/Temporalio/Client/ActivityExecutionDescription.cs
```diff
@@ -13,162 +18,307 @@ namespace Temporalio.Client
...
+        /// <seealso cref="HasResult"/>
+        public async Task<T> GetResultAsync<T>()
+        {
+            if (!requestOptions.IncludeOutcome)
+            {
+                throw new InvalidOperationException("ActivityDescribeOptions.IncludeOutcome must be set to true.");
+            }
+            if (!HasResult)
+            {
+                throw new InvalidOperationException("Result unavailable.");
```
jmaeagle99: This is thrown when the activity has closed with a failure. We should be checking if `HasOutcomeFailure` and throw a specific exception stating that a failure occurred.

src/Temporalio/Client/ActivityExecutionDescription.cs
```diff
@@ -13,162 +18,307 @@ namespace Temporalio.Client
...
+
+        /// <summary>
+        /// Gets the failure of the activity execution, or null if not available.
+        /// </summary>
+        /// <returns>Activity outcome failure, or null if not available.</returns>
+        /// <exception cref="InvalidOperationException">If <see cref="ActivityDescribeOptions.IncludeLastFailure"/> was false.</exception>
+        /// <remarks>Activity outcome failure is only available if the activity has closed with a failure.</remarks>
+        /// <seealso cref="GetResultAsync"/>
+        /// <seealso cref="HasOutcomeFailure"/>
+        public async Task<Exception?> GetOutcomeFailureAsync()
```
jmaeagle99: If we know there is no outcome, and we throw an exception for that, then we should type this without nullability. The backing `Lazy<Task<Exception>>` can be made non-nullable and should be faulted, not return null, if there is no outcome failure.

src/Temporalio/Client/ActivityExecutionDescription.cs
```diff
@@ -13,162 +18,307 @@ namespace Temporalio.Client
...
         public async Task<string?> GetStaticDetailsAsync() =>
             (await userMetadata.Value.ConfigureAwait(false)).Details;
+
+        /// <summary>
+        /// Gets the failure from the last failed attempt, or null if not available.
+        /// </summary>
+        /// <returns>Last failure, or null if not available.</returns>
+        /// <exception cref="InvalidOperationException">If <see cref="ActivityDescribeOptions.IncludeLastFailure"/> was false.</exception>
+        /// <seealso cref="HasLastFailure"/>
+        public async Task<Exception?> GetLastFailureAsync()
```
jmaeagle99: Same as https://github.com/temporalio/sdk-dotnet/pull/869/changes#r3936499680

src/Temporalio/Client/TemporalClient.Activity.cs
```diff
@@ -222,9 +222,13 @@ public override async Task<ActivityExecutionDescription> DescribeActivityAsync(
...
                         ActivityId = input.Id,
                         RunId = input.RunId ?? string.Empty,
+                        IncludeInput = input.Options?.IncludeInput ?? false,
+                        IncludeOutcome = input.Options?.IncludeOutcome ?? false,
+                        IncludeHeartbeatDetails = input.Options?.IncludeHeartbeatDetails ?? false,
+                        IncludeLastFailure = input.Options?.IncludeLastFailure ?? false,
                     },
                     DefaultRetryOptions(input.Options?.Rpc)).ConfigureAwait(false);
-                return new(resp, Client.Options.Namespace, Client.Options.DataConverter);
+                return new((ActivityDescribeOptions?)input.Options?.Clone(), resp, Client.Options.Namespace, Client.Options.DataConverter);
```
jmaeagle99: I would move the clone logic to inside the constructor. The caller shouldn't have to worry about cloning since it doesn't know how the `ActivityExecutionDescription` will use it. Make that the responsibility of `ActivityExecutionDescription`.

src/Temporalio/Client/ActivityExecutionDescription.cs
```diff
@@ -13,162 +18,307 @@ namespace Temporalio.Client
...
         /// <summary>
         /// Gets the schedule-to-start timeout.
         /// </summary>
-        public TimeSpan? ScheduleToStartTimeout { get; private init; }
+        public TimeSpan? ScheduleToStartTimeout { get; }
+
+        /// <summary>
+        /// Gets the time to wait before making the first activity task available for dispatch.
+        /// </summary>
+        public TimeSpan? StartDelay { get; }
```
jmaeagle99, line 211: These new properties do not have test coverage. While I don't expect some functional deficit, I think we should have a test that can validate "are these properties filled when expected".

src/Temporalio/Client/ActivityExecutionDescription.cs
maciejdudko: We throw exception when outcome was not requested, but return null when it was requested but unavailable. This is consistent with input, heartbeat details and last failure. Result is the odd one out because C# cannot express generic method returning `T?` for value- and reference-types at the same time.

src/Temporalio/Client/ActivityExecutionDescription.cs
maciejdudko: Setting up the required test scenarios is quite a bit of work, and I wouldn't want to delay this feature because of it. Are we fine creating a separate follow-up issue for these tests?

src/Temporalio/Client/ActivityExecutionDescription.cs
jmaeagle99: My argument hasn't been about the ability to express culpability. In fact, I stated in the comment I don't think it should be null.

An API called GetOutcomeFailureAsync should either return a failure or throw. It's not called TryGetOutcomeFailureAsync or MaybeGetOutcomeFailureAsync (both would be atypical in .NET).

My argument has been "user cannot distinguish between actively running activity vs one that completed without fault". Similar argument for LastError and for Result. So one might say "well user should check HasOutcomeFailure and then call GetOutcomeFailureAsync". If you are going to force that pattern, then you might as well make GetOutcomeFailureAsync return a non-nullable Exception so the user doesn't have to null check what they already know to not be null. Then they'll just get into a habit of using the null-forgiving operator. It just seems like a not good ergonomic experience.

src/Temporalio/Client/ActivityExecutionDescription.cs
jmaeagle99: Not sure I understand how it's complicated. This is just a unit test of making sure that creating an ActivityExecutionDescription sets its properties correctly based on the constructor arguments. But this can be deferred if necessary.

src/Temporalio/Client/ActivityExecutionDescription.cs
jmaeagle99: Discussed offline. This class is meant to be more of a snapshot rather than provide ergonomically usable API.

src/Temporalio/Client/ActivityExecutionDescription.cs
```diff
@@ -13,162 +18,289 @@ namespace Temporalio.Client
...
+        public async Task<Exception?> GetOutcomeFailureAsync() => await outcomeFailure.Value.ConfigureAwait(false);
 #pragma warning restore VSTHRD003
+
+        /// <summary>
+        /// Gets the result of the activity execution.
+        /// </summary>
+        /// <typeparam name="T">Type to convert the result into.</typeparam>
+        /// <returns>Activity result.</returns>
+        /// <exception cref="InvalidOperationException">If result is not available. See <see cref="HasResult"/>.</exception>
+        /// <exception cref="InvalidOperationException">If <see cref="ActivityDescribeOptions.IncludeOutcome"/> was false.</exception>
```
jmaeagle99: Remove this since it was reverted.

src/Temporalio/Client/ActivityExecutionDescription.cs
```diff
@@ -13,162 +18,289 @@ namespace Temporalio.Client
...
+        /// <remarks>
+        /// Activity result is only available if the activity has completed successfully.
+        /// <para>Always null if <see cref="ActivityDescribeOptions.IncludeOutcome"/> was false.</para>
+        /// </remarks>
+        /// <seealso cref="GetOutcomeFailureAsync"/>
+        /// <seealso cref="HasResult"/>
+        public async Task<T> GetResultAsync<T>()
```
jmaeagle99: Can we do something like:

```
public async Task<T?> GetResultAsync<T>()
{
    var result = RawOutcome?.Result;
    // Use default if the caller is ignoring the result or no payload is present
    if (typeof(T) == typeof(ValueTuple) || result == null || result.Payloads_.Count == 0)
    {
        return default;
    }
    return await dataConverter.ToSingleValueAsync<T>(result.Payloads_).ConfigureAwait(false);
}
```

jmaeagle99 (approved):

PR #880, opened 2026-09-01
Nexus: raw activity starts inherit request ID and links

tekkaya:
## What was changed

A raw activity start issued from inside a synchronous Nexus operation handler now inherits the handler's request ID and inbound links, same as the guarded path already does.

## Why?

Previously a raw activity client call made inside a Nexus handler was not carrying none of the Nexus-aware wiring: no inbound links, and no redelivery-safe request ID. This change aligns with the equivalent fixes in sdk-typescript (#2369), sdk-go (#2633), and sdk-java (#3048).

## Checklist

1. Closes N/A Aligns SDK behaviour for link and request id propagation in raw activity starts, see `Why` for related PRs in other SDKs.

2. How was this tested:
   - New tests added to `TemporalClientNexusOperationTests`
   - Full suite passes locally, including the inbound-link assertion against the pinned dev-server build.

3. Any docs updates needed?
   - No

### DIFF
--- a/src/Temporalio/Client/TemporalClient.Activity.cs
+++ b/src/Temporalio/Client/TemporalClient.Activity.cs
@@ -5,6 +5,7 @@ using System.Linq.Expressions;
 using System.Threading.Tasks;
 using Google.Protobuf;
 using Google.Protobuf.WellKnownTypes;
+using NexusRpc.Handlers;
 using Temporalio.Api.Common.V1;
 using Temporalio.Api.TaskQueue.V1;
 using Temporalio.Api.WorkflowService.V1;
@@ -96,6 +97,7 @@ namespace Temporalio.Client
                 {
                     throw new ArgumentException("StartDelay must be non-negative");
                 }
+                input = ApplyNexusWiringForRawActivityStart(input);
                 try
                 {
                     // Activity-specific data converter
@@ -304,6 +306,32 @@ namespace Temporalio.Client
                         .ToList()
                         .AsReadOnly(),
                     NextPageToken: resp.NextPageToken.IsEmpty ? null : resp.NextPageToken.ToByteArray());
+            }
+
+            // Give a raw start the same request ID and inbound links a guarded start already gets.
+            private static StartActivityInput ApplyNexusWiringForRawActivityStart(StartActivityInput input)
+            {
+                if (!NexusOperationExecutionContext.HasCurrent ||
+                    input.Options.RequestId != null ||
+                    NexusOperationExecutionContext.Current.HandlerContext is not
+                        OperationStartContext nexusStartContext)
+                {
+                    return input;
+                }
+                var options = (StartActivityOptions)input.Options.Clone();
+                var links = NexusOperationStartHelper.CreateInboundLinks(
+                    nexusStartContext, NexusOperationExecutionContext.Current);
+                if (links != null)
+                {
+                    options.Links = options.Links is { } existingLinks ?
+                        existingLinks.Concat(links).ToList() : links;
+                }
+                options.OnConflictOptions = NexusOperationStartHelper.CreateActivityOnConflictOptions(
+                    options.IdConflictPolicy,
+                    hasLinks: options.Links is { Count: > 0 },
+                    hasCompletionCallback: false);
+                options.RequestId = nexusStartContext.RequestId;
+                return input with { Options = options };
             }
 
 #if NETCOREAPP3_0_OR_GREATER
--- a/src/Temporalio/Nexus/NexusActivityStartHelper.cs
+++ b/src/Temporalio/Nexus/NexusActivityStartHelper.cs
@@ -42,25 +42,21 @@ namespace Temporalio.Nexus
             // Build the callback-header token without a run ID (we don't have it yet).
             var callbackToken = NexusActivityExecutionToken.Create(namespace_, activityId, runId: null);
 
-            if (options.IdConflictPolicy == Api.Enums.V1.ActivityIdConflictPolicy.UseExisting)
-            {
-                options.OnConflictOptions = new()
-                {
-                    AttachLinks = true,
-                    AttachCompletionCallbacks = true,
-                    AttachRequestId = true,
-                };
-            }
-            if (NexusOperationStartHelper.CreateInboundLinks(
-                    nexusStartContext, temporalContext) is { } links)
+            var links = NexusOperationStartHelper.CreateInboundLinks(nexusStartContext, temporalContext);
+            if (links != null)
             {
                 options.Links = links;
             }
-            if (NexusOperationStartHelper.CreateCallback(
-                    nexusStartContext, callbackToken, options.Links) is { } callback)
+            var callback = NexusOperationStartHelper.CreateCallback(
+                nexusStartContext, callbackToken, options.Links);
+            if (callback != null)
             {
                 options.CompletionCallbacks = new[] { callback };
             }
+            options.OnConflictOptions = NexusOperationStartHelper.CreateActivityOnConflictOptions(
+                options.IdConflictPolicy,
+                hasLinks: options.Links is { Count: > 0 },
+                hasCompletionCallback: callback != null);
             options.RequestId = nexusStartContext.RequestId;
 
             // Do the start call
--- a/src/Temporalio/Nexus/NexusOperationStartHelper.cs
+++ b/src/Temporalio/Nexus/NexusOperationStartHelper.cs
@@ -91,5 +91,34 @@ namespace Temporalio.Nexus
             }
             return callback;
         }
+
+        /// <summary>
+        /// Build on-conflict options for an activity start under a <c>UseExisting</c> ID conflict
+        /// policy, attaching only the artifacts the start actually carries. The server rejects
+        /// attaching a request ID with no accompanying link or completion callback, so this
+        /// returns <c>null</c> unless there is at least one to attach.
+        /// </summary>
+        /// <param name="idConflictPolicy">The activity ID conflict policy.</param>
+        /// <param name="hasLinks">Whether the start carries links to attach.</param>
+        /// <param name="hasCompletionCallback">Whether the start carries a completion callback to
+        /// attach.</param>
+        /// <returns>On-conflict options, or <c>null</c> if not applicable.</returns>
+        internal static Api.Common.V1.OnConflictOptions? CreateActivityOnConflictOptions(
+            Api.Enums.V1.ActivityIdConflictPolicy idConflictPolicy,
+            bool hasLinks,
+            bool hasCompletionCallback)
+        {
+            if (idConflictPolicy != Api.Enums.V1.ActivityIdConflictPolicy.UseExisting ||
+                (!hasLinks && !hasCompletionCallback))
+            {
+                return null;
+            }
+            return new()
+            {
+                AttachLinks = hasLinks,
+                AttachCompletionCallbacks = hasCompletionCallback,
+                AttachRequestId = true,
+            };
+        }
     }
 }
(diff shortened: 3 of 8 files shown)


src/Temporalio/Worker/NexusWorker.cs
```diff
@@ -46,6 +47,7 @@ internal class NexusWorker
         private readonly ILogger logger;
         private readonly Handler handler;
         private readonly ConcurrentDictionary<ByteString, RunningTask> runningTasks = new();
+        private readonly Lazy<ITemporalClient?> nexusExecutionClient;
```
jmaeagle99: Can't cache the client because the `TemporalWorker` allows swapping the client: https://github.com/temporalio/sdk-dotnet/blob/1.18.0/src/Temporalio/Worker/TemporalWorker.cs

src/Temporalio/Worker/NexusWorker.cs
```diff
@@ -67,6 +69,13 @@ public NexusWorker(TemporalWorker worker)
                 worker.Options.NexusServices,
                 new NexusPayloadSerializer(worker.Client.Options.DataConverter),
                 middleware);
+
+            nexusExecutionClient = new(() =>
```
jmaeagle99, line 73: Can't cache the client because the `TemporalWorker` allows swapping the client: https://github.com/temporalio/sdk-dotnet/blob/1.18.0/src/Temporalio/Worker/TemporalWorker.cs

Maybe can cache, but have to update the cache each time the public API gets invoked. Possibly require locking.

src/Temporalio/Nexus/NexusActivityStartInterceptor.cs
```diff
@@ -0,0 +1,57 @@
...
+                if (links != null &&
+                    options.IdConflictPolicy == Api.Enums.V1.ActivityIdConflictPolicy.UseExisting)
+                {
+                    options.OnConflictOptions = new()
+                    {
+                        AttachLinks = true,
+                        AttachRequestId = true,
+                    };
+                }
+                options.RequestId = nexusStartContext.RequestId;
```
jmaeagle99, line 51:
```suggestion
                options.RequestId = string.IsNullOrEmpty(nexusStartContext.RequestId) ? null : nexusStartContext.RequestId;
```

jmaeagle99: Or, better yet, normalize this in `NexusWorker.HandleStartOperationAsync` if possible so downstream doesn't have to constantly convert it.

jmaeagle99: This also might be an issue with the existing code too: https://github.com/temporalio/sdk-dotnet/blob/1.18.0/src/Temporalio/Nexus/NexusActivityStartHelper.cs#L64

src/Temporalio/Nexus/NexusActivityStartInterceptor.cs
```diff
@@ -0,0 +1,57 @@
...
+        /// Initializes a new instance of the <see cref="NexusActivityStartInterceptor"/> class.
+        /// </summary>
+        /// <param name="next">Next interceptor in the chain.</param>
+        internal NexusActivityStartInterceptor(ClientOutboundInterceptor next)
+            : base(next)
+        {
+        }
+
+        /// <inheritdoc />
+        public override Task<ActivityHandle<TResult>> StartActivityAsync<TResult>(
```
jmaeagle99, line 24: This looks like a very large overlap with https://github.com/temporalio/sdk-dotnet/blob/1.18.0/src/Temporalio/Nexus/NexusActivityStartHelper.cs#L27. Bugs in one are likely to be missed in the other. Can we consolidate these? Maybe much of what occurs in that start helper can be removed an moved to here?

src/Temporalio/Nexus/NexusActivityStartInterceptor.cs
```diff
@@ -0,0 +1,57 @@
...
+                NexusOperationExecutionContext.Current.HandlerContext is
+                    OperationStartContext nexusStartContext)
+            {
+                var nexusExecutionContext = NexusOperationExecutionContext.Current;
+                var options = (StartActivityOptions)input.Options.Clone();
+                var links = NexusOperationStartHelper.CreateInboundLinks(
+                    nexusStartContext, nexusExecutionContext);
+                if (links != null)
+                {
+                    options.Links = links;
```
jmaeagle99, line 38: Shouldn't this append?

src/Temporalio/Worker/NexusWorker.cs
```diff
@@ -67,6 +69,13 @@ public NexusWorker(TemporalWorker worker)
...
+            {
+                var client = worker.Client as ITemporalClient;
+                return client is TemporalClient temporalClient ?
```
jmaeagle99, line 76: This gives up if the client is not a `TemporalClient` (it is a custom client). Would it be possible to wrap it with a decorator and supply the necessary information via ambient context? Could be future improvement.

src/Temporalio/Nexus/NexusActivityStartInterceptor.cs
```diff
@@ -0,0 +1,57 @@
...
+using Temporalio.Client;
+using Temporalio.Client.Interceptors;
+
+namespace Temporalio.Nexus
+{
+    /// <summary>
+    /// Gives a raw activity start issued from inside a Nexus start handler the same request ID
+    /// and inbound links a guarded start already gets.
+    /// </summary>
+    internal sealed class NexusActivityStartInterceptor : ClientOutboundInterceptor
```
jmaeagle99, line 12: Would it make sense to just apply this interceptor within the TemporalClient impl unconditionally and not have to attempt to recreate a new client? It would effectively be a passthrough if not used within the context of Nexus handler, correct?

tests/Temporalio.Tests/Client/TemporalClientNexusOperationTests.cs
```diff
@@ -141,6 +144,255 @@ public IOperationHandler<string, NoValue> NoResult() =>
...
+                }
+                return result;
+            });
+
+        [NexusOperationHandler]
+        public IOperationHandler<string, NoValue> NoResult() =>
+            OperationHandler.Sync<string, NoValue>(
+                (Func<OperationStartContext, string, NoValue>)((ctx, input) => default!));
+    }
+
```
jmaeagle99, line 253: Most of the test additions are about checking that the Nexus worker properly wires the interceptor in correctly and never touch the standalone Nexus client. This test class is also excluded from cloud testing at this time with an unmatched reason. These should probably be moved to NexusWorkerTests, which has a bunch of reusable fixtures, helpers, workflows, activities, etc already.

tekkaya:
Tried to address everything except the future improvement suggestion here: https://github.com/temporalio/sdk-dotnet/pull/880#discussion_r3972861386 ptal

src/Temporalio/Worker/NexusWorker.cs
```diff
@@ -169,6 +169,11 @@ private async Task<StartOperationResponse> HandleStartOperationAsync(
             // Create context
             RemoveInvalidHeaders(task.Request.Header);
             var startOp = task.Request.StartOperation;
+            if (string.IsNullOrEmpty(startOp.RequestId))
+            {
+                throw new HandlerException(
```
jmaeagle99, line 174: This is refusing a Nexus operation because the server didn't supply a request ID rather than repairing it by generating one in place. We cannot refuse at this point. The Go SDK was updated to repair the value: https://github.com/temporalio/sdk-go/pull/2633/changes#diff-8f52a328ae9cd7b145415b64f49a6cca248b367702ea4c011cccd1f39229921eR217

tests/Temporalio.Tests/Nexus/NexusOperationExecutionContextTests.cs
```diff
@@ -28,6 +28,13 @@ public void RequestLinks_RoundTrips()
         Assert.Equal(links, context.RequestLinks);
     }
 
+    [Fact]
+    public void HandlerContext_IsOperationStartContext()
```
jmaeagle99, line 32: I don't think this test adds value. Please remove.

src/Temporalio/Nexus/NexusActivityStartHelper.cs
```diff
@@ -42,25 +42,21 @@ internal static async Task<string> StartActivityAsync(
...
-                options.OnConflictOptions = new()
-                {
-                    AttachLinks = true,
-                    AttachCompletionCallbacks = true,
-                    AttachRequestId = true,
-                };
-            }
-            if (NexusOperationStartHelper.CreateInboundLinks(
-                    nexusStartContext, temporalContext) is { } links)
+            var links = NexusOperationStartHelper.CreateInboundLinks(nexusStartContext, temporalContext);
```
jmaeagle99, line 45: I believe that `NexusWorker` is already doing this at https://github.com/temporalio/sdk-dotnet/blob/main/src/Temporalio/Worker/NexusWorker.cs#L208, so I think you can just get them from `temporalContext.RequestLinks` and check if the count is greater than 0.

src/Temporalio/Nexus/NexusActivityStartHelper.cs
```diff
@@ -42,25 +42,21 @@ internal static async Task<string> StartActivityAsync(
...
+            if (links != null)
             {
                 options.Links = links;
```
jmaeagle99, line 48: Same as https://github.com/temporalio/sdk-dotnet/pull/880#discussion_r3972795161, I think this should append instead of overwrite.

src/Temporalio/Nexus/NexusOperationStartHelper.cs
```diff
@@ -91,5 +91,34 @@ internal static class NexusOperationStartHelper
...
+        /// policy, attaching only the artifacts the start actually carries. The server rejects
+        /// attaching a request ID with no accompanying link or completion callback, so this
+        /// returns <c>null</c> unless there is at least one to attach.
+        /// </summary>
+        /// <param name="idConflictPolicy">The activity ID conflict policy.</param>
+        /// <param name="hasLinks">Whether the start carries links to attach.</param>
+        /// <param name="hasCompletionCallback">Whether the start carries a completion callback to
+        /// attach.</param>
+        /// <returns>On-conflict options, or <c>null</c> if not applicable.</returns>
+        internal static Api.Common.V1.OnConflictOptions? CreateActivityOnConflictOptions(
```
jmaeagle99, line 106: I would prefer we don't over-qualify the namespaces. Add a `using` at the top and remove all full qualification unless absolutely necessary.

jmaeagle99, line 106: Given this is for activities, it probably should be moved to `NexusActivityStartHelper`.

src/Temporalio/Client/TemporalClient.Activity.cs
```diff
@@ -306,6 +308,32 @@ public override async Task<ActivityListPage> ListActivitiesPaginatedAsync(
...
+            {
+                if (!NexusOperationExecutionContext.HasCurrent ||
+                    input.Options.RequestId != null ||
+                    NexusOperationExecutionContext.Current.HandlerContext is not
+                        OperationStartContext nexusStartContext)
+                {
+                    return input;
+                }
+                var options = (StartActivityOptions)input.Options.Clone();
+                var links = NexusOperationStartHelper.CreateInboundLinks(
```
jmaeagle99, line 322: This can use `NexusOperationExecutionContext.Current.RequestLinks` instead of recomputing. In fact, I think all sites that use `CreateInboundLinks` and use the execution context instead and remove this helper method.

PR #884, opened 2026-09-04
Add experimental Workflow Streams extension

brianstrauch:
## Summary

- add an experimental netstandard2.0 Workflow Streams package with durable, offset-based, multi-topic publish/subscribe
- implement the cross-SDK wire protocol, deterministic workflow host, publisher state machine, reusable subscriptions, and exact reserved-handler allowlisting
- add independent tests, package documentation, solution/DocFX/NuGet integration, ownership, and changelog entries

This implementation was derived from the official TypeScript, Go, and Java Workflow Streams protocol.

## Validation

- focused Workflow Streams and reserved-handler tests: 15 passed
- dotnet format --verify-no-changes: passed
- dotnet pack -c Debug: passed
- mise run docs: passed with 0 warnings
- full dotnet test: 801 passed, 13 skipped, 1 unrelated reproducible failure in WorkflowWorkerTests.ExecuteWorkflowAsync_LocalActivityIncludeArgumentsInMarker_OnlyRecordsWhenEnabled

### DIFF
--- a/src/Temporalio.Extensions.WorkflowStreams/.editorconfig
+++ b/src/Temporalio.Extensions.WorkflowStreams/.editorconfig
@@ -1,0 +1,13 @@
+[*.workflow.cs]
+
+# The protocol's established public name intentionally describes a stream.
+dotnet_diagnostic.CA1711.severity = none
+
+# Workflow continuations must remain on Temporal's deterministic task scheduler.
+dotnet_diagnostic.CA2007.severity = none
+
+# Workflow tasks intentionally use the current deterministic scheduler.
+dotnet_diagnostic.CA2008.severity = none
+
+# Workflow continuations intentionally use the current deterministic scheduler.
+dotnet_diagnostic.VSTHRD105.severity = none
--- a/src/Temporalio.Extensions.WorkflowStreams/FlushTimeoutException.cs
+++ b/src/Temporalio.Extensions.WorkflowStreams/FlushTimeoutException.cs
@@ -1,0 +1,21 @@
+using System;
+
+namespace Temporalio.Extensions.WorkflowStreams
+{
+    /// <summary>
+    /// Exception thrown when an ambiguously delivered publish batch exceeds its retry duration.
+    /// </summary>
+    /// <remarks>
+    /// The batch is dropped locally. It is present in the workflow log if the signal reached the
+    /// server, and otherwise is lost. WARNING: Workflow Streams is experimental and may change.
+    /// </remarks>
+    public class FlushTimeoutException : Exception
+    {
+        /// <summary>Initializes a new instance of the <see cref="FlushTimeoutException"/> class.</summary>
+        /// <param name="message">Description of the expired ambiguous-delivery window.</param>
+        internal FlushTimeoutException(string message)
+            : base(message)
+        {
+        }
+    }
+}
--- a/src/Temporalio.Extensions.WorkflowStreams/PollInput.cs
+++ b/src/Temporalio.Extensions.WorkflowStreams/PollInput.cs
@@ -1,0 +1,27 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text.Json.Serialization;
+
+namespace Temporalio.Extensions.WorkflowStreams
+{
+    /// <summary>The cross-language update input for polling stream items.</summary>
+    /// <remarks>WARNING: Workflow Streams is experimental and may change.</remarks>
+    public sealed class PollInput
+    {
+        private IReadOnlyCollection<string> topics = Array.Empty<string>();
+
+        /// <summary>Gets or sets the topic filter. Empty means all topics.</summary>
+        [JsonPropertyName("topics")]
+        public IReadOnlyCollection<string> Topics
+        {
+            get => topics;
+            set => topics = value?.Select(topic => topic ?? string.Empty).ToArray() ??
+                Array.Empty<string>();
+        }
+
+        /// <summary>Gets or sets the global offset at which polling begins.</summary>
+        [JsonPropertyName("from_offset")]
+        public long FromOffset { get; set; }
+    }
+}
--- a/src/Temporalio.Extensions.WorkflowStreams/PublishEntry.cs
+++ b/src/Temporalio.Extensions.WorkflowStreams/PublishEntry.cs
@@ -1,0 +1,28 @@
+using System.Text.Json.Serialization;
+
+namespace Temporalio.Extensions.WorkflowStreams
+{
+    /// <summary>A single item in a publish signal batch.</summary>
+    /// <remarks>WARNING: Workflow Streams is experimental and may change.</remarks>
+    public sealed class PublishEntry
+    {
+        private string topic = string.Empty;
+        private string data = string.Empty;
+
+        /// <summary>Gets or sets the topic. A wire-level null is normalized to an empty string.</summary>
+        [JsonPropertyName("topic")]
+        public string Topic
+        {
+            get => topic;
+            set => topic = value ?? string.Empty;
+        }
+
+        /// <summary>Gets or sets the base64-encoded serialized Temporal payload.</summary>
+        [JsonPropertyName("data")]
+        public string Data
+        {
+            get => data;
+            set => data = value ?? string.Empty;
+        }
+    }
+}
--- a/src/Temporalio.Extensions.WorkflowStreams/README.md
+++ b/src/Temporalio.Extensions.WorkflowStreams/README.md
@@ -1,0 +1,173 @@
+# Temporal .NET Workflow Streams
+
+> [!WARNING]
+> Workflow Streams is experimental. Its API and wire protocol may change before it is declared
+> stable.
+
+`Temporalio.Extensions.WorkflowStreams` provides a durable, offset-addressed, multi-topic log
+hosted by a Temporal Workflow. External publishers append batches with a Signal, subscribers
+long-poll with an Update, and a Query reports the current global offset. The implementation includes
+publisher deduplication, topic filtering, truncation, response paging, and continue-as-new state
+handoff.
+
+This is intended for durable progress, event, and incremental-result streams. Each poll is a
+Temporal Update round trip, so it is not intended for ultra-low-latency media or token streaming.
+
+## Install
+
+```shell
+dotnet add package Temporalio.Extensions.WorkflowStreams
+```
+
+The package targets `netstandard2.0`.
+
+## Host a stream in a Workflow
+
+Create the stream while the Workflow instance is constructed. This ensures its dynamically
+registered Signal, Update, and Query handlers exist before the first handler can be dispatched,
+including on a continue-as-new successor run.
+
+```csharp
+using Temporalio.Extensions.WorkflowStreams;
+using Temporalio.Workflows;
+
+public record OrderInput(int CompletedSteps, WorkflowStreamState? StreamState = null);
+
+[Workflow]
+public class OrderWorkflow
+{
+    private readonly WorkflowStream stream;
+    private bool finished;
+
+    [WorkflowInit]
+    public OrderWorkflow(OrderInput input)
+    {
+        stream = new(input.StreamState);
+    }
+
+    [WorkflowRun]
+    public async Task RunAsync(OrderInput input)
+    {
+        stream.Topic("status").Publish(new { State = "started" });
+
+        await Workflow.WaitConditionAsync(() => finished || Workflow.ContinueAsNewSuggested);
+        if (Workflow.ContinueAsNewSuggested)
+        {
+            await stream.ContinueAsNewAsync(state =>
+                Workflow.CreateContinueAsNewException(
+                    (OrderWorkflow workflow) => workflow.RunAsync(
+                        input with { StreamState = state })));
+        }
+    }
+
+    [WorkflowSignal]
+    public Task FinishAsync()
+    {
+        finished = true;
+        return Task.CompletedTask;
+    }
+}
+```
+
+`ContinueAsNewAsync` first detaches admitted pollers, waits until all handlers finish, and only then
+captures `WorkflowStreamState` and invokes the callback. Thread that state through the next run's
+input as shown above. For custom continue-as-new options, perform the same sequence explicitly:
+call `DetachPollers()`, wait for `Workflow.AllHandlersFinished`, call `GetState()`, and create the
+continue-as-new exception yourself.
+
+The log is retained until the Workflow calls `Truncate(offset)`. Offsets are global across every
+topic. A subscriber that falls behind truncation automatically resumes at the beginning of the
+retained log.
+
+## Publish from a client or Activity
+
+Use one asynchronously disposed client per target Workflow ID. Values are converted to Temporal
+`Payload`s when `Publish` is called, then buffered until the two-second interval, the configured
+batch size, a force flush, an explicit flush, or asynchronous disposal.
+
+```csharp
+await using var streams = new WorkflowStreamClient(temporalClient, workflowId);
+var status = streams.Topic("status");
+
+status.Publish(new { State = "working" });
+status.Publish(new { State = "done" }, forceFlush: true);
+await streams.FlushAsync(cancellationToken);
+```
+
+`FlushAsync` is a barrier for everything buffered before the call. Failed or timed-out Signal RPCs
+retain the same publisher ID and sequence for retry, allowing the Workflow to deduplicate ambiguous
+delivery. If the retry window expires, `FlushTimeoutException` is thrown and that ambiguous batch is
+dropped locally: it may already be present in the Workflow log, or it may be lost. Later batches use
+a new sequence.
+
+Inside an Activity, `FromActivity` obtains the Temporal client, parent Workflow ID, and payload
+converter from the current Activity context:
+
+```csharp
+[Activity]
+public async Task ReportAsync(IEnumerable<Progress> values)
+{
+    await using var streams = WorkflowStreamClient.FromActivity();
+    foreach (var value in values)
+    {
+        streams.Topic("progress").Publish(value);
+    }
+}
+```
+
+Always use `await using` or call `DisposeAsync` so the final buffer is drained and publisher resources
+are released. Publication after disposal throws `ObjectDisposedException`.
+
+## Subscribe
+
+Subscriptions yield raw Temporal `Payload`s so consumers can choose a result type per topic. The
+returned `IAsyncEnumerable<WorkflowStreamItem>` is reusable: each enumeration starts with its own
+offset and polling state.
+
+```csharp
+var subscription = streams.SubscribeAsync(new()
+{
+    Topics = new[] { "status", "progress" },
+    FromOffset = 0,
+});
+
+await foreach (var item in subscription.WithCancellation(cancellationToken))
+{
+    var value = matchingPayloadConverter.ToValue(item.Payload, typeof(MyEvent));
+    Console.WriteLine($"{item.Offset} {item.Topic}: {value}");
+}
+```
+
+For one topic, use `streams.Topic("status").SubscribeAsync(fromOffset)`. An empty topic collection
+subscribes to every topic; the empty string is the cross-SDK no-topic value. Consumer cancellation
+cancels the in-flight RPC and throws `OperationCanceledException`. Disposing the owning client ends
+its active enumerations cleanly. Enumerations also end cleanly when the Workflow reaches a terminal
+state and automatically follow continue-as-new chains.
+
+## Data conversion and interoperability
+
+The fixed protocol handlers are:
+
+- Signal `__temporal_workflow_stream_publish`
+- Update `__temporal_workflow_stream_poll`
+- Query `__temporal_workflow_stream_offset`
+
+The public wire DTOs use the protocol's exact snake-case JSON names. Each item's `data` is standard
+padded base64 containing a serialized `temporal.api.common.v1.Payload` protobuf. This preserves its
+encoding metadata and makes .NET publishers, Workflow hosts, and subscribers interoperable with the
+official Workflow Streams implementations in other Temporal SDKs.
+
+Only payload conversion is applied to an individual item. A client's payload codec chain, if any,
+is applied once to the Signal or Update envelope rather than once per item, avoiding double encoding.
+The envelope itself must use JSON-compatible conversion for cross-language interoperability.
+
+## Operational limits
+
+- Every waiting subscription uses an admitted Workflow Update. Account for concurrent and total
+  Update limits when choosing subscriber counts and continue-as-new frequency.
+- Poll results are paged at an estimated one megabyte and immediately repolled while another page is
+  ready. An individual item must fit in one page.
+- The Workflow log has no automatic retention policy. Truncate items once all required consumers have
+  advanced, and carry only the retained state through continue-as-new.
+- Publishing uses Signals, so malformed or oversized externally supplied wire entries cannot return
+  errors to their sender. The Workflow skips them and emits a replay-safe warning.
--- a/src/Temporalio.Extensions.WorkflowStreams/StreamPublisher.cs
+++ b/src/Temporalio.Extensions.WorkflowStreams/StreamPublisher.cs
@@ -1,0 +1,357 @@
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Threading;
+using System.Threading.Tasks;
+using Temporalio.Api.Common.V1;
+using Temporalio.Converters;
+using Temporalio.Exceptions;
+
+namespace Temporalio.Extensions.WorkflowStreams
+{
+    /// <summary>Owns serialized client publication state and its single timer.</summary>
+    internal sealed class StreamPublisher : IAsyncDisposable
+    {
+        private readonly object stateLock = new();
+        private readonly Func<PublishInput, CancellationToken, Task> signalAsync;
+        private readonly IPayloadConverter payloadConverter;
+        private readonly TimeSpan batchInterval;
+        private readonly TimeSpan maxRetryDuration;
+        private readonly int maxBatchSize;
+        private readonly string publisherId = Guid.NewGuid().ToString("N").Substring(0, 16);
+        private readonly SemaphoreSlim flushGate = new(1, 1);
+        private readonly SemaphoreSlim wakeSignal = new(0, 1);
+        private readonly CancellationTokenSource stopSource = new();
+        private List<PublishEntry> buffer = new();
+        private IReadOnlyCollection<PublishEntry>? pending;
+        private long sequence;
+        private long pendingSequence;
+        private long pendingStartedAt;
+        private Timer? timer;
+        private Task? backgroundTask;
+        private Task? disposeTask;
+        private bool disposing;
+        private FlushTimeoutException? deferredError;
+
+        /// <summary>Initializes a new instance of the <see cref="StreamPublisher"/> class.</summary>
+        /// <param name="signalAsync">Injected workflow signal operation.</param>
+        /// <param name="payloadConverter">Converter applied synchronously during publication.</param>
+        /// <param name="options">Snapshotted client options.</param>
+        internal StreamPublisher(
+            Func<PublishInput, CancellationToken, Task> signalAsync,
+            IPayloadConverter payloadConverter,
+            WorkflowStreamClientOptions options)
+        {
+            this.signalAsync = signalAsync;
+            this.payloadConverter = payloadConverter;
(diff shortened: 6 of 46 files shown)


src/Temporalio.Extensions.WorkflowStreams/WorkflowStream.workflow.cs
```diff
@@ -0,0 +1,374 @@
...
+        internal void Publish(string topic, object? value)
+        {
+            var payload = value as Payload ?? Workflow.PayloadConverter.ToPayload(value);
+            var encoded = PayloadWire.Encode(payload);
+            if (PayloadWire.EstimateSize(encoded, topic) >
+                WorkflowStreamConstants.MaxPollResponseBytes)
+            {
+                throw new ApplicationFailureException(
+                    "Workflow Stream item is too large to fit in a poll response",
+                    WorkflowStreamConstants.OversizedItemErrorType,
```
jmaeagle99: This error type does not existing in any of the other workflow streams implementations.

src/Temporalio.Extensions.WorkflowStreams/WorkflowStream.workflow.cs
```diff
@@ -0,0 +1,374 @@
...
+            log.Add(new(topic, payload));
+        }
+
+        private void Restore(WorkflowStreamState? state)
+        {
+            static ApplicationFailureException InvalidState(string message, Exception? inner) =>
+                new(
+                    message,
+                    inner,
+                    WorkflowStreamConstants.InvalidStateErrorType,
```
jmaeagle99: This error type does not existing in any of the other workflow streams implementations.

src/Temporalio.Extensions.WorkflowStreams/WorkflowStreamConstants.cs
```diff
@@ -0,0 +1,51 @@
...
+        public const string TruncatedOffsetErrorType = "TruncatedOffset";
+
+        /// <summary>Failure type returned when truncation is requested past the end of the log.</summary>
+        public const string TruncateOutOfRangeErrorType = "TruncateOutOfRange";
+
+        /// <summary>Failure type returned while pollers are detaching for continue-as-new.</summary>
+        public const string StreamDrainingErrorType = "StreamDraining";
+
+        /// <summary>Keeps each update result comfortably below server payload limits.</summary>
+        internal const int MaxPollResponseBytes = 1_000_000;
```
jmaeagle99: Does this guard exist in the other workflow streams implementations?

src/Temporalio.Extensions.WorkflowStreams/WorkflowStream.workflow.cs
```diff
@@ -0,0 +1,374 @@
...
+        }
+
+        /// <summary>Keeps workflow-side publications on the same converter and log path.</summary>
+        /// <param name="topic">Normalized topic name.</param>
+        /// <param name="value">Value or raw payload to append.</param>
+        internal void Publish(string topic, object? value)
+        {
+            var payload = value as Payload ?? Workflow.PayloadConverter.ToPayload(value);
+            var encoded = PayloadWire.Encode(payload);
+            if (PayloadWire.EstimateSize(encoded, topic) >
```
jmaeagle99: This is going to cause a Workflow Failure (not just a WFT Failure) because of the size validation. I don't see this in other SDK implementations, such as Python: https://github.com/temporalio/sdk-python/blob/1.32.0/temporalio/contrib/workflow_streams/_stream.py#L173 There are other uses of this size guard that also don't exist in the other implementations.

src/Temporalio.Extensions.WorkflowStreams/README.md
```diff
@@ -0,0 +1,173 @@
...
+
+The log is retained until the Workflow calls `Truncate(offset)`. Offsets are global across every
+topic. A subscriber that falls behind truncation automatically resumes at the beginning of the
+retained log.
+
+## Publish from a client or Activity
+
+Use one asynchronously disposed client per target Workflow ID. Values are converted to Temporal
+`Payload`s when `Publish` is called, then buffered until the two-second interval, the configured
+batch size, a force flush, an explicit flush, or asynchronous disposal.
```
jmaeagle99: How are "a force flush" and "an explicit flush" different?

src/Temporalio.Extensions.WorkflowStreams/WorkflowStreamClient.cs
```diff
@@ -0,0 +1,505 @@
...
+
+        /// <summary>Creates a stream client targeting the current activity's parent workflow.</summary>
+        /// <param name="options">Client options.</param>
+        /// <returns>A client for the activity's parent workflow.</returns>
+        public static WorkflowStreamClient FromActivity(WorkflowStreamClientOptions? options = null)
+        {
+            var context = ActivityExecutionContext.Current;
+            var id = context.Info.WorkflowId ?? throw new InvalidOperationException(
+                "WorkflowStreamClient.FromActivity requires a workflow activity");
+            return new(context.TemporalClient, id, options, context.PayloadConverter);
```
jmaeagle99, line 101: This is passing a payload converter with a serialization context bound to this specific activity. Received will then receive payloads that are _possibly_ serialized with that context information. But they are going to have different context information. So if the payload converter requires matching serialization information, it is likely to throw exceptions. The built-in payload converters don't require this, but `ISerializationContext` is applied to payload converters to enable some scenarios. Not sure this can be fixed from a library standpoint; probably worth documenting that customers shouldn't use payload converters that use serialization context for deserialization in combination with this library.

src/Temporalio.Extensions.WorkflowStreams/WorkflowStream.cs
```diff
@@ -0,0 +1,374 @@
+using System;
```
jmaeagle99, line 1: Why is this file named `WorkflowStream.workflow.cs` (questioning the `.workflow` portion).

tests/Temporalio.Tests/Extensions/WorkflowStreams/WorkflowStreamTests.cs
```diff
@@ -0,0 +1,286 @@
+namespace Temporalio.Tests.Extensions.WorkflowStreams;
+
+using Temporalio.Converters;
+using Temporalio.Extensions.WorkflowStreams;
+using Temporalio.Worker;
+using Temporalio.Workflows;
+using Xunit;
+using Xunit.Abstractions;
+
+public class WorkflowStreamTests : WorkflowEnvironmentTestBase
```
jmaeagle99, line 9: None of these tests are testing `WorkflowStream` (at least in a direct manner), they are testing `WorkflowStreamClient`. Suggest renaming to `WorkflowStreamClientTests`.

jmaeagle99, line 9: There should be some tests for `WorkflowStream` itself, at least as demonstrated in the README.md.

src/Temporalio.Extensions.WorkflowStreams/WorkflowStreamClientOptions.cs
```diff
@@ -0,0 +1,28 @@
...
+    /// <remarks>WARNING: Workflow Streams is experimental and may change.</remarks>
+    public class WorkflowStreamClientOptions : ICloneable
+    {
+        /// <summary>Gets or sets the interval between automatic flush attempts.</summary>
+        public TimeSpan BatchInterval { get; set; } = WorkflowStreamConstants.DefaultBatchInterval;
+
+        /// <summary>
+        /// Gets or sets the item count that triggers a flush. Zero disables size-based flushing.
+        /// </summary>
+        public int MaxBatchSize { get; set; }
```
jmaeagle99, line 15: I don't there are tests for this.

src/Temporalio.Extensions.WorkflowStreams/TopicHandle.cs
```diff
@@ -0,0 +1,39 @@
+using System.Collections.Generic;
+
+namespace Temporalio.Extensions.WorkflowStreams
+{
+    /// <summary>A client-side handle bound to one topic.</summary>
+    /// <remarks>WARNING: Workflow Streams is experimental and may change.</remarks>
+    public sealed class TopicHandle
```
jmaeagle99: I would suggest making the topic handle classes and stream items more strongly typed so customers don't have to deal with weak typing and raw `Payload`s. If you need a weakly typed variant, that's okay; .NET supports both generic and non-generic classes with the same name e.g. `class TopicHandle` and `class TopicHandle<T>` can both coexist.

src/Temporalio.Extensions.WorkflowStreams/WorkflowStreamConstants.cs
```diff
@@ -0,0 +1,51 @@
+namespace Temporalio.Extensions.WorkflowStreams
+{
+    /// <summary>
+    /// Fixed handler names and application-failure types in the Workflow Streams wire protocol.
+    /// </summary>
+    /// <remarks>WARNING: Workflow Streams is experimental and may change.</remarks>
+    public static class WorkflowStreamConstants
```
jmaeagle99: Why public? There looks like internal details.

src/Temporalio.Extensions.WorkflowStreams/WorkflowStreamConstants.cs
```diff
@@ -0,0 +1,51 @@
...
+    {
+        /// <summary>Signal used to append an external publisher's batch.</summary>
+        public const string PublishSignalName = "__temporal_workflow_stream_publish";
```
jmaeagle99: These are duplicated from `TemporalRuntime`. Probably should either reference from there directly or at the very least write unit tests that assert that the values on `TemporalRuntime` are equal to these values; will help with refactoring one side but forgetting to do so on the other.

src/Temporalio.Extensions.WorkflowStreams/.editorconfig
```diff
@@ -0,0 +1,13 @@
+[*.workflow.cs]
```
jmaeagle99: This looks to narrow in on one file. Use inline suppressions instead of this more wide casting `.editorconfig` file. That's the established pattern throughout the rest of this repository.

src/Temporalio.Extensions.WorkflowStreams/StreamPublisher.cs
```diff
@@ -0,0 +1,357 @@
...
+            lock (stateLock)
+            {
+                if (disposeTask != null)
+                {
+                    return disposeTask;
+                }
+                disposing = true;
+                timer?.Dispose();
+                timer = null;
+                stopSource.Cancel();
```
jmaeagle99: I would recommend figuring out a way to cancel this while not under a lock. This can invoke callbacks that might ultimately call back into the `StreamPublisher` and deadlock.

src/Temporalio.Extensions.WorkflowStreams/StreamPublisher.cs
```diff
@@ -0,0 +1,357 @@
...
+        {
+            lock (stateLock)
+            {
+                if (disposeTask != null)
+                {
+                    return disposeTask;
+                }
+                disposing = true;
+                timer?.Dispose();
+                timer = null;
```
jmaeagle99: nit: nulling out fields is not really necessary in disposal scenarios. If this is to help break GC cycle, then those pieces should be intentionally disconnected.

src/Temporalio.Extensions.WorkflowStreams/FlushTimeoutException.cs
```diff
@@ -0,0 +1,21 @@
...
+namespace Temporalio.Extensions.WorkflowStreams
+{
+    /// <summary>
+    /// Exception thrown when an ambiguously delivered publish batch exceeds its retry duration.
+    /// </summary>
+    /// <remarks>
+    /// The batch is dropped locally. It is present in the workflow log if the signal reached the
+    /// server, and otherwise is lost. WARNING: Workflow Streams is experimental and may change.
+    /// </remarks>
+    public class FlushTimeoutException : Exception
```
jmaeagle99: Probably should derive from `TemporalException` at the very least.

src/Temporalio.Extensions.WorkflowStreams/PollInput.cs
```diff
@@ -0,0 +1,27 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text.Json.Serialization;
+
+namespace Temporalio.Extensions.WorkflowStreams
+{
+    /// <summary>The cross-language update input for polling stream items.</summary>
```
jmaeagle99: XML doc comments are usually multi line to make it easier to update them in the future.

src/Temporalio.Extensions.WorkflowStreams/WorkflowStreamConstants.cs
```diff
@@ -0,0 +1,51 @@
+namespace Temporalio.Extensions.WorkflowStreams
```
jmaeagle99, line 4: Add `using System;` and don't qualify `System.TimeSpan` throughout.

src/Temporalio.Extensions.WorkflowStreams/WorkflowStreamClient.cs
```diff
@@ -0,0 +1,505 @@
...
+            /// <param name="error">Poll error, if present.</param>
+            internal PollAttempt(PollResult? result, string? runId, Exception? error)
+            {
+                this.result = result;
+                RunId = runId;
+                Error = error;
+            }
+
+            /// <summary>Gets the successful result, if present.</summary>
+            internal PollResult? Result => result;
```
jmaeagle99: Remove the `result` field and use auto-implemented properties like the other two.

src/Temporalio.Extensions.WorkflowStreams/WorkflowStreamClient.cs
```diff
@@ -0,0 +1,505 @@
...
+                if (!attempt.Result.MoreReady &&
+                    !await DelayAsync(
+                        options.PollCooldown, cancellationToken, rpcToken).ConfigureAwait(false))
+                {
+                    yield break;
+                }
+            }
+        }
+
+        private async Task<PollAttempt> PollOnceAsync(
```
jmaeagle99, line 393: Several of these methods are taking two `CancellationToken` instances, for which I don't understand the reason. These probably should just be one `CancellationToken cancellationToken` parameter and then the caller can create a combined token (if it really needs to) via `CancellationTokenSource.CreateLinkedTokenSource`.

src/Temporalio.Extensions.WorkflowStreams/WorkflowStream.workflow.cs
```diff
@@ -0,0 +1,374 @@
...
+            Workflow.Queries[WorkflowStreamConstants.OffsetQueryName] =
+                WorkflowQueryDefinition.CreateWithoutAttribute(
+                    WorkflowStreamConstants.OffsetQueryName,
+                    (Func<long>)(() => baseOffset + log.Count));
+        }
+
+        /// <summary>Gets a workflow-side publisher for a topic.</summary>
+        /// <param name="name">Topic name. Null is represented by the empty topic.</param>
+        /// <returns>A memoized topic handle.</returns>
+        public WorkflowTopicHandle Topic(string? name)
```
jmaeagle99: Methods should be named as actions e.g. `GetTopic`.

src/Temporalio.Extensions.WorkflowStreams/WorkflowStreamClient.cs
```diff
@@ -0,0 +1,505 @@
...
+            var context = ActivityExecutionContext.Current;
+            var id = context.Info.WorkflowId ?? throw new InvalidOperationException(
+                "WorkflowStreamClient.FromActivity requires a workflow activity");
+            return new(context.TemporalClient, id, options, context.PayloadConverter);
+        }
+
+        /// <summary>Gets a memoized handle for a topic.</summary>
+        /// <param name="name">Topic name. Null is represented by the empty topic.</param>
+        /// <returns>A topic handle.</returns>
+        public TopicHandle Topic(string? name)
```
jmaeagle99: Methods should be named as actions e.g. `GetTopic`.

src/Temporalio.Extensions.WorkflowStreams/WorkflowTopicHandle.cs
```diff
@@ -0,0 +1,25 @@
+namespace Temporalio.Extensions.WorkflowStreams
+{
+    /// <summary>A workflow-side handle bound to one topic.</summary>
+    /// <remarks>WARNING: Workflow Streams is experimental and may change.</remarks>
+    public sealed class WorkflowTopicHandle
```
jmaeagle99: `WorkflowStreamTopicHandle`?

src/Temporalio.Extensions.WorkflowStreams/TopicHandle.cs
jmaeagle99: `WorkflowStreamClientTopicHandle`?

src/Temporalio.Extensions.WorkflowStreams/PublishEntry.cs
```diff
@@ -0,0 +1,28 @@
+using System.Text.Json.Serialization;
+
+namespace Temporalio.Extensions.WorkflowStreams
+{
+    /// <summary>A single item in a publish signal batch.</summary>
+    /// <remarks>WARNING: Workflow Streams is experimental and may change.</remarks>
+    public sealed class PublishEntry
```
jmaeagle99: `internal`?

src/Temporalio.Extensions.WorkflowStreams/PollInput.cs
```diff
@@ -0,0 +1,27 @@
...
+    /// <remarks>WARNING: Workflow Streams is experimental and may change.</remarks>
+    public sealed class PollInput
```
jmaeagle99: All of the `Poll*` types should be internal.

src/Temporalio.Extensions.WorkflowStreams/WireItem.cs
```diff
@@ -0,0 +1,32 @@
+using System.Text.Json.Serialization;
+
+namespace Temporalio.Extensions.WorkflowStreams
+{
+    /// <summary>A stream item in the cross-language JSON envelope.</summary>
+    /// <remarks>WARNING: Workflow Streams is experimental and may change.</remarks>
+    public sealed class WireItem
```
jmaeagle99: `WorkflowStreamWireItem`?

jmaeagle99 (changes requested):

src/Temporalio.Extensions.WorkflowStreams/WorkflowStream.cs
brianstrauch: Renamed to WorkflowStream.cs!

src/Temporalio.Extensions.WorkflowStreams/README.md
brianstrauch: Updated to explicitly describe `Publish(value, forceFlush: true)` vs `await FlushAsync()`

src/Temporalio.Extensions.WorkflowStreams/WorkflowStreamConstants.cs
brianstrauch: Only three of these are needed for cross-language interoperability, marking the others as internal.

brianstrauch: Revisited this after comparing the official implementations. Go exports the three handler names and three standard error types, and Java exposes the same values through a public WorkflowStreamConstants class; both keep operational limits/defaults private. I restored public visibility for those six cross-language protocol constants while keeping paging limits and option defaults internal. The handler-name constants still reference TemporalRuntime as the single source of truth.

src/Temporalio.Extensions.WorkflowStreams/PollInput.cs
jmaeagle99: This fix should be applied to all of the comments.

src/Temporalio.Extensions.WorkflowStreams/WorkflowStreamClient.cs
```diff
@@ -0,0 +1,537 @@
...
+
+        /// <summary>Stops owned subscriptions and drains buffered publications.</summary>
+        /// <returns>A value task shared by concurrent disposal calls.</returns>
+        public ValueTask DisposeAsync()
+        {
+            lock (stateLock)
+            {
+                if (disposeTask == null)
+                {
+                    disposeSource.Cancel();
```
jmaeagle99, line 256: Similar to https://github.com/temporalio/sdk-dotnet/pull/884#discussion_r3994057165

src/Temporalio.Extensions.WorkflowStreams/WorkflowStreamClient.cs
```diff
@@ -0,0 +1,537 @@
...
+                        }
+                        continue;
+                    }
+
+                    WorkflowExecutionStatus status = WorkflowExecutionStatus.Unspecified;
+                    var describeCanceledByDispose = false;
+                    try
+                    {
+                        status = await DescribeStatusAsync(
+                            attempt?.RunId, rpcToken).ConfigureAwait(false);
```
jmaeagle99: I think the run id needs to be hoisted out of the loop and only updated if a new run ID actually exists. Otherwise, attempt failures will have is set to `null` and DescribeStatusAsync is then going to get status for the latest run instead of the prescribed run. Similar to https://github.com/temporalio/sdk-go/blob/main/contrib/workflowstreams/client.go#L170

src/Temporalio.Extensions.WorkflowStreams/WorkflowStreamClient.cs
```diff
@@ -0,0 +1,537 @@
...
+                        },
+                    }).ConfigureAwait(false);
+                return description.Status;
+            }
+            catch (OperationCanceledException)
+            {
+                cancellationToken.ThrowIfCancellationRequested();
+                return WorkflowExecutionStatus.Unspecified;
+            }
+            catch (RpcException err) when (
```
jmaeagle99: There are likely more conditions for while this exception is thrown. Consider catching broadly rather than for specific error codes. Otherwise, the caller (SubscribeCoreAsync) is going to be in a faulted state, masking the original poll error with an error from this call.

src/Temporalio.Extensions.WorkflowStreams/WorkflowStreamClient.cs
```diff
@@ -0,0 +1,537 @@
...
+                            Id = updateId,
+                            Rpc = new()
+                            {
+                                CancellationToken = cancellationToken,
+                                Timeout = options.RpcTimeout,
+                            },
+                        }).ConfigureAwait(false);
+                    break;
+                }
+                catch (WorkflowUpdateRpcTimeoutOrCanceledException)
```
jmaeagle99, line 486: Any concern with this being a possible hot path in error cases? Maybe check if the duration of the RPC is substantially less than the configured timeout and then wait some time (~PollCooldown) before starting a new update.

src/Temporalio.Extensions.WorkflowStreams/WorkflowStream.cs
```diff
@@ -0,0 +1,360 @@
...
+                    (Func<long>)(() => baseOffset + log.Count));
+        }
+
+        /// <summary>Gets a workflow-side publisher for a topic.</summary>
+        /// <param name="name">Topic name. Null is represented by the empty topic.</param>
+        /// <returns>A memoized topic handle.</returns>
+        public WorkflowStreamTopicHandle GetTopic(string? name)
+        {
+            name ??= string.Empty;
+            if (!topicHandles.TryGetValue(name, out var untypedHandle))
```
jmaeagle99: Are the `topicHandles` caches necessary? Seems handle construction is cheap. Nothing is bounding the cache or expiring handles. Nothing depends on the handle identity. Etc.

If they are kept, there is a subtle problem if a caller does:
```
var typed = client.GetTopic<Order>("orders");
var raw   = client.GetTopic("orders");
```
The untyped accessor will throw. Similar if these are reversed. Split the cache into two, one for typed and one for untyped.

src/Temporalio.Extensions.WorkflowStreams/WorkflowStreamClient.cs
```diff
@@ -0,0 +1,537 @@
...
+
+        /// <summary>Gets a memoized handle for a topic.</summary>
+        /// <param name="name">Topic name. Null is represented by the empty topic.</param>
+        /// <returns>A topic handle.</returns>
+        public WorkflowStreamClientTopicHandle GetTopic(string? name)
+        {
+            name ??= string.Empty;
+            lock (stateLock)
+            {
+                if (!topicHandles.TryGetValue(name, out var untypedHandle))
```
jmaeagle99: Same as https://github.com/temporalio/sdk-dotnet/pull/884/changes#r4020504076

src/Temporalio.Extensions.WorkflowStreams/WorkflowStreamConstants.cs
jmaeagle99: I'm not sure what customer scenario is being served by having these publicly available. If there is a legitimate customer use case, then that's fine. If not, I would make them internal. Making them public for a hypothetical scenario that never comes to fruition is a maintenance burden. If the only reason is "the other SDKs do it", I don't think that's enough to justify them being public. I'll leave this up for you to decide and won't block on it if they are not changed.

src/Temporalio.Extensions.WorkflowStreams/WorkflowStream.cs
```diff
@@ -0,0 +1,389 @@
...
+        /// <summary>
+        /// Detaches pollers, waits for handlers, captures state, and continues as new.
+        /// </summary>
+        /// <param name="createException">
+        /// Callback that creates the SDK continue-as-new exception from the captured state.
+        /// </param>
+        /// <returns>
+        /// A task that does not complete normally.
+        /// </returns>
+        public async Task ContinueAsNewAsync(
```
jmaeagle99: This API is analogous to `Workflow.CreateContinueAsNewException`. I would suggest refactoring to one of the following:
1. `Task<ContinueAsNewException> CreateContinueAsNewExceptionAsync(Func<WorkflowStreamState, ContinueAsNewException> createException)` or
2. `Task<WorkflowStreamState> CaptureStateForContinueAsNewAsync()` and let the caller call `Workflow.CreateContinueAsNewException` themselves.

The callback (either the existing code or suggestion 1) doesn't really help much in my opinion; the important part is detaching pollers and draining handlers.

Suggestion 1 makes it analogous to the existing continue-as-new API, expectation, and semantics for this SDK. Suggestion 2 eliminates the callback that doesn't provide much value.

Suggestion 1 usage:
```csharp
throw await stream.CreateContinueAsNewExceptionAsync(state =>
    Workflow.CreateContinueAsNewException(
        (OrderWorkflow workflow) => workflow.RunAsync(
            input with { StreamState = state })));
```

Suggestion 2 usage:
```csharp
var state = await stream.CaptureStateForContinueAsNewAsync();
throw Workflow.CreateContinueAsNewException(
    (OrderWorkflow workflow) => workflow.RunAsync(
          input with { StreamState = state }));
```

src/Temporalio.Extensions.WorkflowStreams/WorkflowStream.cs
```diff
@@ -0,0 +1,389 @@
...
+            Func<WorkflowStreamState, ContinueAsNewException> createException)
+        {
+            DetachPollers();
+            await Workflow.WaitConditionAsync(() => Workflow.AllHandlersFinished);
+            throw createException(GetState());
```
jmaeagle99: I think it would be good to somehow have a latch or guard for the final state capture to prevent publishing to any more topics, similar to how the draining state is used as a guard. But I'm not certain if that can be done _cleanly_ for the manual construction route without introducing another API or passing some kind of enum value to `GetState` to indicate the use.

jmaeagle99 (approved):
One more suggestion and another comment, otherwise LGTM.

src/Temporalio.Extensions.WorkflowStreams/WorkflowStream.cs
brianstrauch: Went with suggestion 2

src/Temporalio.Extensions.WorkflowStreams/WorkflowStream.cs
brianstrauch: Introduced `stateCapturedForContinueAsNew`

PR #887, opened 2026-09-08
Fix System Nexus codec serialization context

tconley1428:
## Summary
- retain operation-derived serialization context for pending System Nexus operations
- apply that context to scheduled command codec processing and Nexus resolution payload/failure decoding
- recognize WIT-declared no-payload System Nexus envelopes, including signal-with-start responses

## Testing
- mise run gen:nexus
- dotnet build tests/Temporalio.Tests --no-restore
- dotnet test tests/Temporalio.Tests --no-restore --filter "FullyQualifiedName=Temporalio.Tests.Worker.SystemNexusTests.ExecuteWorkflowAsync_SignalWithStart_CodecUsesTargetSerializationContext"

Full dotnet format --verify-no-changes is currently blocked by pre-existing IDE0005 errors in tests/Temporalio.SimpleBench/Program.cs.

### DIFF
--- a/src/Temporalio/Worker/WorkflowCodecHelper.cs
+++ b/src/Temporalio/Worker/WorkflowCodecHelper.cs
@@ -118,8 +118,13 @@ namespace Temporalio.Worker
                         }
                         break;
                     case WorkflowActivationJob.VariantOneofCase.ResolveNexusOperation:
-                        // TODO(cretz): Support Nexus serialization context
                         var nexusCodec = context.CodecNoContext;
+                        if (nexusCodec is IWithSerializationContext<IPayloadCodec> withNexus &&
+                            context.Instance?.GetPendingNexusOperationSerializationContext(
+                                job.ResolveNexusOperation.Seq) is { } nexusContext)
+                        {
+                            nexusCodec = withNexus.WithSerializationContext(nexusContext);
+                        }
                         if (nexusCodec == null)
                         {
                             break;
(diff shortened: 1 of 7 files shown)


src/Temporalio/Worker/WorkflowInstance.cs
jmaeagle99, line 2753: I know we changed this in https://github.com/temporalio/sdk-dotnet/commit/ca27a44bf8f829d2bb4fc302651b184dd12429a8#diff-7d5de4dd86e4fda729feeeac183fc71ca2f35e30324a3a631228d311e150f1d8, but I think it needs to be changed back to payloadConverter so that the `IPayloadConverter` and the `IPayloadCodec` that are effectively used are given the same context information.

src/Temporalio/Worker/WorkflowCodecHelper.cs
```diff
@@ -118,8 +118,13 @@ await childCodec2.DecodeFailureAsync(
                         }
                         break;
                     case WorkflowActivationJob.VariantOneofCase.ResolveNexusOperation:
-                        // TODO(cretz): Support Nexus serialization context
```
jmaeagle99, line 121: These comments should remain, or be adjusted to state that user-defined Nexus Operations do not get converted and encoded with a serialization context. You've only fixed this for System Nexus Operations.

jmaeagle99 (approved):

src/Temporalio/Worker/WorkflowInstance.cs
jmaeagle99: Opened a PR to fix this separately: https://github.com/temporalio/sdk-dotnet/pull/894

src/Temporalio/Worker/WorkflowCodecHelper.cs
tconley1428: Only sorta true. Today nexus does not have serialization context, though work on that is ongoing. But it isn't really a todo.

PR #892, opened 2026-09-10
Handle System Nexus payloads in Nexus Worker

VegetarianOrc:
## What was changed

- Detect marked Temporal System Nexus inputs in the Nexus worker and deserialize their outer envelope with the System Nexus payload converter.
- Apply payload codecs only to embedded user payloads while preserving the marked outer envelope.
- Move shared payload codec traversal into `PayloadCodecHelper` for use by workflow and Nexus worker paths.
- Add serializer and workflow-to-worker coverage for System Nexus conversion and nested codec application.

## Why?

System Nexus inputs are protobuf envelopes containing user payloads. The Nexus worker needs to handle these envelopes to support upcoming features.

## Checklist

1. How was this tested:
   - Added tests for new functionality
   - Existing tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,8 @@ to docs, or any other relevant information.
 
 ### Added
 
+- Nexus workers now decode marked Temporal System Nexus operation inputs with the System Nexus converter.
+- NexusPayloadSerializer now preserves marked System Nexus envelopes while applying payload codecs only to their embedded user payloads.
 - Added operation-specific outbound workflow interceptors for Temporal System Nexus operations.
   `SignalWithStartWorkflowAsync` can now be intercepted before the existing
   `ScheduleSystemNexusOperationAsync` hook.
@@ -54,7 +56,7 @@ to docs, or any other relevant information.
   non-retryable `BadRequest` handler exception (with the application failure as its cause) instead of
   an `Internal` one. This lets a data converter signal that the input itself is invalid. The handler
   exception is reported as `Invalid operation input`, which is distinct from the `failed to decode
-  Nexus operation input` message used when decoding itself fails. Application failures of any other
+Nexus operation input` message used when decoding itself fails. Application failures of any other
   error type, and retryable `PayloadValidationError` failures, keep their existing behavior.
 
 ### Fixed
@@ -197,6 +199,7 @@ to docs, or any other relevant information.
 - Fixed `ClientEnvConfig` TLS-disabled profiles to preserve disabled TLS in connection options.
 - OTLP metric export failures are now logged through Core telemetry when OpenTelemetry's periodic metric reader reports an export error.
 - Worker heartbeat now samples host CPU/memory at the heartbeat interval (only when enabled) rather than every 100ms.
+
 ### Added
 
 - Added the `[TemporalOperation]` attribute for declaring a Temporal-backed Nexus operation start
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -69,7 +69,7 @@ to docs, or any other relevant information.
   non-retryable `BadRequest` handler exception (with the application failure as its cause) instead of
   an `Internal` one. This lets a data converter signal that the input itself is invalid. The handler
   exception is reported as `Invalid operation input`, which is distinct from the `failed to decode
-Nexus operation input` message used when decoding itself fails. Application failures of any other
+- Nexus operation input` message used when decoding itself fails. Application failures of any other
   error type, and retryable `PayloadValidationError` failures, keep their existing behavior.
 
 ### Fixed
@@ -212,7 +212,6 @@ Nexus operation input` message used when decoding itself fails. Application fail
 - Fixed `ClientEnvConfig` TLS-disabled profiles to preserve disabled TLS in connection options.
 - OTLP metric export failures are now logged through Core telemetry when OpenTelemetry's periodic metric reader reports an export error.
 - Worker heartbeat now samples host CPU/memory at the heartbeat interval (only when enabled) rather than every 100ms.
-
 ### Added
 
 - Added the `[TemporalOperation]` attribute for declaring a Temporal-backed Nexus operation start
--- a/src/Temporalio/Worker/NexusPayloadSerializer.cs
+++ b/src/Temporalio/Worker/NexusPayloadSerializer.cs
@@ -1,5 +1,4 @@
 using System;
-using System.Linq;
 using System.Threading.Tasks;
 using Google.Protobuf;
 using NexusRpc;
@@ -7,6 +6,7 @@ using NexusRpc.Handlers;
 using Temporalio.Api.Common.V1;
 using Temporalio.Converters;
 using Temporalio.Exceptions;
+using Temporalio.Nexus;
 
 namespace Temporalio.Worker
 {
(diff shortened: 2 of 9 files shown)


tests/Temporalio.Tests/Worker/NexusPayloadSerializerTests.cs
```diff
@@ -9,15 +9,132 @@ namespace Temporalio.Tests.Worker;
...
+            DecodeCount++;
+            return Task.FromResult<IReadOnlyCollection<Payload>>(new[]
+            {
+                DataConverter.Default.PayloadConverter.ToPayload("decoded-input"),
+            });
+        }
+    }
+
+    [Fact]
+    public async Task DeserializeAsync_UnmarkedPayload_UsesUserConverter()
```
tconley1428, line 127: We should probably add some tests around serialization context as well.

CHANGELOG.md
```diff
@@ -210,6 +212,7 @@ to docs, or any other relevant information.
 - Fixed `ClientEnvConfig` TLS-disabled profiles to preserve disabled TLS in connection options.
 - OTLP metric export failures are now logged through Core telemetry when OpenTelemetry's periodic metric reader reports an export error.
 - Worker heartbeat now samples host CPU/memory at the heartbeat interval (only when enabled) rather than every 100ms.
+
```
jmaeagle99, line 215: Let's remove the extra line to make the tooling happy.

CHANGELOG.md
```diff
@@ -67,7 +69,7 @@ to docs, or any other relevant information.
   non-retryable `BadRequest` handler exception (with the application failure as its cause) instead of
   an `Internal` one. This lets a data converter signal that the input itself is invalid. The handler
   exception is reported as `Invalid operation input`, which is distinct from the `failed to decode
-  Nexus operation input` message used when decoding itself fails. Application failures of any other
+Nexus operation input` message used when decoding itself fails. Application failures of any other
```
jmaeagle99, line 72: Please fix the indent.

src/Temporalio/Worker/NexusPayloadSerializer.cs
```diff
@@ -62,13 +62,8 @@ public NexusPayloadSerializer(DataConverter dataConverter) =>
...
                 try
                 {
-                    var decoded = await dataConverter.PayloadCodec.DecodeAsync(
-                        new Payload[] { payload }).ConfigureAwait(false);
-                    if (decoded.Count != 1)
-                    {
-                        throw new ArgumentException($"Expected 1 payload, found {decoded.Count}");
-                    }
-                    payload = decoded.First();
+                    await PayloadCodecHelper.DecodeAsync(
```
jmaeagle99, line 65: This will now throw an exception if a payload is marked as a System Nexus Payload but the SDK doesn't recognize the message type. It'll raise an InvalidOperationException stating "Unrecognized marked System Nexus envelope message type:", which is correct. However, this is caught by the general exception catch clause (and passes the filter because it is not an `ApplicationFailureException` nor a `HandlerException`) and state that `Payload codec failed to decode Nexus operation input`, which is very misleading because a codec wasn't involved. This will continuously retry.

src/Temporalio/Worker/NexusPayloadSerializer.cs
```diff
@@ -97,7 +92,11 @@ public NexusPayloadSerializer(DataConverter dataConverter) =>
             object? result;
             try
             {
-                result = dataConverter.PayloadConverter.ToValue(payload, type);
+                var payloadConverter = SystemNexusPayloadVisitor.IsSystemPayload(payload) ?
+                    new SystemNexusPayloadConverter(
+                        dataConverter.PayloadConverter, dataConverter.FailureConverter) :
+                    dataConverter.PayloadConverter;
+                result = payloadConverter.ToValue(payload, type);
```
jmaeagle99, line 99: In the case where no codec is configured and the payload is an unrecognized system payload, the general exception catch will throw a non-retryable error. This is a different behavior than in https://github.com/temporalio/sdk-dotnet/pull/892/changes#r3984290006 only based on whether a codec was configured or not.

The behavior should be uniform.

I'm thinking maybe we should check if the payload is an unknown system payload (because someone is using another lang SDK that does know about that system payload type but this version of the .NET SDK does not) and throw a more targeted exception message. I think retryable would be good because it's possible they just need to update their SDK version.

src/Temporalio/Worker/NexusPayloadSerializer.cs
jmaeagle99, line 35: Should this be updated to handle System Nexus Payloads as well? Not sure if there is going to be an actual need for it.

jmaeagle99 (changes requested):

src/Temporalio/Worker/NexusPayloadSerializer.cs
VegetarianOrc: There isn't a need for it currently so I think I'd prefer to wait until we have a need to make the change. When responding to a start operation task, we aren't able to check if the target endpoint is the system and the value doesn't currently have any indication that it's a System Nexus Payload as either. One option would be to assume the response is a system payload if the input was a system payload, but I'm not 100% sure that would be correct in all situations.

Open to opinions and options here as always!

jmaeagle99: Deferring until there's an actual use case is good with me.

src/Temporalio/Worker/NexusPayloadSerializer.cs
VegetarianOrc: I did a bit of refactoring to align the behavior, let me know what you think!

CHANGELOG.md
```diff
@@ -71,8 +76,8 @@ to docs, or any other relevant information.
   non-retryable `BadRequest` handler exception (with the application failure as its cause) instead of
   an `Internal` one. This lets a data converter signal that the input itself is invalid. The handler
   exception is reported as `Invalid operation input`, which is distinct from the `failed to decode
-  Nexus operation input` message used when decoding itself fails. Application failures of any other
-  error type, and retryable `PayloadValidationError` failures, keep their existing behavior.
+- Nexus operation input`message used when decoding itself fails. Application failures of any other
```
jmaeagle99: Please remove the dash and make sure the indentation matches. It's a continuation of the sentence on the previous line.

jmaeagle99 (commented):
I agree with https://github.com/temporalio/sdk-dotnet/pull/892#discussion_r3982759218 in that we should have some basic serialization context test here. Doesn't have to be thorough (that should be the job of `SystemNexusPayloadConverterTests`) but just enough to validate that `NexusPayloadSerializer` does wire it through and its effect is observable.

VegetarianOrc:
> I agree with [#892 (comment)](https://github.com/temporalio/sdk-dotnet/pull/892#discussion_r3982759218) in that we should have some basic serialization context test here. Doesn't have to be thorough (that should be the job of `SystemNexusPayloadConverterTests`) but just enough to validate that `NexusPayloadSerializer` does wire it through and its effect is observable.

There currently isn't any applicable context to verify. The existing System Nexus calls happen from within a workflow and receive `ISerializationContext.Workflow`. This is adding system payload handling to the Nexus worker which processes tasks unrelated to the workflow context. The type of tasks possible are 1) the handler task from a 'normal' Nexus operation, e.g. a workflow or a standalone client calls `startNexusOperation` and 2) the server sends a Nexus Handler Callback with the envelope payload marked as a system payload to trigger the expected codec visitation and conversion. #908 Adds the first instance of `ISerializationContext.Nexus` that we could use. I think we should ensure the coverage here is added as part of that change.

